PR #3, opened 2020-06-11
fix all test cases

hicqu:

- [x] update test cases about pipe log file's name
- [x] truncate can accept an offset larger than the current but does nothing
- [x] seek active fd to end and update some position pointers

### DIFF
--- a/src/pipe_log.rs
+++ b/src/pipe_log.rs
@@ -170,6 +170,9 @@ impl PipeLog {
             manager.all_files.push_back(fd);
             if current_file == manager.active_file_num {
                 manager.active_log_fd = fd;
+                manager.active_log_size = unsafe { libc::lseek(fd, 0, libc::SEEK_END) as usize };
+                manager.active_log_capacity = manager.active_log_size;
+                manager.last_sync_size = manager.active_log_size;
             }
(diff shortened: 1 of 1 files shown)


src/pipe_log.rs
```diff
@@ -453,6 +456,9 @@ impl PipeLog {
     pub fn truncate_active_log(&self, offset: usize) -> Result<()> {
         {
             let manager = self.log_manager.read().unwrap();
+            if manager.active_log_size <= offset {
+                return Ok(());
```
zhangjinpeng1987, line 460: panic

src/pipe_log.rs
```diff
@@ -170,6 +170,9 @@ impl PipeLog {
             manager.all_files.push_back(fd);
             if current_file == manager.active_file_num {
                 manager.active_log_fd = fd;
+                manager.active_log_size = unsafe { libc::lseek(fd, 0, libc::SEEK_END) as usize };
+                manager.active_log_capacity = manager.active_log_size;
+                manager.last_sync_size = manager.active_log_size;
```
zhangjinpeng1987, line 175: Can you confirm all these data has been synced to disk?

PR #13, opened 2020-08-12
add gc and purge

hicqu:

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -51,10 +51,10 @@ impl Default for Config {
             recovery_mode: RecoveryMode::TolerateCorruptedTailRecords,
             bytes_per_sync: ReadableSize::kb(256),
             target_file_size: ReadableSize::mb(128),
-            purge_threshold: ReadableSize::gb(2),
+            purge_threshold: ReadableSize::gb(20),
             garbage_threshold: 0.75,
             cache_limit: ReadableSize::gb(1),
-            cache_limit_per_raft: ReadableSize::mb(4),
+            cache_limit_per_raft: ReadableSize::mb(128),
         }
     }
 }
--- a/src/config.rs
+++ b/src/config.rs
@@ -32,9 +32,6 @@ pub struct Config {
     /// Only purge if disk file size is greater than `purge_threshold`.
     pub purge_threshold: ReadableSize,
 
-    /// Only purge if garbage is greater than `garbage_threshold`.
-    pub garbage_threshold: f64,
-
     /// Total size limit to cache log entries.
     ///
     /// FIXME: it doesn't make effect currently.
@@ -51,8 +48,7 @@ impl Default for Config {
             recovery_mode: RecoveryMode::TolerateCorruptedTailRecords,
             bytes_per_sync: ReadableSize::kb(256),
             target_file_size: ReadableSize::mb(128),
-            purge_threshold: ReadableSize::gb(20),
-            garbage_threshold: 0.75,
+            purge_threshold: ReadableSize::gb(10),
             cache_limit: ReadableSize::gb(1),
             cache_limit_per_raft: ReadableSize::mb(128),
         }
@@ -65,9 +61,6 @@ impl Config {
     }
 
     pub fn validate(&self) -> Result<()> {
-        if self.garbage_threshold >= 1.0 {
-            return Err(box_err!("Invalid garbage_threshold"));
-        }
         if self.purge_threshold.0 < self.target_file_size.0 {
             return Err(box_err!("purge_threshold < target_file_size"));
         }
@@ -98,7 +91,6 @@ mod tests {
             bytes-per-sync = "2KB"
             target-file-size = "1MB"
             purge-threshold = "3MB"
-            garbage-threshold = 0.9
             cache-limit = "1GB"
             cache-limit-per-raft = "8MB"
         "#;
@@ -108,7 +100,6 @@ mod tests {
         assert_eq!(load.bytes_per_sync, ReadableSize::kb(2));
         assert_eq!(load.target_file_size, ReadableSize::mb(1));
         assert_eq!(load.purge_threshold, ReadableSize::mb(3));
-        assert_eq!(load.garbage_threshold, 0.9);
         assert_eq!(load.cache_limit, ReadableSize::gb(1));
         assert_eq!(load.cache_limit_per_raft, ReadableSize::mb(8));
     }
--- a/src/config.rs
+++ b/src/config.rs
@@ -13,39 +13,48 @@
 
 use crate::{util::ReadableSize, Result};
 
+#[derive(Clone, Copy, Debug, Eq, PartialEq, Serialize, Deserialize)]
+#[serde(rename_all = "kebab-case")]
+pub enum RecoveryMode {
+    TolerateCorruptedTailRecords = 0,
+    AbsoluteConsistency = 1,
+}
+
 #[derive(Clone, Debug, Serialize, Deserialize, PartialEq)]
 #[serde(default)]
 #[serde(rename_all = "kebab-case")]
 pub struct Config {
     pub dir: String,
-    pub recovery_mode: i32,
+    pub recovery_mode: RecoveryMode,
     pub bytes_per_sync: ReadableSize,
     pub target_file_size: ReadableSize,
-    pub cache_size_limit: ReadableSize,
-    pub total_size_limit: ReadableSize,
 
-    // Use raftstore.cfg.raft_log_gc_threshold
-    #[doc(hidden)]
-    #[serde(skip_serializing)]
-    pub compact_threshold: usize,
+    /// Only purge if disk file size is greater than `purge_threshold`.
+    pub purge_threshold: ReadableSize,
 
-    // Use raftstore.cfg.region_split_size
-    #[doc(hidden)]
-    #[serde(skip_serializing)]
-    pub region_size: ReadableSize,
+    /// Only purge if garbage is greater than `garbage_threshold`.
+    pub garbage_threshold: f64,
+
+    /// Total size limit to cache log entries.
+    ///
+    /// FIXME: it doesn't make effect currently.
+    pub cache_limit: ReadableSize,
+
+    /// Size limit to cache log entries for every Raft.
+    pub cache_limit_per_raft: ReadableSize,
 }
 
 impl Default for Config {
     fn default() -> Config {
         Config {
             dir: "".to_owned(),
-            recovery_mode: 0,
+            recovery_mode: RecoveryMode::TolerateCorruptedTailRecords,
             bytes_per_sync: ReadableSize::kb(256),
             target_file_size: ReadableSize::mb(128),
-            cache_size_limit: ReadableSize::gb(2),
-            total_size_limit: ReadableSize::gb(20),
-            compact_threshold: 0,
-            region_size: ReadableSize::mb(0),
+            purge_threshold: ReadableSize::gb(2),
+            garbage_threshold: 0.75,
+            cache_limit: ReadableSize::gb(1),
+            cache_limit_per_raft: ReadableSize::mb(4),
         }
     }
 }
@@ -56,37 +65,15 @@ impl Config {
     }
 
     pub fn validate(&self) -> Result<()> {
-        if self.total_size_limit.0 <= self.target_file_size.0 {
-            return Err(box_err!(
-                "Total size limit {:?} less than target file size {:?}",
-                self.total_size_limit,
-                self.target_file_size
-            ));
+        if self.garbage_threshold >= 1.0 {
+            return Err(box_err!("Invalid garbage_threshold"));
         }
-
-        if self.cache_size_limit.0 < self.target_file_size.0 {
-            return Err(box_err!(
-                "Cache size limit {:?} less than target file size {:?}",
-                self.cache_size_limit,
-                self.target_file_size
-            ));
+        if self.purge_threshold.0 < self.target_file_size.0 {
+            return Err(box_err!("purge_threshold < target_file_size"));
         }
-
-        if self.total_size_limit.0 < self.cache_size_limit.0 {
-            return Err(box_err!(
-                "Total size limit {:?} is less than cache limit size {:?}",
-                self.total_size_limit,
-                self.cache_size_limit
-            ));
+        if self.cache_limit_per_raft.0 > self.cache_limit.0 {
+            return Err(box_err!("cache_limit_per_raft > cache_limit"));
         }
-
-        if self.recovery_mode < 0 || self.recovery_mode > 1 {
-            return Err(box_err!(
-                "Unknown recovery mode {} for raftengine",
-                self.recovery_mode
-            ));
-        }
-
         Ok(())
     }
 }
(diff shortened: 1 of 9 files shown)


src/config.rs
```diff
@@ -13,39 +13,48 @@
...
     pub target_file_size: ReadableSize,
-    pub cache_size_limit: ReadableSize,
-    pub total_size_limit: ReadableSize,
 
-    // Use raftstore.cfg.raft_log_gc_threshold
-    #[doc(hidden)]
-    #[serde(skip_serializing)]
-    pub compact_threshold: usize,
+    /// Only purge if disk file size is greater than `purge_threshold`.
+    pub purge_threshold: ReadableSize,
```
zhangjinpeng1987, line 33: I think total-size-limit is more friendly to user.

zhangjinpeng1987: Or capacity

src/engine.rs
```diff
@@ -141,28 +123,10 @@ impl FileEngineInner {
                 }
             }
 
-            // Only keep latest entries in cache, keep cache below limited size.
```
zhangjinpeng1987, line 144: Why remove this logic?

src/engine.rs
```diff
@@ -176,7 +140,7 @@ impl FileEngineInner {
                         .write()
                         .unwrap();
                     let memtable = memtables.entry(region_id).or_insert_with(|| {
-                        let cache_limit = self.cfg.region_size.0 / 2;
+                        let cache_limit = self.cfg.cache_limit.0;
```
zhangjinpeng1987, line 143: s/cache_limit/cache_limit_per_raft

src/engine.rs
hicqu: It's already handled in `MemTable::append`.

src/config.rs
hicqu: `total-size-limit` looks like a hard limit for storage space.

src/config.rs
```diff
@@ -13,39 +13,48 @@
...
             dir: "".to_owned(),
-            recovery_mode: 0,
+            recovery_mode: RecoveryMode::TolerateCorruptedTailRecords,
             bytes_per_sync: ReadableSize::kb(256),
             target_file_size: ReadableSize::mb(128),
-            cache_size_limit: ReadableSize::gb(2),
-            total_size_limit: ReadableSize::gb(20),
-            compact_threshold: 0,
-            region_size: ReadableSize::mb(0),
+            purge_threshold: ReadableSize::gb(2),
```
zhangjinpeng1987, line 54: 2GB is too small

zhangjinpeng1987: Please consider the rolling update scenario.

zhangjinpeng1987: And the scenario that one TiKV shutdown for 10minutes and then start again.

src/engine.rs
```diff
@@ -219,215 +191,40 @@ impl FileEngineInner {
...
+    }
+
+    fn compact_to(&self, region_id: u64, index: u64) -> u64 {
+        let first_index = match self.first_index(region_id) {
+            Some(index) => index,
+            None => return 0,
+        };
+
+        let log_batch = LogBatch::new();
+        log_batch.add_command(Command::Compact { region_id, index });
```
zhangjinpeng1987, line 224: I don't know if trigger a write for very compact is effective

src/engine.rs
```diff
@@ -219,215 +191,40 @@ impl FileEngineInner {
...
+        self.write(log_batch, true).map(|_| ()).unwrap();
```
zhangjinpeng1987, line 225: I prefer buffer write for compact command, this will reduce the consumption of fsync call.

src/config.rs
```diff
@@ -13,39 +13,48 @@
...
+            garbage_threshold: 0.75,
+            cache_limit: ReadableSize::gb(1),
+            cache_limit_per_raft: ReadableSize::mb(4),
```
zhangjinpeng1987, line 57: I think the default value is a bit of small, please think about the scenario that there is a very little number of hot regions.

src/engine.rs
```diff
@@ -219,215 +191,40 @@ impl FileEngineInner {
...
-        if let Some(memtable) = memtables.get_mut(&region_id) {
-            return memtable.compact_to(index);
+        if let Some(memtable) = memtables.get(&region_id) {
+            return memtable.first_index();
         }
-        0
+        None
+    }
+
+    fn compact_to(&self, region_id: u64, index: u64) -> u64 {
```
zhangjinpeng1987, line 217: Seems you forget to compact the memtable.

src/engine.rs
```diff
@@ -663,6 +485,29 @@ impl FileEngine {
...
     }
+
+    // Only gc but not purge.
+    fn do_gc(&self, raft_group_id: u64, to: u64) -> usize {
+        self.inner.compact_to(raft_group_id, to) as usize
+    }
+
+    // Do purge and update some internal stats.
+    fn do_purge(&self) {
+        // TODO: rewrite inactive logs.
```
zhangjinpeng1987, line 496: So the rewrite inactive logs logic is missing?

src/engine.rs
```diff
@@ -663,6 +485,29 @@ impl FileEngine {
...
+        self.inner.purge_expired_files().unwrap();
+        self.inner.cache_stats.on_purge();
```
zhangjinpeng1987, line 498: I think the calculating logic of on_purge is not correct before the rewrite inactive logs logic is done.

src/config.rs
```diff
@@ -13,39 +13,48 @@
...
+    /// Only purge if disk file size is greater than `purge_threshold`.
+    pub purge_threshold: ReadableSize,
+
+    /// Only purge if garbage is greater than `garbage_threshold`.
+    pub garbage_threshold: f64,
+
+    /// Total size limit to cache log entries.
+    ///
+    /// FIXME: it doesn't make effect currently.
+    pub cache_limit: ReadableSize,
```
zhangjinpeng1987, line 41: Seems this parameter is not used?

src/engine.rs
hicqu: It's done when the write batch is applied to memtable.

src/engine.rs
hicqu: It's necessary, otherwise we can't know which logs are compacted after recovery.

src/config.rs
```diff
@@ -56,37 +61,12 @@ impl Config {
...
-            ));
-        }
-
-        if self.recovery_mode < 0 || self.recovery_mode > 1 {
-            return Err(box_err!(
-                "Unknown recovery mode {} for raftengine",
-                self.recovery_mode
-            ));
+        if self.cache_limit_per_raft.0 > self.cache_limit.0 {
+            return Err(box_err!("cache_limit_per_raft > cache_limit"));
```
zhangjinpeng1987, line 68: How about adjusting to the default value?

src/engine.rs
```diff
@@ -1,41 +1,24 @@
...
-impl From<i32> for RecoveryMode {
-    fn from(i: i32) -> RecoveryMode {
-        assert!(
-            RecoveryMode::TolerateCorruptedTailRecords as i32 <= i
-                && i <= RecoveryMode::AbsoluteConsistency as i32
-        );
-        unsafe { mem::transmute(i) }
-    }
-}
+const REWRITE_ENTRY_COUNT_THRESHOLD: usize = 128;
```
zhangjinpeng1987, line 21: I think it is too large. How about 16? Because we assume that we only rewrite these inactive regions, these regions should not have too many entries.

src/engine.rs
zhangjinpeng1987: But the total cache size is not limited when start, this may cause loop OOM.

src/engine.rs
```diff
@@ -219,188 +195,58 @@ impl FileEngineInner {
...
-    #[allow(dead_code)]
-    fn rewrite_inactive(&self) -> bool {
+    fn rewrite_inactive_logs(&self, will_force_compact: &mut Vec<u64>) {
         let inactive_file_num = {
-            self.pipe_log
-                .files_before(self.cfg.cache_size_limit.0 as usize)
+            let total_size = self.pipe_log.total_size();
+            let garbage_ratio = self.cache_stats.garbage_ratio();
+            let rewrite_limit = (total_size as f64 * (1.0 - garbage_ratio)) as usize;
+            self.pipe_log.last_file_before(rewrite_limit)
```
zhangjinpeng1987, line 203: The write amplification will increase, the rewrite line and force compact should use different size watermarks.

src/engine.rs
```diff
@@ -219,188 +195,58 @@ impl FileEngineInner {
...
+                    continue;
                 }
-            }
-        }
-        RAFTENGINE_MEMORY_USAGE_GAUGE.set(memory_usage as f64);
 
-        has_write
-    }
+                let log_batch = LogBatch::new();
+                log_batch.clean_region(region_id);
```
zhangjinpeng1987, line 226: I don't think this is necessary.

src/engine.rs
```diff
@@ -219,188 +195,58 @@ impl FileEngineInner {
...
-            return;
-        }
-
-        for slot in 0..SLOTS_COUNT {
-            let mut memtables = self.memtables[slot].write().unwrap();
-            for memtable in memtables.values_mut() {
-                memtable.evict_old_from_cache(inactive_file_num);
-            }
+        for log_batch in log_batches {
+            self.write(log_batch, false).unwrap();
```
zhangjinpeng1987, line 249: How about using multiple writes to prevent a huge batch which will trigger a jitter.

src/engine.rs
zhangjinpeng1987: it is ok if there are some redundant entries after a restart.

src/engine.rs
zhangjinpeng1987: In the 3 TiKV deployment, let's assume most of these regions are active, if one TiKV shut down, all these raft logs will not trigger compact command, and the garbage_ratio will almost equal 1, then the total raft log size will increase and reach an extremely large total size. There will no compact command before the region's raft logs size reaches the 2/3 region size threshold.

src/config.rs
hicqu: It confuses people. Let's keep it simple?

src/config.rs
hicqu: Yes, I will make it effect later, in an another PR.

src/engine.rs
```diff
@@ -1,56 +1,76 @@
...
-            RecoveryMode::TolerateCorruptedTailRecords as i32 <= i
-                && i <= RecoveryMode::AbsoluteConsistency as i32
-        );
-        unsafe { mem::transmute(i) }
-    }
-}
+// If a region has some very old raft logs less than this threshold,
+// rewrite them to clean stale log files ASAP.
+const REWRITE_ENTRY_COUNT_THRESHOLD: usize = 32;
+const PURGE_REWRITE_BEFORE: f64 = 0.3;
```
zhangjinpeng1987, line 25: hard to understand.

src/engine.rs
```diff
@@ -141,344 +161,250 @@ impl FileEngineInner {
...
-        let mut has_write = false;
-        let mut memory_usage = 0;
-        for slot in 0..SLOTS_COUNT {
-            let mut memtables = self.memtables[slot].write().unwrap();
-            for memtable in memtables.values_mut() {
-                memory_usage += memtable.entries_size();
+    // 0 means no inactive file.
+    fn max_inactive_file_num(&self) -> u64 {
+        let total_size = self.pipe_log.total_size();
+        let garbage_ratio = PURGE_REWRITE_BEFORE;
```
zhangjinpeng1987, line 202: Please remove the garbage-ratio concept.

src/engine.rs
```diff
@@ -141,344 +161,250 @@ impl FileEngineInner {
...
-                    Some(file_num) => file_num,
-                    None => continue,
-                };
+    // Scan all regions and for every one
+    // 1. return it if raft logs size between committed index and last index is greater than
+    //    `force_compact_threshold`;
+    // 2. rewrite inactive logs if it has less entries than `REWRITE_ENTRY_COUNT_THRESHOLD`.
+    fn regions_rewrite_or_force_compact(
+        &self,
+        last_inactive: u64,
```
zhangjinpeng1987, line 213: how about lastes_inactive_file_num

src/engine.rs
```diff
@@ -141,344 +161,250 @@ impl FileEngineInner {
...
+    }
 
-        let mut has_write = false;
-        let mut memory_usage = 0;
-        for slot in 0..SLOTS_COUNT {
-            let mut memtables = self.memtables[slot].write().unwrap();
-            for memtable in memtables.values_mut() {
-                memory_usage += memtable.entries_size();
+    // 0 means no inactive file.
+    fn max_inactive_file_num(&self) -> u64 {
```
zhangjinpeng1987, line 200: s/max/latest/

src/engine.rs
```diff
@@ -141,344 +161,250 @@ impl FileEngineInner {
...
+            let memtable = m.wl();
+            let region_id = memtable.region_id();
+
+            // Check the memtable needs force compaction or not.
+            if let Some(value) = memtable.get(RAFT_LOG_STATE_KEY) {
+                let mut raft_state = RaftLocalState::new();
+                raft_state.merge_from_bytes(&value).unwrap();
+                let committed = raft_state.get_hard_state().commit;
+                let last = raft_state.get_last_index();
+                let delay_size = memtable.entries_size_in(committed + 1, last + 1);
```
zhangjinpeng1987, line 232: heavy computing when lock held.

src/engine.rs
```diff
@@ -141,344 +161,250 @@ impl FileEngineInner {
...
-        }
-        RAFTENGINE_MEMORY_USAGE_GAUGE.set(memory_usage as f64);
+            all_ents.extend(ents.into_iter());
 
-        has_write
+            let mut kvs = Vec::new();
+            memtable.fetch_all_kvs(&mut kvs);
+            self.rewrite(region_id, all_ents, kvs, memtable).unwrap();
+        }
+        will_force_compact.sort();
```
zhangjinpeng1987, line 261: why sort?

src/engine.rs
```diff
@@ -141,344 +161,250 @@ impl FileEngineInner {
...
+    // 3. update all involved memtables.
+    // The lock logic is a little complex. However it's necessary because
+    // 1. "Inactive log rewrite" needs to keep logs on pipe log order;
+    // 2. Users can call `append` on one raft group concurrently.
+    // Maybe we can improve the implement of "inactive log rewrite" and
+    // forbid concurrent `append` to remove locks here.
     fn write(&self, log_batch: LogBatch, sync: bool) -> Result<usize> {
+        let mut memtables = Vec::with_capacity(log_batch.items.len());
+        for raft in log_batch.items.iter().map(|item| item.raft_group_id) {
+            memtables.push((raft, self.get_or_insert_memtable(raft)));
```
zhangjinpeng1987, line 332: Please get raft ids first, and deduplicate, and then get memtables. Because get memtable will acquire the mutex.

src/engine.rs
```diff
@@ -663,6 +601,32 @@ impl FileEngine {
...
     }
+
+    #[cfg(test)]
+    fn commit_to(&self, raft_group_id: u64, index: u64) {
+        let mut raft_state = self.get_raft_state(raft_group_id).unwrap().unwrap();
+        raft_state.mut_hard_state().commit = index;
+        self.put_raft_state(raft_group_id, &raft_state).unwrap();
+    }
+
+    pub fn all_rafts(&self) -> Vec<u64> {
```
zhangjinpeng1987, line 603: s/all_rafts/all_raft_group_ids

src/memtable.rs
```diff
@@ -249,11 +273,25 @@ impl MemTable {
         let cache_distance = self.cache_distance();
         if ioffset < cache_distance {
             self.cache_stats.miss_cache(1);
+            // TODO: remove it.
```
zhangjinpeng1987, line 276: please remove it

zhangjinpeng1987 (approved):
LGTM

PR #29, opened 2020-09-01
make pipe_log.rs more simple and clear

hicqu:
This PR trys to simplify `pipe_log.rs` and make it more clear. So that we can introduce other improvements easilier.

### DIFF
--- a/src/pipe_log.rs
+++ b/src/pipe_log.rs
@@ -1,10 +1,17 @@
 use std::collections::VecDeque;
 use std::fs::{self, File};
-use std::io::Read;
+use std::io::{Error as IoError, ErrorKind as IoErrorKind, Read};
 use std::os::unix::io::RawFd;
 use std::path::{Path, PathBuf};
 use std::sync::{Arc, Mutex, MutexGuard, RwLock};
 use std::{cmp, u64};
+
+use nix::errno::Errno;
+use nix::fcntl::{self, fallocate, OFlag};
+use nix::sys::stat::Mode;
+use nix::sys::uio::{pread, pwrite};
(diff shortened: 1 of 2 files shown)


src/pipe_log.rs
```diff
@@ -84,92 +124,76 @@ impl PipeLog {
...
-            let oflag = if current_file < manager.active_file_num {
-                // Open inactive files with readonly mode.
-                OFlag::O_RDONLY
-            } else {
-                // Open active file with readwrite mode.
-                OFlag::O_RDWR
-            };
-            let fd = fcntl::open(&path, oflag, Mode::S_IRWXU)
-                .map_err(|e| raft::StorageError::Other(e.into()))?;
+        for file_name in &logs[0..logs.len() - 1] {
```
Fullstop000, line 182: I think `&logs` is enough?

src/pipe_log.rs
```diff
@@ -311,40 +263,18 @@ impl PipeLog {
...
+        self.log_manager.wl().new_log_file(&self.dir)?;
+        self.write_header()?;
         Ok(())
     }
 
     pub fn write(&self, batch: &LogBatch, sync: bool, file_num: &mut u64) -> Result<usize> {
         if let Some(content) = batch.encode_to_bytes() {
             let bytes = content.len();
 
+            // TODO: `write` holds a big lock. Move the lock to log manager.
```
Fullstop000, line 275: Maybe create an issue for this?

Fullstop000 (approved):
Rest LGTM

src/pipe_log.rs
hicqu: the last one file is handled after the for loop. That file needs to be opened with flag `RDWR`.

PR #30, opened 2020-09-02
Remove dependency for raft-rs and kvproto

Little-Wallace:
Signed-off-by: Little-Wallace <bupt2013211450@gmail.com>

I think the repo raft-engine shall be a based library and could be used by other developer even if they does not use `raft-rs` (maybe some other raftlib) and `kvproto`.  I will implement the trait `RaftEngine` in `tikv:component/raft_engine` and define the trait `RaftEngine` in `tikv:component/engine_traits`. Their dependencies will be as following:
```
tikv ----> component/engine_traits, component/raft_engine
component/raft_engine --> raft-engine, component/engine_traits, kvproto, raft-rs
component/engine_rocks --> component/engine_traits, kvproto, raft-rs
```


src/engine.rs
```diff
@@ -372,12 +374,12 @@ impl FileEngine {
         Ok(bytes)
     }
 
-    fn rewrite(
+    pub fn rewrite(
```
hicqu, line 377: It shouldn't be `pub`.

src/engine.rs
```diff
@@ -396,22 +398,22 @@ impl FileEngine {
         Ok(())
     }
 
-    fn sync(&self) -> Result<()> {
+    pub fn sync(&self) -> Result<()> {
         self.pipe_log.sync()
     }
 
-    fn put_msg<M: protobuf::Message>(&self, region_id: u64, key: &[u8], m: &M) -> Result<()> {
+    pub fn put_msg<M: protobuf::Message>(&self, region_id: u64, key: &[u8], m: &M) -> Result<()> {
```
hicqu, line 405: Seems there is no `fn put`? It's better to add it.

src/engine.rs
```diff
@@ -733,11 +660,42 @@ impl RaftEngine for FileEngine {
...
+        }
+    }
 
-    fn append_log(engine: &FileEngine, raft: u64, entry: &Entry) {
+    fn append_log(engine: &RaftLogEngine, raft: u64, entry: &RaftEntry) {
         engine.append(raft, vec![entry.clone()]).unwrap();
-        let mut state = RaftLocalState::new();
-        state.last_index = entry.index;
+        let mut state = HardState::new();
+        state.commit = 0;
```
hicqu, line 698: I think it's not necessary.

src/engine.rs
```diff
@@ -733,11 +660,42 @@ impl RaftEngine for FileEngine {
...
+    const RAFT_LOG_STATE_KEY: &[u8] = b"R";
+    impl Entry for RaftEntry {
+        fn index(&self) -> u64 {
+            self.get_index()
+        }
+    }
+    pub trait RaftEngine {
+        fn put_raft_state(&self, raft_group_id: u64, state: &HardState) -> Result<()>;
+        fn get_raft_state(&self, raft_group_id: u64) -> Result<Option<HardState>>;
+        fn commit_to(&self, raft_group_id: u64, index: u64) {
```
hicqu, line 674: Seems it's not necessary. Could you take a try to remove it?

src/engine.rs
```diff
@@ -707,23 +638,19 @@ impl RaftEngine for FileEngine {
         Ok(will_force_compact)
     }
 
-    fn has_builtin_entry_cache(&self) -> bool {
+    pub fn has_builtin_entry_cache(&self) -> bool {
```
hicqu, line 641: It can be removed.

src/engine.rs
```diff
@@ -707,23 +638,19 @@ impl RaftEngine for FileEngine {
...
         true
     }
 
-    fn gc_entry_cache(&self, raft_group_id: u64, to: u64) {
-        self.compact_cache_to(raft_group_id, to)
-    }
-
-    fn flush_stats(&self) -> CacheStats {
+    pub fn flush_stats(&self) -> CacheStats {
```
hicqu, line 645: Would you mind to document the method?

hicqu:
rest LGTM. Thanks very much!

src/engine.rs
Little-Wallace: Ok

hicqu (approved):

PR #35, opened 2020-09-03
Prevent using fallocate on platforms other than linux

### LINKED ISSUE #34: Necessary to support macOS in pipelog
Maybe we can support macOS for the convenience in development from macOS users
@hicqu

longfangsong:
`fallocate` works only under linux, and many of our developers are not using linux.
Currently `fallocate` is always on, which prevent raft-engine and crates depends on it from building, eg. [tikv#8570](https://github.com/tikv/tikv/issues/8570).

This pr and `cfg`s to enable fallocate only on linux.

This should close #34.

### DIFF
--- a/src/pipe_log.rs
+++ b/src/pipe_log.rs
@@ -7,7 +7,9 @@ use std::sync::{Arc, Mutex, MutexGuard, RwLock};
 use std::{cmp, u64};
 
 use nix::errno::Errno;
-use nix::fcntl::{self, fallocate, OFlag};
+#[cfg(target_os = "linux")]
+use nix::fcntl::fallocate;
+use nix::fcntl::{self, OFlag};
 use nix::sys::stat::Mode;
 use nix::sys::uio::{pread, pwrite};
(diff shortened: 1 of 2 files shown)


longfangsong:
PS. Do we need edit the ci to make sure whether this can be built on any platforms?

src/pipe_log.rs
```diff
@@ -7,7 +7,9 @@ use std::sync::{Arc, Mutex, MutexGuard, RwLock};
 use std::{cmp, u64};
 
 use nix::errno::Errno;
-use nix::fcntl::{self, fallocate, OFlag};
+#[cfg(target_os = "linux")]
+use nix::fcntl::fallocate;
```
hicqu, line 11: I prefer use `nix::fcntl::fallocate` where it's called.

hicqu:
@longfangsong thanks a lot! Could you edit the github hook for CI on macOS?

longfangsong:
@hicqu PTAL
I don't know whether it is a good idea to make this library also work under windows, but that seems a huge work.

hicqu (approved):

PR #37, opened 2020-09-03
Refactor entry-trait

Little-Wallace:
EntryTrait can not be implement by third crate. So I create another `EntryExt` so that TiKV could implement it.


Cargo.toml
```diff
@@ -33,4 +32,4 @@ features = ["nightly"]
 git = "https://github.com/pingcap/raft-rs"
 branch = "master"
 default-features = false
-features = ["protobuf-codec"]
+features = ["protobuf-codec", "prost-codec"]
```
hicqu, line 35: Only one of them is necessary.

hicqu:
rust LGTM.

Issue #31, opened 2020-09-02
Change `PipeLog` to a trait


hicqu:
So that we can use a black hole implementation or a simple memory implementation to benchmark raft engine without any disk I/O.

Renkai:
pick this issue

Fullstop000:
For a simple mutex based memory storage system, maybe you can refer to https://github.com/Fullstop000/wickdb/blob/master/src/storage/mem.rs
@Renkai

PR #38, opened 2020-09-03
Pipe log trait

Renkai:
Fixed https://github.com/tikv/raft-engine/issues/31

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -35,8 +35,7 @@ pub mod pipe_log;
 pub mod util;
 
 pub use self::config::Config;
-pub use self::engine::Engine as RaftLogEngine;
-pub type FileEngine<X, Y> = self::engine::Engine<X, Y, FilePipeLog>;
+pub type RaftLogEngine<X, Y> = self::engine::Engine<X, Y, FilePipeLog>;
 
 pub use self::errors::{Error, Result};
 pub use self::log_batch::{EntryExt, LogBatch};
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -35,9 +35,12 @@ pub mod pipe_log;
 pub mod util;
 
 pub use self::config::Config;
-pub use self::engine::FileEngine as RaftLogEngine;
+pub use self::engine::Engine as RaftLogEngine;
+pub type FileEngine<X, Y> = self::engine::Engine<X, Y, FilePipeLog>;
+
 pub use self::errors::{Error, Result};
 pub use self::log_batch::{EntryExt, LogBatch};
+use crate::pipe_log::FilePipeLog;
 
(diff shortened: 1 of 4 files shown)


src/pipe_log.rs
```diff
@@ -232,11 +232,70 @@ pub struct PipeLog {
     current_read_file_num: u64,
 }
 
-impl PipeLog {
-    fn new(cfg: &Config, cache_submitor: CacheSubmitor) -> PipeLog {
+pub trait PipeLog: Sized + Clone + Send {
+    fn new(cfg: &Config, cache_submitor: CacheSubmitor) -> Self;
+
+    fn open(cfg: &Config, cache_submitor: CacheSubmitor) -> Result<Self>;
```
hicqu, line 238: Not all methods are necessary.

src/pipe_log.rs
```diff
@@ -219,7 +219,7 @@ impl LogManager {
 }
 
 #[derive(Clone)]
-pub struct PipeLog {
+pub struct PipeLogImpl {
```
Fullstop000, line 222: Prefer using a more specific name like `FilePipeLog` as there could be `MemPipleLog`

Renkai: > Prefer using a more specific name like `FilePipeLog` as there could be `MemPipleLog`

Renamed!

hicqu: @Renkai I guess you made a mistake. Should be `FilePipeLog` instead of `MemPipeLog`. The latter can be added later for benchmarks.

hicqu (approved):

hicqu:
I think we still need to
1. change `FileEngine` to `Engine`.
2. in `src/lib`, `pub use self::engine::Engine<X, Y, FilePipeLog> as FileEngine<X, Y>`.

src/lib.rs
```diff
@@ -35,9 +35,12 @@ pub mod pipe_log;
 pub mod util;
 
 pub use self::config::Config;
-pub use self::engine::FileEngine as RaftLogEngine;
+pub use self::engine::Engine as RaftLogEngine;
```
hicqu, line 38: This line can be removed.

src/lib.rs
```diff
@@ -35,9 +35,12 @@ pub mod pipe_log;
...
+pub type FileEngine<X, Y> = self::engine::Engine<X, Y, FilePipeLog>;
```
hicqu, line 39: `pub type RaftLogEngine<X, Y>` is better.

hicqu (approved):

hicqu:
@Fullstop000 PTAL if you are free, thanks.

Fullstop000 (approved):
LGTM

PR #40, opened 2020-09-11
re-implement rewrite logic

### LINKED ISSUE #36: Relax the lock during writing
The current write procedue:
```
sort memtables
lock all memtables
    lock pipe log -> append pipe log -> unlock pipe log
    write memtables
unlock memtables
```

At least it can be improved to
```
lock pipe log
    append pipe log
    lock all memtables
unlock pipe log
write memtables
unlock memtables
```

And maybe there are more potential improvements.

hicqu:
Close #36 .

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -182,7 +182,7 @@ where
                                 }
                             }
                         }
-                        Self::apply_to_memtable(memtables, log_batch, file_num);
+                        Self::apply_to_memtable(memtables, log_batch, queue, file_num);
                         offset = (buf.as_ptr() as usize - start_ptr as usize) as u64;
                     }
                     Ok(None) => {
@@ -213,13 +213,18 @@ where
     fn apply_log_item_to_memtable(
         memtables: &MemTableAccessor<E, W>,
         item: LogItem<E>,
+        queue: LogQueue,
         file_num: u64,
     ) {
         let memtable = memtables.get_or_insert(item.raft_group_id);
         match item.content {
             LogItemContent::Entries(entries_to_add) => {
                 let entries = entries_to_add.entries;
-                let entries_index = entries_to_add.entries_index.into_inner();
+                let mut entries_index = entries_to_add.entries_index.into_inner();
+                if queue == LogQueue::Rewrite {
+                    // By default the queue is for appending.
+                    entries_index.iter_mut().for_each(|ei| ei.queue = queue);
+                }
                 memtable.wl().append(entries, entries_index);
             }
             LogItemContent::Command(Command::Clean) => {
@@ -229,7 +234,13 @@ where
                 memtable.wl().compact_to(index);
             }
             LogItemContent::Kv(kv) => match kv.op_type {
-                OpType::Put => memtable.wl().put(kv.key, kv.value.unwrap(), file_num),
+                OpType::Put => {
+                    let (key, value) = (kv.key, kv.value.unwrap());
+                    match queue {
+                        LogQueue::Append => memtable.wl().put(key, value, file_num),
+                        LogQueue::Rewrite => memtable.wl().put_rewrite(key, value, file_num),
+                    }
+                }
                 OpType::Del => memtable.wl().delete(kv.key.as_slice()),
             },
         }
@@ -258,10 +269,11 @@ where
     fn apply_to_memtable(
         memtables: &MemTableAccessor<E, W>,
         log_batch: LogBatch<E, W>,
+        queue: LogQueue,
         file_num: u64,
     ) {
         for item in log_batch.items {
-            Self::apply_log_item_to_memtable(memtables, item, file_num);
+            Self::apply_log_item_to_memtable(memtables, item, queue, file_num);
         }
     }
 
@@ -350,11 +362,12 @@ where
     // Maybe we can improve the implement of "inactive log rewrite" and
     // forbid concurrent `append` to remove locks here.
     fn write_impl(&self, log_batch: &mut LogBatch<E, W>, sync: bool) -> Result<usize> {
+        let queue = LogQueue::Append;
         let mut file_num = 0;
         let bytes = self.pipe_log.write(log_batch, sync, &mut file_num)?;
         if file_num > 0 {
             for item in log_batch.items.drain(..) {
-                Self::apply_log_item_to_memtable(&self.memtables, item, file_num);
+                Self::apply_log_item_to_memtable(&self.memtables, item, queue, file_num);
             }
         }
         Ok(bytes)
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -1,8 +1,10 @@
 use std::io::BufRead;
 use std::sync::atomic::{AtomicUsize, Ordering};
-use std::sync::{Arc, Mutex, RwLock};
+use std::sync::{Arc, RwLock};
 use std::time::{Duration, Instant};
-use std::{cmp, fmt, u64};
+use std::{fmt, u64};
+
+use protobuf::Message;
 
 use crate::cache_evict::{
     CacheSubmitor, CacheTask, Runner as CacheEvictRunner, DEFAULT_CACHE_CHUNK_SIZE,
@@ -15,28 +17,23 @@ use crate::log_batch::{
 use crate::memtable::{EntryIndex, MemTable};
 use crate::pipe_log::{GenericPipeLog, LogQueue, PipeLog, FILE_MAGIC_HEADER, VERSION};
 use crate::util::{HandyRwLock, HashMap, Worker};
+use crate::PurgeManager;
 use crate::{codec, CacheStats, Result};
-use protobuf::Message;
 
 const SLOTS_COUNT: usize = 128;
-
-// If a region has some very old raft logs less than this threshold,
-// rewrite them to clean stale log files ASAP.
-const REWRITE_ENTRY_COUNT_THRESHOLD: usize = 32;
-const REWRITE_INACTIVE_RATIO: f64 = 0.7;
-const FORCE_COMPACT_RATIO: f64 = 0.2;
 
 // When add/delete a memtable, a write lock is required.
 // We can use LRU cache or slots to reduce conflicts.
 type MemTables<E, W> = HashMap<u64, Arc<RwLock<MemTable<E, W>>>>;
-pub struct MemTableAccessor<E: Message, W: EntryExt<E>> {
+
+pub struct MemTableAccessor<E: Message + Clone, W: EntryExt<E>> {
     slots: Vec<Arc<RwLock<MemTables<E, W>>>>,
     creator: Arc<dyn Fn(u64) -> MemTable<E, W> + Send + Sync>,
 }
 
 impl<E, W> Clone for MemTableAccessor<E, W>
 where
-    E: Message,
+    E: Message + Clone,
     W: EntryExt<E>,
 {
     fn clone(&self) -> Self {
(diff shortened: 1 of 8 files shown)


src/engine.rs
```diff
@@ -745,6 +732,36 @@ mod tests {
...
+            .tempdir()
+            .unwrap();
+
+        let mut cfg = Config::default();
+        cfg.dir = dir.path().to_str().unwrap().to_owned();
+        cfg.target_file_size = ReadableSize::kb(5);
+        cfg.purge_threshold = ReadableSize::kb(80);
+        let engine = RaftLogEngine::new(cfg.clone());
+
+        // Put 100 entries into 10 regions.
```
Fullstop000, line 757: This can be removed

src/pipe_log.rs
```diff
@@ -349,175 +369,187 @@ impl FilePipeLog {
...
+            LogQueue::Rewrite => (self.rewriter.wl(), REWRITE_SUFFIX),
+        };
+        let purge_count = manager.purge_to(file_num)?;
+        drop(manager);
+
+        for file_num in (file_num - purge_count as u64)..file_num {
+            let mut path = PathBuf::from(&self.dir);
+            path.push(generate_file_name(file_num, name_suffix));
+            if let Err(e) = fs::remove_file(&path) {
+                warn!("Remove purged log file {:?} fail: {}", path, e);
```
Fullstop000, line 522: Is this safe swallowing the error here?

src/pipe_log.rs
```diff
@@ -349,175 +369,187 @@ impl FilePipeLog {
...
+
+    fn purge_to(&self, queue: LogQueue, file_num: u64) -> Result<usize> {
+        let (mut manager, name_suffix) = match queue {
+            LogQueue::Append => (self.appender.wl(), LOG_SUFFIX),
+            LogQueue::Rewrite => (self.rewriter.wl(), REWRITE_SUFFIX),
+        };
+        let purge_count = manager.purge_to(file_num)?;
+        drop(manager);
+
+        for file_num in (file_num - purge_count as u64)..file_num {
```
Fullstop000, line 518: What if `file_num != purge_count` ? Will the stale file retain forever?

Little-Wallace:
Could we split code about `split` into another file just like `rewrite.rs` ?

src/engine.rs
```diff
@@ -190,144 +182,176 @@ where
...
-            let m = memtables.get_or_insert(item.raft_group_id);
-            Self::apply_log_item_to_memtable(item.content, &mut *m.wl(), file_num);
+            Self::apply_log_item_to_memtable(memtables, item, queue, file_num);
         }
     }
 
     // Returns (`latest_needs_rewrite`, `latest_needs_force_compact`).
     fn latest_inactive_file_num(&self) -> (u64, u64) {
-        let total_size = self.pipe_log.total_size() as f64;
+        let queue = LogQueue::Append;
```
Little-Wallace, line 282: How we remove inactive `LogQueue::Rewrite` file?

src/engine.rs
```diff
@@ -190,144 +182,176 @@ where
...
     fn apply_to_memtable(
         memtables: &MemTableAccessor<E, W>,
         log_batch: LogBatch<E, W>,
+        queue: LogQueue,
         file_num: u64,
     ) {
         for item in log_batch.items {
-            let m = memtables.get_or_insert(item.raft_group_id);
-            Self::apply_log_item_to_memtable(item.content, &mut *m.wl(), file_num);
+            Self::apply_log_item_to_memtable(memtables, item, queue, file_num);
```
Little-Wallace, line 276: Could we move the method `apply_log_item_to_memtable` as a member method of `MemTableAccessor `?

hicqu: I make it as a normal function.

src/engine.rs
```diff
@@ -113,74 +110,64 @@ where
...
-
-            // Verify file header
+            // Verify file header.
             let mut buf = content.as_slice();
             if buf.len() < FILE_MAGIC_HEADER.len() || !buf.starts_with(FILE_MAGIC_HEADER) {
-                if current_read_file != active_file_num {
-                    panic!("Raft log file {} is corrupted.", current_read_file);
+                if file_num != active_file_num {
+                    warn!("Raft log header is corrupted at {:?}.{}", queue, file_num);
+                    return Err(box_err!("Raft log file header is corrupted"));
```
Little-Wallace, line 154: Would the corruption happens just after the file has been created but not written HEADER?

src/memtable.rs
```diff
@@ -181,6 +187,18 @@ impl<E: Message + Clone, W: EntryExt<E>> MemTable<E, W> {
...
 
+        if let Some((queue, index)) = self.entries_index.back().map(|e| (e.queue, e.index)) {
+            if first_index_to_add > index + 1 {
+                assert_eq!(
+                    queue,
+                    LogQueue::Rewrite,
+                    "memtable {} has a hole",
+                    self.region_id
+                );
+                self.compact_to(index + 1);
```
Little-Wallace, line 198: Would it be a corruption that may lose data?

src/memtable.rs
```diff
@@ -200,8 +218,50 @@ impl<E: Message + Clone, W: EntryExt<E>> MemTable<E, W> {
...
     }
 
+    pub fn rewrite(&mut self, entries_index: Vec<EntryIndex>, latest_rewrite: u64) {
+        if entries_index.is_empty() || self.entries_index.is_empty() {
+            return;
+        }
+
+        let first = entries_index.first().unwrap().index;
+        let front = self.entries_index[self.rewrite_count].index;
+        assert!(front >= first);
```
Little-Wallace, line 228: How about add `assert_eq!(first, self.rewrite_count);`?

Little-Wallace: What if `self.entries_index` has been changed so that  the first index is not equal to the first index of `self.entries_index`?

src/engine.rs
hicqu: In this case `file_num == active_file_num` will be true.

src/memtable.rs
hicqu: It could be a corruption. Do you have any suggestions?

src/memtable.rs
hicqu: So `first` could not equal to `rewrite_count`. It's considered in the later code.

src/memtable.rs
hicqu: There could still be a corruption even if we do the index contiguous check when `self.entries_index` is empty. So, I prefer to omit the check.

src/pipe_log.rs
```diff
@@ -210,86 +270,46 @@ impl LogManager {
             self.active_log_capacity += alloc_size;
         }
 
-        if sync || self.reach_sync_limit() {
+        if *sync || self.reach_sync_limit() {
             self.last_sync_size = self.active_log_size;
-            sync = true;
+            *sync = true;
```
Little-Wallace, line 275: It is unnecessary because OS will flush the page cache to disk automatic.  So we do not need to check the size of data in os buffer.

src/pipe_log.rs
```diff
@@ -349,175 +370,195 @@ impl FilePipeLog {
...
+
+    fn latest_file_before(&self, queue: LogQueue, size: usize) -> u64 {
+        let cur_size = self.total_size(queue);
         if cur_size <= size {
             return 0;
         }
         let count = (cur_size - size) / self.rotate_size;
-        let manager = self.appender.rl();
-        manager.first_file_num + count as u64
+        self.get_queue(queue).first_file_num + count as u64
```
Little-Wallace, line 557: Should we check this number will be never larger than current_active_file?

Little-Wallace (approved):
mostly LGTM.
I think we shall define a method in a structure as much as possible rather than in module diredctly.

src/pipe_log.rs
```diff
@@ -148,6 +208,10 @@ impl LogManager {
         }
         let active_fd = self.get_active_fd().unwrap();
         ftruncate(active_fd.0, offset as _).map_err(|e| parse_nix_error(e, "ftruncate"))?;
+        if offset == 0 {
```
Fullstop000, line 211: How about `offset <= header_len`

src/pipe_log.rs
```diff
@@ -349,175 +370,195 @@ impl FilePipeLog {
...
     fn write<E: Message, W: EntryExt<E>>(
         &self,
         batch: &LogBatch<E, W>,
-        sync: bool,
+        mut sync: bool,
         file_num: &mut u64,
     ) -> Result<usize> {
         if let Some(content) = batch.encode_to_bytes() {
-            let bytes = content.len();
+            let entries_size = batch.entries_size();
```
Fullstop000, line 447: Two overall iterating in log batch here. Maybe optimize this later.

src/memtable.rs
Fullstop000: Maybe we can provide a public method to verify this from the upper level application

src/engine.rs
```diff
@@ -113,74 +110,64 @@ where
...
-                        "Read content of file {} failed, error {:?}",
-                        current_read_file, e
-                    )
-                })
-                .unwrap_or_else(|| panic!("Expect has content, but get None"));
-
-            // Verify file header
+            // Verify file header.
             let mut buf = content.as_slice();
             if buf.len() < FILE_MAGIC_HEADER.len() || !buf.starts_with(FILE_MAGIC_HEADER) {
```
Fullstop000, line 151: `!buf.starts_with(FILE_MAGIC_HEADER)` should be enough

src/pipe_log.rs
hicqu: Generally `bytes_per_sync` is less than the OS dirty page limit. So It's still meaningful.

src/engine.rs
```diff
@@ -338,107 +215,26 @@ where
...
             for item in log_batch.items.drain(..) {
-                let offset = rafts.binary_search(&item.raft_group_id).unwrap();
-                let m = &mut locked[offset];
-                Self::apply_log_item_to_memtable(item.content, &mut *m, file_num);
-            }
-        }
-        for memtable in locked {
-            if memtable.uninitialized() {
-                self.memtables.remove(memtable.region_id());
+                apply_log_item_to_memtable(&self.memtables, item, queue, file_num);
```
Fullstop000, line 223: Is `apply_to_memtable` suitable here?

src/engine.rs
```diff
@@ -896,13 +813,60 @@ mod tests {
...
+        }
+
+        // The engine needs purge, and all old entries should be rewritten.
+        assert!(engine.purge_manager.needs_purge_log_files());
+        assert!(engine.purge_expired_files().unwrap().is_empty());
+        assert!(engine.pipe_log.first_file_num(LogQueue::Append) > 1);
+
+        let active_num = engine.pipe_log.active_file_num(LogQueue::Rewrite);
+        let active_len = engine.pipe_log.file_len(LogQueue::Rewrite, active_num);
+        assert!(active_num > 1 || active_len > 59); // The rewrite queue isn't empty.
```
Fullstop000, line 851: What's the point of `active_len > 59`

Fullstop000:
Rest LGTM

src/engine.rs
hicqu: The rewrite queue must not be empty.

src/memtable.rs
hicqu: The API is not suitable here, because we can't get `truncated_index` from raft engine.

Fullstop000 (approved):

PR #45, opened 2020-09-16
A weird problem

koushiro:
Signed-off-by: koushiro <koushiro.cqx@gmail.com>


Fullstop000:
This looks confusing /cc @hicqu

hicqu:
The reason is `MemTableAccessor::collect`. This method will scan all slots in `MemTableAccessor`. In your change all slots reference to one same hashmap, so that the method can return duplicate memtables. Rewrite a memtable twice causes the panic.

hicqu:
And in `MemTable::fetch_rewrite_entries`, the check logic should be updated for this case. I will send a PR for this later.

hicqu:
I think it can be closed now.

PR #46, opened 2020-09-17
improve write logic to reduce cpu usage

hicqu:
* Avoid copy when compressing log batches;
* Calculate encoded_size during encoding a log batch;

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -165,11 +165,17 @@ where
             loop {
                 match LogBatch::from_bytes(&mut buf, file_num, offset) {
                     Ok(Some(mut log_batch)) => {
-                        let entries_size = log_batch.entries_size();
+                        let mut encoded_size = 0;
+                        for item in &log_batch.items {
+                            if let LogItemContent::Entries(ref entries) = item.content {
+                                encoded_size += entries.encoded_size.get();
+                            }
+                        }
+
                         if let Some(tracker) = pipe_log.cache_submitor().get_cache_tracker(
                             file_num,
                             offset,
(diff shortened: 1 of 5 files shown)


src/engine.rs
```diff
@@ -165,11 +165,17 @@ where
             loop {
                 match LogBatch::from_bytes(&mut buf, file_num, offset) {
                     Ok(Some(mut log_batch)) => {
-                        let entries_size = log_batch.entries_size();
+                        let mut encoded_size = 0;
+                        for item in &log_batch.items {
```
zhangjinpeng1987, line 169: Does this can reduce CPU usage?

hicqu: It does same thing as the old `entries_size` function.

src/memtable.rs
```diff
@@ -444,7 +441,7 @@ impl<E: Message + Clone, W: EntryExt<E>> MemTable<E, W> {
             .rev()
             .find(|e| e.file_num <= latest_rewrite);
         if let (Some(begin), Some(end)) = (begin, end) {
-            if begin.index < end.index {
+            if begin.index <= end.index {
```
Fullstop000, line 444: What's the corner case of using `<=` instead of `<`

Fullstop000 (approved):

src/memtable.rs
hicqu: If `begin.index == end.index`, there is one and only one log entry in the append queue. If we don't rewrite it, the append queue can't be purged on time. I meet the bug yesterday.

PR #49, opened 2020-09-24
Refactor evict to avoid reading  pipelog

Little-Wallace:
Signed-off-by: Little-Wallace <bupt2013211450@gmail.com>

I collect index of entries when writing it to `PipeLog` so that we can only compact cache with these index and do not need to read `LogBatch` from `PipeLog` any more.


Little-Wallace:
This refactor may cause too large memory use for cache. We will discuss it in future.

PR #54, opened 2020-09-27
clean regions from rewrite queue

hicqu:

It's possible for a region:
1. rewrite some entries and kvs into the rewrite queue
2. then the region has been removed from the engine
3. The log contains the `Command::Clean` for the region has been purged.
4. The regine restarts, the removed region appears again because there are some entires in the rewrite queue.

This PR resolves it. The `Command::Clean` will be written into the rewrite queue before the append queue has been purged.

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -145,7 +145,10 @@ where
                     }
                 }
                 LogItemContent::Command(Command::Clean) => {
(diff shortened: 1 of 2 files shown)


Little-Wallace (approved):
LGTM

PR #55, opened 2020-09-28
squeeze the rewrite queue if necessary

hicqu:
Previously the rewrite queue's size can only grow but not decrease even if content in it are compacted.
This PR GCs the rewrite queue if necessary, so that it's size won't grow too large.

### DIFF
--- a/src/purge.rs
+++ b/src/purge.rs
@@ -61,7 +61,7 @@ where
     }
 
     pub fn purge_expired_files(&self) -> Result<Vec<u64>> {
-        let purge_mutex = match self.purge_mutex.try_lock() {
+        let _purge_mutex = match self.purge_mutex.try_lock() {
             Ok(context) => context,
             _ => return Ok(vec![]),
         };
@@ -78,11 +78,7 @@ where
             &mut will_force_compact,
         );
 
-        let rewrite_operations = self.global_stats.rewrite_operations();
-        let compacted_rewrite_operations = self.global_stats.compacted_rewrite_operations();
-        if compacted_rewrite_operations as f64 / rewrite_operations as f64 > 0.75
-            && self.needs_purge_log_files(LogQueue::Rewrite)
-        {
+        if self.rewrite_queue_needs_squeeze() && self.needs_purge_log_files(LogQueue::Rewrite) {
             self.squeeze_rewrite_queue();
         }
 
@@ -101,7 +97,6 @@ where
             info!("purged {} expired rewrite files", purged_2);
         }
 
-        drop(purge_mutex);
         Ok(will_force_compact)
     }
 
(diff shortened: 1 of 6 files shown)


src/purge.rs
```diff
@@ -66,19 +78,46 @@ where
...
         );
 
-        let min_file_num = self.memtables.fold(u64::MAX, |min, t| {
-            cmp::min(min, t.min_file_num(LogQueue::Append).unwrap_or(u64::MAX))
-        });
-        let purged = self.pipe_log.purge_to(LogQueue::Append, min_file_num)?;
-        info!("purged {} expired log files", purged);
+        let rewrite_operations = self.global_stats.rewrite_operations();
+        let compacted_rewrite_operations = self.global_stats.compacted_rewrite_operations();
+        if compacted_rewrite_operations as f64 / rewrite_operations as f64 > 0.75
```
Little-Wallace, line 83: How about let it be a method?

src/purge.rs
```diff
@@ -66,19 +78,46 @@ where
...
+        if active_file_num == first_file_num {
+            return false;
+        }
+
+        let total_size = self.pipe_log.total_size(queue);
         let purge_threshold = self.cfg.purge_threshold.0 as usize;
-        total_size > purge_threshold
+        match queue {
+            LogQueue::Append => total_size > purge_threshold,
+            LogQueue::Rewrite => total_size * 10 > purge_threshold,
```
Little-Wallace, line 119: It is two small.... This means that we only allow 10% of data is rewritten? Maybe we could do some test and then adjust this ratio.

src/purge.rs
```diff
@@ -66,19 +78,46 @@ where
...
+                });
+        let purged_1 = self.pipe_log.purge_to(LogQueue::Append, min_file_1)?;
+        info!("purged {} expired log files", purged_1);
 
+        if min_file_2 < self.pipe_log.active_file_num(LogQueue::Rewrite) {
+            let purged_2 = self.pipe_log.purge_to(LogQueue::Rewrite, min_file_2)?;
+            info!("purged {} expired rewrite files", purged_2);
+        }
+
+        drop(purge_mutex);
```
Little-Wallace, line 104: This drop is unnecessary.

src/purge.rs
```diff
@@ -194,6 +238,35 @@ where
             }
         }
     }
+
+    fn squeeze_rewrite_queue(&self) {
+        let memtables = self
+            .memtables
+            .collect(|t| t.min_file_num(LogQueue::Rewrite).unwrap_or_default() > 0);
```
Little-Wallace, line 245: It means that all of data in `Rewrite` queue will be rewritten again? I do not think it is a  good idea.... At least most of them may be  in current write file and they shall not be rewritten. How about only rewriting the oldest file? Because we only allowed 10% of rewrite queue keep, the wal file may not exceed 2(both of them are 128MB size). So let we just rewrite the oldest file and keep the current file.

Little-Wallace:
We shall check whether the file num of files is continuous when open files for LogManager.

src/purge.rs
hicqu: It's used to reduce the frequence of squeeze the rewrite queue. The rewrite queue can only be squeezed when the garbage ratio is greater than 75%, and the total log file size of the rewrite queue reaches 1GB (by default).

Little-Wallace (approved):
LGTM

PR #64, opened 2020-10-26
code refactor about purge and rewrite

hicqu:

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -21,7 +21,7 @@ jobs:
         uses: actions-rs/toolchain@v1
         with:
           profile: minimal
-          toolchain: nightly-2020-07-01
+          toolchain: nightly-2020-10-19
           override: true
           components: clippy
       - name: Clippy
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 Cargo.lock
 target
+append-compact-purge-data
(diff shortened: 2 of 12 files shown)


Little-Wallace:
Could we split the compression from `LogBatch`? And we can encode the data when put any new item. Just like this:
```
struct LogBatch{
    items: Vec<LogItem>,
    data: Vec<u8>,
}

impl LogBatch {
      fn put(&mut self, entries: Vec<Entry>) {
           let item = LogItem::Entries(entries);
           item.encode_to(&mut self.data);
           self.items.push(item);
      }
     fn data(&self) -> &[u8] {
          &self.data
     }
}

impl PipeLog {
         fn write<E: Message, W: EntryExt<E>>(
        &self,
        data: &[u8],
        mut sync: bool,
    ) -> Result<(usize,usize, Option<CacheTracker)> {
         if data.len() > self.compression_threshold {
            let compress_data = lz4::encode_block(data, HEADER_LEN);
            self.append(LogQueue::Append, &compress_data, &mut sync)?
         }  else {
            self.append(LogQueue::Append, &data, &mut sync)?
         }
        .....
    }
}




```

Little-Wallace:
And we do not need to store compression type in `EntryIndex`. Because we must read the data from file at first, we can judge whether this data has been compressed by `HEADER`.

Little-Wallace (approved):
LGTM

PR #66, opened 2020-11-02
truncate and sync active log correctly

hicqu:
Without the patch `truncate_log_file` will do nothing for switching active log file. This PR fixes it.


Little-Wallace (approved):
LGTM

PR #74, opened 2021-06-10
Refine PipeLog abstraction

tabokie:

Refining `PipeLog` trait, in order for the ease of implementing more complex logging features, such as asynchronous IO, sharded WAL.

Several changes in this PR:
- Hiding the u64 file number and the validation logic behind `FileId`, ~and introducing the concept of `FileCheckpoint`. Unlike before, `FileId` only implements `PartialOrd`, and can't be used to establish a global time point. I'm abstracting them this way specifically to better implement the multi-threaded logger, but whether to go down this direction is yet to be decided (based on prototype performance, etc). So the necessity of this change is open for debate. An alternative would be only keeping the `FileId` abstraction and giving it total order.~ which I think makes the file-coupled code slightly cleaner.
- Remove direct member accessors like `cache_submitor()` and `hooks()` from `PipeLog`
- Code refactor

Most implementation details are unchanged, except for:
- `removed_memtables` is now `HashMap` instead of `BinaryHeap`


Connor1996 (commented):
why need to abstract the file number

tabokie:
> why need to abstract the file number

Originally I abstracted it because sharded logger has multiple file streams, which means there is no total order between file ids. But as it turns out the ideal abstraction for sharded logger is too cumbersome. So I revert some of the changes (file checkpoint).

I agree it's not really needed to abstract the `FileId` now. But I do insist in wrapping it behind an interface (think `Fd`), before all the smelly codes like `num == 0`, `num1 < num2` further contaminates this codebase.

I'll convert the trait into struct later.

MrCroxx:
LGTM

MrCroxx (approved):

src/file_pipe_log.rs
```diff
@@ -0,0 +1,959 @@
...
+                warn!("Remove purged log file {:?} fail: {}", path, e);
+                return Ok(i);
+            }
+            cur_file_id = cur_file_id.forward(1);
+        }
+        Ok(purge_count)
+    }
+}
+
+fn generate_file_name(file_num: FileId, suffix: &'static str) -> String {
```
MrCroxx, line 627: Maybe replace `file_num` with `file_id` is better to be consistent with the other interfaces.

src/file_pipe_log.rs
```diff
@@ -0,0 +1,959 @@
...
+            FILE_MAGIC_HEADER.len() + VERSION.len()
+        );
+        let trunc_big_offset = pipe_log.truncate_active_log(
+            queue,
+            Some(FILE_MAGIC_HEADER.len() + VERSION.len() + s_content.len()),
+        );
+        assert!(trunc_big_offset.is_err());
+
+        /**************************
+        // read next file
```
hicqu, line 850: Seems it  isn't commentted in the origin file?

tabokie: It is, see https://github.com/tikv/raft-engine/blob/master/src/pipe_log.rs#L857-L865

PR #80, opened 2021-07-29
Change `version` format to u64

MrCroxx:
Previously we use [u8] to define version. It's hard to make sure everything is right (see [raft-engine#79](https://github.com/tikv/raft-engine/pull/79)).

For now RaftEngine is not GA, it's better change the version format to u64.

**IMPORTANT: Make sure that there MUST BE NO user using RaftEngine.**

**This patch is not compatibility with the previous log file foramt!!!**

### DIFF
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -1,9 +1,19 @@
 use std::error;
+use std::fmt::Display;
 use std::io::Error as IoError;
 
 use thiserror::Error;
 
 use crate::codec::Error as CodecError;
+
+#[derive(Debug, Error)]
+pub struct Corruption(pub Box<dyn error::Error + Send + Sync>);
+
+impl Display for Corruption {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", self.0)
+    }
+}
 
 #[derive(Debug, Error)]
 pub enum Error {
@@ -27,10 +37,8 @@ pub enum Error {
     StorageUnavailable,
     #[error("The entry acquired has been compacted")]
     StorageCompacted,
-    #[error("Unrecognized log file version: {0}")]
-    UnrecognizedLogFileVersion(u64),
-    #[error("Unrecognized log file magic header")]
-    UnrecognizedLogFileMagicHeader,
+    #[error("Corruption: {0}")]
+    Corruption(#[from] Corruption),
 }
 
 pub type Result<T> = ::std::result::Result<T, Error>;
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -27,6 +27,10 @@ pub enum Error {
     StorageUnavailable,
     #[error("The entry acquired has been compacted")]
     StorageCompacted,
+    #[error("Unrecognized log file version: {0}")]
+    UnrecognizedLogFileVersion(u64),
+    #[error("Unrecognized log file magic header")]
+    UnrecognizedLogFileMagicHeader,
 }
 
 pub type Result<T> = ::std::result::Result<T, Error>;
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -1,19 +1,9 @@
 use std::error;
-use std::fmt::Display;
 use std::io::Error as IoError;
 
 use thiserror::Error;
 
 use crate::codec::Error as CodecError;
-
-#[derive(Debug, Error)]
-pub struct Corruption(pub Box<dyn error::Error + Send + Sync>);
-
-impl Display for Corruption {
-    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        write!(f, "{}", self.0)
-    }
-}
 
 #[derive(Debug, Error)]
 pub enum Error {
@@ -38,7 +28,7 @@ pub enum Error {
     #[error("The entry acquired has been compacted")]
     StorageCompacted,
     #[error("Corruption: {0}")]
-    Corruption(#[from] Corruption),
+    Corruption(String),
 }
 
 pub type Result<T> = ::std::result::Result<T, Error>;
--- a/src/file_pipe_log.rs
+++ b/src/file_pipe_log.rs
@@ -58,15 +58,15 @@ impl Version {
         *self as u64
     }
 
-    fn from_u64(v: u64) -> Result<Self> {
+    fn from_u64(v: u64) -> Option<Self> {
         match v {
-            1 => Ok(Self::V1),
-            _ => Err(Error::UnrecognizedLogFileVersion(v)),
+            1 => Some(Self::V1),
+            _ => None,
         }
     }
 }
 
-pub struct LogFd(RawFd, Version);
+pub struct LogFd(RawFd);
 
 impl LogFd {
     fn close(&self) -> Result<()> {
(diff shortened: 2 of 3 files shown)


tabokie (commented):
Right now the version number is nothing more than another magic number. To improve, you can implement a `open_file_with_version(version) -> Result<LogFd>` and do the version check inside.

MrCroxx:
> Right now the version number is nothing more than another magic number. To improve, you can implement a `open_file_with_version(version) -> Result<LogFd>` and do the version check inside.

@tabokie I think parsing version when opening files is not appropriate. Now when RaftEngine switch to a new log file, it also use `fn open_active_file` but there's no data yet. IMO, version parsing should be done when actually reading a file.

tabokie:
@MrCroxx I think during `LogFd` initialization the version should be checked and the meta info should be saved (e.g. encryption type, compression type, ..), because subsequent foreground requests directly use them without checking. In fact the `LogFd` should be treated as `BlockReader` in RocksDB, rather than a fd handle. And I don't think `open_active_file` is a blocker, you can simply rename it to `create_file`.

src/file_pipe_log.rs
```diff
@@ -470,18 +496,9 @@ impl PipeLog for FilePipeLog {
     ) -> Result<()> {
         let content = self.read_file_bytes(queue, file_id)?;
```
tabokie, line 497: Leave a TODO in `fn read_file_bytes` to use `LogFd` for reading file bytes instead of std::fs?

src/file_pipe_log.rs
```diff
@@ -35,12 +36,37 @@ const REWRITE_NAME_LEN: usize = REWRITE_NUM_LEN + REWRITE_SUFFIX_LEN;
...
+
+    fn from_u64(v: u64) -> Result<Self> {
+        match v {
+            1 => Ok(Self::V1),
+            _ => Err(Error::UnrecognizedLogFileVersion(v)),
+        }
+    }
+}
+
+pub struct LogFd(RawFd, Version);
```
tabokie, line 69: No need to hold `Version`, in the future `Version` can translate to a set of concrete flags that can be stored here.

src/file_pipe_log.rs
```diff
@@ -144,8 +170,8 @@ impl LogManager {
 
         let mut path = PathBuf::from(&self.dir);
         path.push(generate_file_name(self.active_file_id, self.name_suffix));
-        let fd = Arc::new(LogFd(open_active_file(&path)?));
-        let bytes = write_file_header(fd.0)?;
+        let (fd, bytes) = create_active_file(&path)?;
+        let fd = Arc::new(fd);
 
         self.active_log_size = bytes;
         self.active_log_capacity = bytes;
```
tabokie, line 177: Ideally, we could use a `ReadableFile` for inactive files, and `RandomAccessFile` for active file, and put `active_log_size` and `active_log_capacity` inside. Maybe as a followup task.

src/file_pipe_log.rs
```diff
@@ -35,12 +36,37 @@ const REWRITE_NAME_LEN: usize = REWRITE_NUM_LEN + REWRITE_SUFFIX_LEN;
...
+
+    fn len() -> usize {
+        8
+    }
+
+    fn as_u64(&self) -> u64 {
+        *self as u64
+    }
+
+    fn from_u64(v: u64) -> Result<Self> {
```
tabokie, line 61:
```suggestion
    fn from_u64(v: u64) -> Option<Self> {
```

src/file_pipe_log.rs
```diff
@@ -634,18 +651,46 @@ fn parse_nix_error(e: nix::Error, custom: &'static str) -> Error {
...
-fn open_frozen_file<P: ?Sized + NixPath>(path: &P) -> Result<RawFd> {
+fn open_frozen_file_raw<P: ?Sized + NixPath>(path: &P) -> Result<RawFd> {
     let flags = OFlag::O_RDONLY;
     let mode = Mode::S_IRWXU;
     fcntl::open(path, flags, mode).map_err(|e| parse_nix_error(e, "open_frozen_file"))
 }
 
+fn check_version(fd: RawFd) -> Result<Version> {
+    if file_size(fd)? < FILE_MAGIC_HEADER.len() + Version::len() {
+        return Err(Error::TooShort);
```
tabokie, line 684: Let's not use this, it's reserved for `LogBatch` too short.

src/errors.rs
```diff
@@ -27,6 +27,10 @@ pub enum Error {
     StorageUnavailable,
     #[error("The entry acquired has been compacted")]
     StorageCompacted,
+    #[error("Unrecognized log file version: {0}")]
+    UnrecognizedLogFileVersion(u64),
+    #[error("Unrecognized log file magic header")]
+    UnrecognizedLogFileMagicHeader,
```
tabokie, line 33: How about `Corruption(#[from] Box<dyn error::Error + Send + Sync>),`? I feel like there are too many errors...

tabokie: And add an unit test for newly introduced error conditions.

src/file_pipe_log.rs
MrCroxx: Agree.

MrCroxx: I will leave the `RandomAccessFile` work in the next pr, this pr just refines the version format.

src/errors.rs
```diff
@@ -1,10 +1,20 @@
 use std::error;
+use std::fmt::Display;
 use std::io::Error as IoError;
 
 use thiserror::Error;
 
 use crate::codec::Error as CodecError;
 
+#[derive(Debug, Error)]
+pub struct Corruption(pub Box<dyn error::Error + Send + Sync>);
```
tabokie, line 10: Why is this extra abstraction needed?

src/file_pipe_log.rs
```diff
@@ -832,4 +885,41 @@ mod tests {
...
+        let mut buf: Vec<u8> = Vec::with_capacity(1024);
+        buf.write(FILE_MAGIC_HEADER).unwrap();
+        buf.encode_u64(u64::MAX).unwrap();
+        let fd = test_write_tmp_file(&buf);
+        assert!(check_version(fd).is_err());
+
+        // correct
+        let mut buf: Vec<u8> = Vec::with_capacity(1024);
+        buf.write(FILE_MAGIC_HEADER).unwrap();
+        buf.encode_u64(Version::current().as_u64()).unwrap();
```
tabokie, line 921: Call `write_file_header` instead?

src/file_pipe_log.rs
```diff
@@ -832,4 +885,41 @@ mod tests {
     fn test_pipe_log_rewrite() {
         test_pipe_log_impl(LogQueue::Rewrite)
     }
+
+    fn test_write_tmp_file(buf: &[u8]) -> RawFd {
```
tabokie, line 889:
```suggestion
    fn create_file_with_content(name: &str, buf: &[u8]) -> RawFd {
```

src/file_pipe_log.rs
```diff
@@ -832,4 +885,41 @@ mod tests {
...
+            .map_err(|e| parse_nix_error(e, "open_active_file"))
+            .unwrap();
+        pwrite(fd, buf, 0).unwrap();
+        fd
+    }
+
+    #[test]
+    fn test_check_version() {
+        // magic header corruption
+        let mut buf: Vec<u8> = Vec::with_capacity(1024);
```
tabokie, line 904:
```suggestion
        let mut buf = Vec::new();
```

src/file_pipe_log.rs
```diff
@@ -832,4 +885,41 @@ mod tests {
...
+        buf.write(b"RAFT-LOG-FILE-HEADER-********************************")
+            .unwrap();
+        buf.encode_u64(Version::current().as_u64()).unwrap();
+        let fd = test_write_tmp_file(&buf);
+        assert!(check_version(fd).is_err());
+
+        // unrecognized version
+        let mut buf: Vec<u8> = Vec::with_capacity(1024);
```
tabokie, line 912:
```suggestion
        buf.clear();
```

src/errors.rs
MrCroxx: There's already `Other(#[from] Box<dyn error::Error + Send + Sync>)` in `Error`. Without another abstraction the errors created by `box_err!` cannot decide which item to convert into.

tabokie: Then let's replace box_err with `String`

tabokie (approved):
Rest looks good.

PR #81, opened 2021-08-01
Support hint files for fast recovery.

MrCroxx:
`Hint` is a special kind of log file which only save the indexes
of log entries that refer to the real log file. When recovering
RaftEngine, needs to restore memtable which maintains all log
indexes. Previously, RaftEngine had to scan all of the log files
which is slow and reading actuall data is unneeded.

This implemenation constructs and flushes hint file in a background
thread.

TODO:
- [x] Implement hint worker and tasks.
- [x] Implement hint file format.
- [x] Save to & restore from file.
- [x] Refine hint code. Not modifying `PipeLog` trait.
- [x] Delete hint file when dropping log file.
- [ ] Unittests.
- [ ] Benchmark.

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -381,7 +381,6 @@ where
         let mut batches = Vec::new();
         let mut file_id = first_file;
         while file_id <= active_file {
-            // TODO(MrCroxx): dir
             match hint_manager.read_hint_file(queue, file_id) {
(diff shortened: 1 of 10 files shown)


MrCroxx:
New log file format works well. There is no need of hint now.

PR #82, opened 2021-08-04
add merge and approximate_size for LogBatch

tabokie:

### DIFF
--- a/src/log_batch.rs
+++ b/src/log_batch.rs
@@ -655,7 +655,11 @@ where
     }
 
     pub fn approximate_size(&self) -> usize {
-        self.items_size + 20
+        if self.items.is_empty() {
+            0
+        } else {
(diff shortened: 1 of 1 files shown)


src/log_batch.rs
```diff
@@ -616,8 +654,8 @@ where
         Some(vec)
     }
 
-    pub fn entries_size(&self) -> usize {
-        self.entries_size
+    pub fn approximate_size(&self) -> usize {
+        self.items_size + 20
```
gengliqi, line 658: What's the meaning of 20?

tabokie: The log batch layout is `{ 8 bytes len | item count | multiple items | 4 bytes checksum }`

gengliqi (approved):

PR #83, opened 2021-08-05
Optimize log file format for faster recovery

MrCroxx:
The new log file format can be found in annotation in
`LogBatch::encode_to_bytes`.

What's changed?
+ Optimize log file format for faster recovery.
+ Refine `EntryIndex` fields in memtable.

TODO:
+ [x] There are 2 [PLEASE REVIEW] todo(s).
+ [x] Optimize `read_file` for the new log format.
+ [x] Benchmark


MrCroxx:
@tabokie I found that `LogBatch::entries_size` was replaced with `LogBatch::approximate_size` in [raft-engine#82](https://github.com/tikv/raft-engine/pull/82), but it is not used for now. I'm confused why we need the the field? Is it prepared for metric?

src/memtable.rs
```diff
@@ -210,7 +215,8 @@ impl<E: Message + Clone, W: EntryExt<E>> MemTable<E, W> {
         let first = cmp::max(entries_index[0].index as usize, front);
         let last = cmp::min(entries_index[ents_len - 1].index as usize, back);
 
-        if last < front {
+        // TODO(MrCroxx): [PLEASE REVIEW] was this a bug?
+        if last < first {
```
MrCroxx, line 219: @hicqu PTAL, was this line a bug?

tabokie: It is, it can cause usize underflow below. You can add an unit test for it.

src/log_batch.rs
```diff
@@ -480,7 +479,7 @@ where
     W: EntryExt<E>,
 {
     pub items: Vec<LogItem<E>>,
-    items_size: usize,
+    approximate_size: usize,
```
tabokie, line 482: Let's not drop the `items_` prefix, you can rename it to `items_approximate_size` if you like.

src/log_batch.rs
```diff
@@ -79,92 +80,92 @@ impl<M: MessageExt> Entries<M> {
...
-        // entries layout = { entry layout | ... | entry layout }
-        // entry layout = { len | entry content }
-        vec.encode_var_u64(self.entries.len() as u64)?;
+    pub fn encode_to(
+        &mut self,
+        buf: &mut Vec<u8>,
+        entries_buf: &mut Vec<u8>,
+        entries_size: &mut usize,
+    ) -> Result<()> {
+        // buf layout = { entries count | [ entry_index | entry_tail_offset ] }
```
tabokie, line 108: Only need to encode the first index, indexes in one `Entries` are always contiguous

MrCroxx: Ohh, I forgot that. Let me add an `assert` here to make sure of it.

src/log_batch.rs
```diff
@@ -550,50 +542,90 @@ where
             return None;
         }
 
-        // layout = { 8 bytes len | item count | multiple items | 4 bytes checksum }
-        let mut vec = Vec::with_capacity(4096);
-        let mut entries_size = 0;
-        vec.encode_u64(0).unwrap();
-        vec.encode_var_u64(self.items.len() as u64).unwrap();
+        // layout = { 8 bytes batch_len | 8 bytes section offset | item count | multiple items (without entries) | crc32 | section (entries) | crc32 }
```
tabokie, line 545:
```suggestion
        // Log Batch Layout:
        // <---    header     ---> <---       entries        ---> <---                   footer                    --->
        // { u32 len | u32 offset | (compressed) entries | crc32 | item count | entry indexes / kvs / commands | crc32 }
        //                    `------------------------------------^
```

src/reader.rs
```diff
@@ -0,0 +1,125 @@
+// Copyright 2021 TiKV Project Authors. Licensed under Apache-2.0.
+
+use crate::file_pipe_log::LogFd;
+use crate::{Error, LogBatch, MessageExt, Result};
+
+use log::trace;
+
+#[derive(Debug)]
+pub struct LogBatchFileReader<'a> {
```
tabokie, line 9: Why read the full batch? You only need to read `(index, EntryIndex)` pair for recovery, that's why recovery can be faster with new log format.

src/log_batch.rs
tabokie: To avoid using two buffers, you can encode entries first (and update `EntryIndex` struct), then encode entry indexes and other items in the second pass.

tabokie: BTW, during recovery, it's more economical to read the footer of current batch and the header of next batch together.

src/reader.rs
MrCroxx: `LogBatchFileReader` does not read the full batch. Its read size is based on the max value of the entries_index size and a configurable `recovery_read_block_size` (see next pr), and caches if needed. This will reduce read times if log batches are small.

src/log_batch.rs
MrCroxx: > To avoid using two buffers, you can encode entries first (and update EntryIndex struct), then encode entry indexes and other items in the second pass.

Yep, that's my first design. I didn't implement like that becase when recovering a log batch, we need to read the header part and the entries_index part. That design will make utilizing cache a little bit completed before (before I implemented `LogBatchFileReader`).

tabokie: Not sure I understand, but "read the footer of current batch and the header of next batch together" can is as cache friendly as current implementation.

Cargo.toml
```diff
@@ -48,3 +51,8 @@ protobuf-codegen = { git = "https://github.com/pingcap/rust-protobuf", rev = "82
 
 [workspace]
 members = [ "stress" ]
+
+[[bench]]
+name = "bench_recovery"
+harness = false
+required-features = ["fail/failpoints"]
```
tabokie, line 58: maybe ["failpoints"]?

benches/bench_recovery.rs
```diff
@@ -0,0 +1,185 @@
...
+            index = entries.iter_mut().fold(index, |index, e| {
+                e.index = index + 1;
+                index + 1
+            });
+            *indexes.get_mut(&region_id).unwrap() = index;
+            batch.add_entries(region_id, entries);
+        }
+        engine.write(&mut batch, false).unwrap();
+    }
+    drop(engine);
```
tabokie, line 99: sync before drop?

benches/bench_recovery.rs
```diff
@@ -0,0 +1,185 @@
...
+fn dir_size(path: &str) -> u128 {
+    let mut size: u128 = 0;
+    for entry in std::fs::read_dir(PathBuf::from(path)).unwrap() {
+        let entry = entry.unwrap();
+        size += std::fs::metadata(entry.path()).unwrap().len() as u128;
+    }
+    size
+}
+
+fn size_to_readable_string(size: u128) -> String {
```
tabokie, line 112: Why not use `util::ReadableSize`?

tabokie (commented):
Good work, let's see some results! BTW You can put the benchmark into a separate PR.

benches/bench_recovery.rs
MrCroxx: It seems I must import `serde_json` to serialize it? Can I do that just use the core of `serde`?

MrCroxx: Oh, forget it. I just found a method.

src/log_batch.rs
```diff
@@ -485,115 +472,128 @@ where
...
             0
         } else {
-            8 /*len*/ + 8 /*count*/ + self.items_size + 4 /*checksum*/
+            8 /*len*/ + 8 /*section offset*/ + 8/*items count*/
+            +self.items_approximate_size + CHECKSUM_LEN * 2 /*checksum*/
         }
     }
+
+    /// min buffer size for recover memtable ()
+    pub fn recovery_size(buf: &mut SliceReader<'_>) -> Result<usize> {
```
tabokie, line 587: Not used?

src/log_batch.rs
```diff
@@ -485,115 +472,128 @@ where
...
 
-        // layout = { 8 bytes len | item count | multiple items | 4 bytes checksum }
-        let mut vec = Vec::with_capacity(4096);
-        let mut entries_size = 0;
-        vec.encode_u64(0).unwrap();
-        vec.encode_var_u64(self.items.len() as u64).unwrap();
+        // layout:
+        // <---     header     ---><---        entries        ---><---                    footer                    --->
+        // { u64 len | u64 offset | (compressed) entries | crc32 | item count | entry indexes / commands / kvs | crc32 }
+        // ^ offset = 0                                         ^ offset = <offset>
```
tabokie, line 524:
```suggestion
        // +----------------------------------------------------^ offset
```

src/log_batch.rs
```diff
@@ -616,28 +616,68 @@ pub fn decompress(buf: &[u8]) -> Vec<u8> {
 #[cfg(test)]
 mod tests {
     use super::*;
+
+    use protobuf::parse_from_bytes;
     use raft::eraftpb::Entry;
 
+    fn entries(first_index: u64, len: usize, data: Option<Vec<u8>>) -> Vec<Entry> {
```
tabokie, line 623: similar to `generate_entry_indexes`? How about sharing it in a `test_util` file?

src/log_batch.rs
```diff
@@ -485,115 +472,128 @@ where
...
-            let content_offset = (content_len - reader.len()) as u64;
-            let item = LogItem::from_bytes(
-                &mut reader,
-                file_id,
-                base_offset,
-                content_offset,
-                &mut entries_size,
-            )?;
-            log_batch.items_size += item.compute_size();
+    pub fn from_footer(
```
tabokie, line 484: I feel like we shouldn't allow partially filled `LogBatch`.
Could we do this instead: `LogBatch = (Vec<Entry>, LogItemBatch), LogItemBatch = (Vec<LogItem>), LogItemContent::Entries = (Vec<EntryIndex>)`, so here `from_footer` can belong to `LogItemBatch`.

src/file_pipe_log.rs
```diff
@@ -599,44 +604,30 @@ impl PipeLog for FilePipeLog {
...
-        path.push(build_file_name(queue, file_id));
-        // TODO: use `LogFd` for reading file bytes instead of `std::fs`
-        let content = fs::read(&path)?;
-        let mut buf = content.as_slice();
-        let start_ptr = buf.as_ptr();
-        buf.consume(FILE_MAGIC_HEADER.len() + Version::len());
-        let mut offset = (FILE_MAGIC_HEADER.len() + Version::len()) as u64;
+        debug!("recover from log file {:?}:{:?}", queue, file_id);
+        let fd = self.get_queue(queue).get_fd(file_id)?;
+        let mut reader = LogBatchFileReader::new(
```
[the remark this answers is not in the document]
MrCroxx: That's a good idea. But now some function like `get_queue` is only implemented for `FilePipeLog`, but `recover_queue` is implemented for the trait `PepeLog`. Is that fine that we implement a `recover_queue` only for `FilePipeLog` version?

src/reader.rs
```diff
@@ -0,0 +1,133 @@
...
+            self.buffer.extend(self.fd.read(offset as i64, len)?);
+        } else {
+            trace!("::replace buffer {}:{}", offset, len);
+            self.buffer = self.fd.read(offset as i64, len)?;
+            self.offset = offset;
+        }
+        self.len = self.buffer.len();
+        trace!("new buffer {}:{}", self.offset, self.len);
+        Ok(())
+    }
```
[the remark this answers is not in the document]
MrCroxx: I don't think that's a good idea. These four function has different purpose, and some of them are reused multiple times. I should add some comments for docs for these function to make them more understandable.

src/log_batch.rs
MrCroxx: It's from the first version of the new log foramt. I forgot to remove them.

src/reader.rs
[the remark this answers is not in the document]
tabokie: They are reused only because you leak out the details of reading data to buffer. DRY won't be violated If you merge them into `peek`, instead the code is much cleaner because I can see how all the internal states are mutated in one single function.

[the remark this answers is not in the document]
MrCroxx: I get you. Let me fix it.

src/file_pipe_log.rs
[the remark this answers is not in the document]
tabokie: Indeed, ideally we shouldn't expose log file. Maybe we could replace  `read_file_into_log_batch` with `replay_log_batch`, which takes in a closure `Fn(LogBatch)`.

src/reader.rs
```diff
@@ -0,0 +1,114 @@
...
+
+        let roffset = std::cmp::max(self.offset + self.buffer.len() as u64, self.cursor);
+        let rsize = std::cmp::min(
+            std::cmp::max(size + hint, self.read_block_size),
+            self.fsize - roffset as usize,
+        );
+
+        if roffset == self.offset + self.buffer.len() as u64 {
+            trace!("::extend buffer {}:{}", roffset, rsize);
+            self.buffer.extend(self.fd.read(roffset as i64, rsize)?);
```
tabokie, line 90: Why not use `read_to` as well?

src/reader.rs
```diff
@@ -0,0 +1,114 @@
...
+            entries_len,
+            compression_type,
+        )?;
+        self.cursor += footer_size as u64;
+        Ok(Some(item_batch))
+    }
+
+    fn peek(&mut self, size: usize, hint: usize) -> Result<()> {
+        let remain = (self.offset as usize + self.buffer.len())
+            .checked_sub(self.cursor as usize)
```
tabokie, line 76: saturating_sub?

src/reader.rs
```diff
@@ -0,0 +1,114 @@
...
+            &mut self.slice(footer_size),
+            entries_offset,
+            entries_len,
+            compression_type,
+        )?;
+        self.cursor += footer_size as u64;
+        Ok(Some(item_batch))
+    }
+
+    fn peek(&mut self, size: usize, hint: usize) -> Result<()> {
```
tabokie, line 74: I thought `peek` was supposed to return a slice?

src/log_batch.rs
```diff
@@ -392,208 +332,363 @@ impl<M: MessageExt> LogItem<M> {
...
     // Don't account for compression and varint encoding.
+    // Not meaningful after recovery.
     pub fn approximate_size(&self) -> usize {
-        if self.items.is_empty() {
+        if self.item_batch.is_empty() {
             0
         } else {
-            8 /*len*/ + 8 /*count*/ + self.items_size + 4 /*checksum*/
+            8 /*len*/ + 8 /*offset*/ + 8 /*items count*/
+            + self.content_approximate_size + CHECKSUM_LEN *2
```
tabokie, line 686: Looks like the code isn't formatted?

src/log_batch.rs
```diff
@@ -392,208 +332,363 @@ impl<M: MessageExt> LogItem<M> {
...
 
-    pub fn drain(&mut self) -> std::vec::Drain<'_, LogItem<M>> {
-        self.items_size = 0;
+    pub fn drain(&mut self) -> std::vec::Drain<'_, LogItem> {
         self.items.drain(..)
     }
 
+    pub fn is_empty(&self) -> bool {
+        self.items.is_empty()
+    }
```
tabokie, line 387: I think these can be removed, we are implementing them for `LogBatch` to hide data members from other modules.

src/log_batch.rs
```diff
@@ -392,208 +332,363 @@ impl<M: MessageExt> LogItem<M> {
...
+        buf: &mut Vec<u8>,
+        entries_size: &mut usize,
+    ) -> Result<Vec<EntriesIndex>> {
+        let mut entries_indexes = Vec::with_capacity(self.0.len());
+        for entries in self.0.drain(..) {
+            let entries_index = entries.encode_to(buf, entries_size)?;
+            entries_indexes.push(entries_index);
         }
+        Ok(entries_indexes)
+    }
```
tabokie, line 535: Maybe this wrapper can be removed too? Doesn't seem very necessary.

tabokie (approved):
Code changes look good overall, still needs some benchmarks to evaluate the regression of write performance and IO usage.

src/reader.rs
MrCroxx: `read_to` will resize the buffer to given `len`, and overwrite data from the begining of the buffer. `extend` is used when there is already partial data in buffer, and we just want to read the missing part.

src/log_batch.rs
MrCroxx: Weird again, my formatter doesn't report it. This happens a lot of times these days.

src/log_batch.rs
MrCroxx: Do you mean `EntriesBatch` of just these 3 methods?

src/reader.rs
tabokie: You can `resize(len() + size)`, the purpose of adding a `read_to` is to avoid one unnecessary allocation at this line.

src/log_batch.rs
tabokie: I mean remove `EntriesBatch` and directly store `Vec` in `LogBatch`.

src/file_pipe_log.rs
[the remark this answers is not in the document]
MrCroxx: In the next pr I'll implement parallel recovery. I'm not sure if the modification is good for it. What about modify it in the next pr or a seperate pr?

src/reader.rs
```diff
@@ -0,0 +1,114 @@
...
+        let entries_offset = self.cursor + HEADER_LEN as u64;
+        let entries_len = offset as usize - HEADER_LEN;
+
+        // skip entries
+        self.cursor += offset;
+
+        // read footer & recover
+        let footer_size = len - offset as usize;
+        trace!("LogBatch::recovery_size = {}", footer_size);
+        self.peek(footer_size, HEADER_LEN)?;
```
tabokie, line 57: can remove

src/reader.rs
```diff
@@ -0,0 +1,114 @@
...
+        if (self.offset as usize + self.buffer.len() - self.cursor as usize) < size {
+            return Err(Error::Corruption(format!(
+                "unexpected eof at {}",
+                self.offset + self.buffer.len() as u64
+            )));
+        }
+        Ok(self.slice(size))
+    }
+
+    fn slice(&self, len: usize) -> &[u8] {
```
tabokie, line 107: can remove as well

MrCroxx:
**Benchmark (recovery):**

configs:

| config | 0 - default | 1 - compressed | 2 - small-batch | 3 - 10GB |
| --- | --- | --- | --- | --- |
| total size | 1GB | 1GB | 1GB | 10GB |
| region count | 100 | 100 | 100 | 1000 |
| batch size | 1MB | 1MB | 1KB | 1MB |
| item size | 1KB | 1KB | 256B | 1KB |
| entry size | 256B | 256B | 32B | 256B |
| compression threshold | - | 8KB | - | - |

*env: slow ssd, fadvise - DONTNEED has been called before reads to disable page cache.*

| config | 0 - default | 1 - compressed | 2 - small-batch | 3 - 10GB |
| --- | --- | --- | --- | --- |
| old | 5.3239s | 5.7234s | 17.207s | 64.410s |
| new | 1.1660s | 1.2687s | 7.5835s | 15.332s |
| | -78.1% | -77.8% | -55.9% | -76.2% |

src/file_pipe_log.rs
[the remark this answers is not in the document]
tabokie: Actually I have a big PR coming up tomorrow, and will fix this issue there.

MrCroxx:
**Benchmark (stress):**

old:
```
--compression-threshold 8KB --regions 1000 --write-entry-count 10
--write-ops-per-thread 10 --write-region-count 10 --write-sync true --write-threads 100 --time 600
[write]
Throughput(QPS) = 998.78
Latency(μs) min = 208, avg = 23671.98, p50 = 18719, p90 = 48607, p95 = 58911, p99 = 84415, p99.9 = 122367, max = 187007
Fairness = 100.0%
```

new:
```
--compression-threshold 8KB --regions 1000 --write-entry-count
 10 --write-ops-per-thread 10 --write-region-count 10 --write-sync true --write-threads 100 --time 600
[write]
Throughput(QPS) = 999.37
Latency(μs) min = 210, avg = 19851.76, p50 = 16847, p90 = 39871, p95 = 48639, p99 = 68351, p99.9 = 95551, max = 158719
Fairness = 100.0%
```

No obvious influence on fore-end writes.

MrCroxx:
@tabokie Is this pr ready to merge?

MrCroxx:
**Benchmark (recovery, limited by batch count):**

configs:

| config | 0 - default | 1 - compressed | 3 - 10k |
| --- | --- | --- | --- | --- |
| total batch count | 1000 | 1000 |  10000 |
| region count | 100 | 100 | 1000 |
| batch size | 1MB | 1MB | 1MB |
| item size | 1KB | 1KB | 1KB |
| entry size | 256B | 256B | 256B |
| compression threshold | - | 8KB | - |

*env: slow ssd, fadvise - DONTNEED has been called before reads to disable page cache.*

| config | 0 - default | 1 - compressed | 2 - small-batch | 3 - 10k |
| --- | --- | --- | --- | --- |
| old | 4.1913s | 4.3947s | 50.943s |
| new | 0.9803s | 1.0350s | 11.382s |
| | -77.1% | -76.4% | -77.7% |

PR #87, opened 2021-08-12
Add bench case for recovery.

MrCroxx:
Add recovery benchmarking.

use `cargo bench --features=failpoints` to run `bench_recovery`.

To get the complete output, you may want to run:
```bash
cargo bench --bench bench_recovery --features="failpoints" -- --verbose --nocapture
```

### DIFF
--- a/benches/bench_recovery.rs
+++ b/benches/bench_recovery.rs
@@ -1,4 +1,5 @@
-use byte_unit::Byte;
+// Copyright (c) 2017-present, PingCAP, Inc. Licensed under Apache-2.0.
+
 use criterion::{criterion_group, criterion_main, BenchmarkId, Criterion};
 use raft::eraftpb::Entry;
 use raft_engine::ReadableSize;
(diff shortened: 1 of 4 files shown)


benches/bench_recovery.rs
```diff
@@ -0,0 +1,176 @@
...
+impl fmt::Display for Config {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{} [region-count: {}][batch-size: {}][item-size: {}][entry-size: {}][batch-compression-threshold: {}]",
+        self.total_size,
+        self.region_count,
+        self.batch_size,
+        self.item_size,
+        self.entry_size,
+        self.batch_compression_threshold
+    )
```
tabokie, line 58:
```suggestion
        write!(f, "{} [region-count: {}][batch-size: {}][item-size: {}][entry-size: {}][batch-compression-threshold: {}]",
            self.total_size,
            self.region_count,
            self.batch_size,
            self.item_size,
            self.entry_size,
            self.batch_compression_threshold
        )
```

tabokie (approved):
LGTM

benches/bench_recovery.rs
MrCroxx: It weird that my formatter tells no difference on both.

PR #88, opened 2021-08-13
Fix build with older version on some platform.

MrCroxx:
With `nightly-2021-07-28-x86_64-unknown-linux-gnu(rustc 1.56.0-nightly,
2faabf579)` on Centos, `bench_recovery` cannot compile. Add this line
will fix the problem.


tabokie (approved):
Looks good.

PR #91, opened 2021-08-20
Introduce file system and implement streaming encoding

tabokie:
Signed-off-by: tabokie <xy.tao@outlook.com>

PR #96, opened 2021-09-07
replace file system with file builder

tabokie:

Replace file system(#91) with an Allocator-like type `FileBuilder`.

File builder is better, it's strongly typed with our own file writer/reader, so that 1) we don't need to propagate `Send` and `Sync` trait bound outside of this library, 2) user writers or readers are promised to use our writer/reader as a backend.

There still are some reasonable requirements on user types, such as no buffering and no length altering, which are documented in trait definition.

Related: [TiKV commit](https://github.com/tabokie/tikv/commit/1e3ad3a9921f462ec7ec55a858fe9f9b1bb1a4f0) that implements a `FileBuilder` for encryption and rate limiting.


tabokie:
@yiwu-arbug @Connor1996 @MrCroxx Could any of you help take a look at this while you have time? The `FileBuilder` trait uses an incomplete feature https://github.com/rust-lang/rust/issues/44265 which I haven't found an alternative to work around. Maybe you have better idea.

skyzh:
> @yiwu-arbug @Connor1996 @MrCroxx Could any of you help take a look at this while you have time? The `FileBuilder` trait uses an incomplete feature [rust-lang/rust#44265](https://github.com/rust-lang/rust/issues/44265) which I haven't found an alternative to work around. Maybe you have better idea.

GAT will be stable soon. I think it's okay to use it.

src/file_builder.rs
```diff
@@ -0,0 +1,62 @@
...
+    /// bytes stream, nor buffer bytes between operations.
+    type Reader<R: Seek + Read>: Seek + Read;
+    type Writer<W: Seek + Write>: Seek + Write;
+
+    fn build_reader<R>(
+        &self,
+        path: &Path,
+        reader: R,
+    ) -> Result<Self::Reader<R>, Box<dyn std::error::Error>>
+    where
```
skyzh, line 19: Why not `reader: Self::Reader`?

src/file_builder.rs
```diff
@@ -0,0 +1,62 @@
...
+
+use std::io::{Read, Seek, Write};
+use std::path::Path;
+
+/// A `FileBuilder` generates file readers or writers that are composed upon
+/// existing ones.
+pub trait FileBuilder: Send + Sync {
+    /// Types of outcome file reader/writer. They must not alter the length of
+    /// bytes stream, nor buffer bytes between operations.
+    type Reader<R: Seek + Read>: Seek + Read;
```
skyzh, line 11: `type Reader: Seek + Read` should be enough? Then all across the file builder should use `Reader`.

tabokie:
> > @yiwu-arbug @Connor1996 @MrCroxx Could any of you help take a look at this while you have time? The `FileBuilder` trait uses an incomplete feature [rust-lang/rust#44265](https://github.com/rust-lang/rust/issues/44265) which I haven't found an alternative to work around. Maybe you have better idea.
> 
> GAT will be stable soon. I think it's okay to use it.

Good to know! Reviews are still welcome.

src/file_builder.rs
tabokie: It is enough, but then we can't guarantee the implementor will actually use `R`/`W` from typing perspective. Though I'm not sure this is the idiomatic way to contracting such requirement.

tabokie: Also, I think it would be hard for implementor to wrap a generic `R`/`W` into a non-generic `Reader` type.

Connor1996:
> @yiwu-arbug @Connor1996 @MrCroxx Could any of you help take a look at this while you have time? The `FileBuilder` trait uses an incomplete feature [rust-lang/rust#44265](https://github.com/rust-lang/rust/issues/44265) which I haven't found an alternative to work around. Maybe you have better idea.

LGTM

src/log_file.rs
```diff
@@ -116,10 +59,7 @@ impl LogFd {
     }
 
     pub fn sync(&self) -> IoResult<()> {
-        let start = Instant::now();
-        let res = fsync(self.0).map_err(|e| from_nix_error(e, "fsync"));
-        LOG_SYNC_TIME_HISTOGRAM.observe(start.saturating_elapsed().as_secs_f64());
```
Connor1996, line 121: keep the metrics?

tabokie: They are actually moved out to file_pipe_log.rs.

PR #97, opened 2021-09-08
Implement parallel recovery

MrCroxx:
+ Add config `recovery_threads` to control concurrency of recovery.
+ Implement parallel recovery & apply.
+ Update unit tests.

TODOs:
- [x] Unittests for all conditions.
- [x] Benchmark (final version).


MrCroxx:
## Parallel Recovery Benchmark
### Configuration
| config | value |
| --- | --- |
| total size | 10GB |
| region count | 1000 |
| batch size | 1MB |
| item size | 1KB |
| entry size | 256B |
| compression threshold | - |
### Benchmark
benchmark
*Note: PRSA stands for "parallel read sequential apply" referring to my own branch [concurrent-reader](https://github.com/mrcroxx/raft-engine/tree/concurrent-reader). PRPA stands for "parallel read parallel apply" which refers to this pr.*

src/memtable.rs
```diff
@@ -300,6 +300,19 @@ impl MemTable {
         }
     }
 
+    /// Mrege from `rhs`, which is in the range right-continuous to this.
+    /// Only called during parllel recovery.
+    pub fn merge(&mut self, rhs: &mut Self) {
```
tabokie, line 305:
```suggestion
    pub fn merge_newer_neighbor(&mut self, rhs: &mut Self) {
```

src/recover.rs
```diff
@@ -0,0 +1,294 @@
...
+    pub fn merge(&mut self, right: &mut Self) {
+        self.removed_memtables
+            .extend(right.removed_memtables.drain());
+        self.compacted_memtables
+            .extend(right.compacted_memtables.drain());
+        self.removed_keys.extend(right.removed_keys.drain());
+    }
+}
+
+pub struct RecoverManager<F>
```
tabokie, line 50: Still not satisfied of this arrangement. How about abstracting the memtable recover logic into this:
```
    trait ReplayMachine : Default {
        fn replay(item_batch: LogItemBatch) -> Result<()>;
        fn merge(rhs: ReplayMachine) -> Result<()>;
    }
```
And let pipe log decides how to parallelize the recovery.

MrCroxx: I agree that the current design can still be improved, but let me  explain that why I choose it.

The basic rule is to avoid `file_pipe_log` and `memtable` rely on each other or rely on `engine`. But to make recovery parallel and under good performance, the recovery worker needs to:
1. Create new `MemTableAccessor` during parallel "reduce". Rayon will pick the appropriate points to insert an new empty `MemTableAccessor` to make "reduce" parallel.
2.  Create `ParallelRecoverContext` from `LogItemBatch`. Parallel recovery needs to get tombstones from item batch to mask previous memtables.

So, I think it's necessary to introduce `recover` mod and let mod `log_batch`, `memtable`, `file_pipe_log` and `engine` rely on it rather than on each other.

I don't think introduce `SequentialReplayMachine` will eliminate the dependency problem. But define it in `recover` mod and use it to separate functions for different usages is a good idea. But still, the information that recovery needs is not enough in `MemTableAccessor` (e.g. tombstones, currently in `ParallelRecoverContext`), and I don't want `MemTableAccessor` hold date that only for recovery. I'm not sure if GAT fits this condition.

src/config.rs
```diff
@@ -5,6 +5,7 @@ use serde::{Deserialize, Serialize};
 use crate::{util::ReadableSize, Result};
 
 const MIN_RECOVERY_READ_BLOCK_SIZE: usize = 512;
+const MIN_RECOVERY_THREADS: usize = 2;
```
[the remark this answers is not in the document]
MrCroxx: Now `recovery_threads` is not the actual thread count. Speak accurately, it's how many segments we split the total log files. The min segment count is 2 because either append log or rewrite log can be split to at least one segments. 

I agree that we need to refine the config. I will make the config control the actual threads that rayon use and split log segments based on that.

src/recover.rs
tabokie: The `SequentialReplayMachine::default()` could be used to "reduce"? And the machine can holds both `MemTable` and `ParallelRecoverContext`.

src/config.rs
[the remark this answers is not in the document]
tabokie: Okay, just want to mention that threads=1 should be supported, and as a side note, Append and Rewrite should have different concurrency based on their actual size.

MrCroxx:
## Benchmark

*Call FADV_DONTNEED before reading disks. Measuerd on slow SSD.*

### Config

| config | total size | region count | batch size | item size | entry size | compression threshold |
| --- | --- | --- | --- | --- | --- | --- |
| default | 1GB | 100 | 1MB | 1KB | 256B | - |
| compressed | 1GB | 100 | 1MB | 1KB | 256B | 8KB |
| small batch | 1GB | 100 | 1KB | 256B | 64B | - |
| 10GB | 10GB | 1000 | 1MB | 1KB | 256B | - |

### Measurement

| | default | compressed | small batch | 10GB |
| --- | --- | --- | --- | --- |
| baseline | 1.0650s | 1.1071s | 5.3022s | 13.869s |
| 1-thread | 1.2436s | 1.2060s | 6.5620s | 15.395s |
| 4-thread | 702.21ms | 707.39ms | 5.2430s | 8.7042s |
| 8-thread | 630.79ms | 612.19ms | 4.2615s | 6.7961s |
| 16-thread | 547.59ms | 556.91ms | 3.9654s | 6.1792s |

src/engine.rs
```diff
@@ -1066,4 +1066,99 @@ mod tests {
         drop(engine);
         RaftLogEngine::open(cfg).unwrap();
     }
+
+    #[test]
+    fn test_parallel_recover_merge() {
```
tabokie, line 1071: Could you verify the `GlobalStats` after the merge too? I previously spotted a bug with it and fixed it.

src/engine.rs
```diff
@@ -540,7 +540,7 @@ where
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::util::ReadableSize;
+    use crate::{test_util::generate_entry_indexes, util::ReadableSize};
```
tabokie, line 543: Let's not merge them, we normally only merge objects from the same file.

Issue #101, opened 2021-09-23
Physical lock file so that multiple processes can't operate on the same db directory simultaneously


yuqi1129:
This is a subtask of #99

Issue #99, opened 2021-09-15
Add administration tool for recovery from data corruption


tabokie:
# Development Task

This task attempts to add a data administration tool similar to `ldb` offered by [RocksDB](https://github.com/facebook/rocksdb/wiki/Administration-and-Data-Access-Tool).

Raft engine is unable to start when it encounters data corruption (checksum mismatch, unexpected EOF). But user might employs data redundancy at application level. So we should allow user to manually decide how to deal with corrupted data, and recover the data files to a consistent state once the affected raft groups are properly handled.

Workload estimation: M (Two man week)

Task breakdown:

- [x] Lock file: multiple processes can't operate on the same db directory simultaneously: #102
- [ ] Dump: output operations in log files
- [x] Consistency check*: scan for holes in raft group and truncate corrupted file tails: #132 
- [ ] Unsafe recover**: recover to consistent state by modifying log files
- [x] A standalone binary that encapsulates the two methods as sub-commands: #144

\* Ideally user should be able to extract the remaining log entries for more complex recovery logic, but that is too cumbersome for this phrase.

\*\* For now, we only offer three options for unsafe recovery:
1. Remove the entire raft group in question
2. Remove log entries before a given index, making the rest of log entries in consistent state
3. Fill in empty messages to repair log entry holes

Connor1996:
Better to provide it in tikv-ctl instead of a standalone binary

yiwu-arbug:
We also need a tool to print the content of log files, preferably if it can deserialize the protobuf Entry struct.

tabokie:
@Connor1996 tikv-ctl can still use those public interfaces. If you consider raft engine as an independent project, then it's good to have a standalone binary too.
@yiwu-arbug User should be responsible for any protobuf parsing, raft engine only store raw blob data. So data access tool should also be implemented by user (tikv-ctl) using existing engine methods.

This issue mainly focuses on recovery from fatal error (log index or key-value corruption). For rocksdb, it's simply a matter of deleting corrupted SSTs, but for raft engine, things are more complicated because deletion of one log file might or might not break the continuum of other undamaged raft groups. So I'm inclined to handle corruption based on single region (rather than single file), and design interfaces around this assumption.

yiwu-arbug:
> @yiwu-arbug User should be responsible for any protobuf parsing, raft engine only store raw blob data. So data access tool should also be implemented by user (tikv-ctl) using existing engine methods.

[...]

PR #102, opened 2021-09-23
Add Physical lock file so that multiple processes can't operate on the same db directory simultaneously

yuqi1129:
Description: see #101 and #99

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -25,6 +25,7 @@ use crate::pipe_log::{FileId, LogQueue, PipeLog, SequentialReplayMachine};
 use crate::purge::{PurgeHook, PurgeManager};
 use crate::util::{HashMap, InstantExt};
 use crate::{GlobalStats, Result};
+use std::fs::File;
 
 const SLOTS_COUNT: usize = 128;
 
@@ -307,6 +308,8 @@ where
 
     listeners: Vec<Arc<dyn EventListener>>,
 
+    engine_file: File,
+
     _phantom: PhantomData<B>,
 }
 
@@ -348,7 +351,7 @@ where
         listeners.push(Arc::new(PurgeHook::new()) as Arc<dyn EventListener>);
 
         let start = Instant::now();
-        let (pipe_log, append, rewrite) =
+        let (pipe_log, append, rewrite, file) =
             FilePipeLog::open::<ParallelRecoverContext>(&cfg, file_builder, listeners.clone())?;
         let pipe_log = Arc::new(pipe_log);
         info!("Recovering raft logs takes {:?}", start.elapsed());
@@ -384,6 +387,7 @@ where
             pipe_log,
             purge_manager,
             listeners,
+            engine_file: file,
             _phantom: PhantomData,
         })
     }
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -25,7 +25,6 @@ use crate::pipe_log::{FileId, LogQueue, PipeLog, SequentialReplayMachine};
 use crate::purge::{PurgeHook, PurgeManager};
 use crate::util::{HashMap, InstantExt};
 use crate::{GlobalStats, Result};
-use std::fs::File;
 
 const SLOTS_COUNT: usize = 128;
 
@@ -308,8 +307,6 @@ where
 
     listeners: Vec<Arc<dyn EventListener>>,
 
-    engine_file: File,
-
     _phantom: PhantomData<B>,
 }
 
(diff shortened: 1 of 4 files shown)


src/engine.rs
```diff
@@ -307,6 +308,8 @@ where
 
     listeners: Vec<Arc<dyn EventListener>>,
 
+    engine_file: File,
```
tabokie, line 311: Put this inside `FilePipeLog`?

src/file_pipe_log.rs
```diff
@@ -25,6 +25,8 @@ use crate::pipe_log::{FileId, LogQueue, PipeLog, SequentialReplayMachine};
 use crate::reader::LogItemBatchFileReader;
 use crate::util::InstantExt;
 use crate::{Error, Result};
+use std::fs::File;
```
tabokie, line 28: It should be put above with other `std::xx`

src/file_pipe_log.rs
```diff
@@ -25,6 +25,8 @@ use crate::pipe_log::{FileId, LogQueue, PipeLog, SequentialReplayMachine};
...
+use fs2::FileExt;
```
tabokie, line 29: ditto, belong to `third_party::xx`

src/file_pipe_log.rs
```diff
@@ -794,6 +811,32 @@ mod tests {
...
 
+    #[test]
+    fn test_dir_lock() {
+        let dir = Builder::new().prefix("test_lock").tempdir().unwrap();
+        let path = dir.path().to_str().unwrap();
+
+        let mut cfg = Config::default();
+        cfg.dir = path.to_owned();
+        cfg.bytes_per_sync = ReadableSize(1024u64);
+        cfg.target_file_size = ReadableSize(32 * 1024 as u64);
```
tabokie, line 822: No need to specify them.

src/file_pipe_log.rs
```diff
@@ -794,6 +811,32 @@ mod tests {
...
+
+        let r1 = FilePipeLog::open::<BlackholeSequentialReplayMachine>(
+            &cfg,
+            Arc::new(DefaultFileBuilder {}),
+            vec![],
+        );
```
tabokie, line 828: In tests, we normally unwrap the result rather than asserting `is_ok()`.
```suggestion
        let r1 = FilePipeLog::open::<BlackholeSequentialReplayMachine>(
            &cfg,
            Arc::new(DefaultFileBuilder {}),
            vec![],
        ).unwrap();
```

src/file_pipe_log.rs
```diff
@@ -794,6 +811,32 @@ mod tests {
...
+
+        assert_eq!(r1.is_ok(), true);
+        // Only one thread can hold file lock
+        let r2 = FilePipeLog::open::<BlackholeSequentialReplayMachine>(
+            &cfg,
+            Arc::new(DefaultFileBuilder {}),
+            vec![],
+        );
+        assert_eq!(r2.is_err(), true);
```
tabokie, line 837: Could you verify the error message is `failed to create lock` or `maybe another instance is using this directory`? I checked that file lock is a re-entrant lock, which means the second `FilePipeLog` can lock the file just fine.

tabokie:
BTW, you can run `cargo clippy --all` before committing, it can detect lint and format issues.

src/engine.rs
yuqi1129: @tabokie I have try to put it in `FilePipeLog `, but `FilePipeLog ` has marked as `Clone` and File is not clonable and had no idea to solve this problem, so i put it there, do you have any idea about this?

src/file_pipe_log.rs
yuqi1129: Sure, Seems that run `cargo clippy` and `cargo build` can't detect this problem

src/file_pipe_log.rs
yuqi1129: OK, i will more the test more accurate

src/file_pipe_log.rs
yuqi1129: > BTW, you can run `cargo clippy --all` before committing, it can detect lint and format issues.

OK, thanks for your kind suggestion

src/file_pipe_log.rs
tabokie: Yes, clippy only sorts the imports, not group them.

src/engine.rs
tabokie: You can remove the `#[derive(Clone)]` for `FilePipeLog`, it's not needed now. As a side note, you can make any type `T` cloneable by wrapping it into `Arc<T>`.

yuqi1129: Got it

yuqi1129:
@tabokie , Can you take time to review a again

tabokie (approved):
Rest LGTM

src/file_pipe_log.rs
```diff
@@ -794,6 +813,29 @@ mod tests {
...
+        ).unwrap();
+
+        // Only one thread can hold file lock
+        let r2 = FilePipeLog::open::<BlackholeSequentialReplayMachine>(
+            &cfg,
+            Arc::new(DefaultFileBuilder {}),
+            vec![],
+        );
+        assert_eq!(r2.is_err(), true);
+        assert_eq!(format!("{}", r2.err().unwrap()).contains("maybe another instance is using this directory"), true);
```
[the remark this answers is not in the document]
yuqi1129: Got it

Issue #114, opened 2021-10-11
Add config to ignore all corrupted records when recover raft-log


yuqi1129:
Currently , raft-engine only support ignore corrupted raft-log at tail, When raft-log in the middle get corrupted,  raft-engine can't boot normally and issue #99 have not completed yet. So i propose to add an option to ignore all corrupted files

PR #115, opened 2021-10-11
Add config to ignore all corrupted records when recover raft-log

yuqi1129:
see #114

### DIFF
--- a/src/file_pipe_log.rs
+++ b/src/file_pipe_log.rs
@@ -596,6 +596,13 @@ impl<B: FileBuilder> FilePipeLog<B> {
                                 f.fd.truncate(reader.valid_offset())?;
                                 break;
                             }
+
+                            Err(e) if recovery_mode == RecoveryMode::TolerateCorruptedAllRecords =>
+                            {
+                                warn!("All corrupted raft logs will be truncated,  error of one file: {}",  e);
+                                f.fd.truncate(reader.valid_offset())?;
+                                break
+                            }
                             Err(e) => return Err(e),
                         }
                     }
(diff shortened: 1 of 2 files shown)


src/file_pipe_log.rs
```diff
@@ -596,6 +596,13 @@ impl<B: FileBuilder> FilePipeLog<B> {
                                 f.fd.truncate(reader.valid_offset())?;
                                 break;
                             }
+
+                            Err(e) if recovery_mode == RecoveryMode::TolerateCorruptedAllRecords =>
+                            {
+                                warn!("All corrupted raft logs will be truncated,  error of one file: {}",  e);
+                                f.fd.truncate(reader.valid_offset())?;
+                                break
+                            }
```
tabokie, line 605: It's not as simple as that. A corruption here would break the continuum of certain raft groups, and cause panics when memtables of those raft groups are restored. In specific, this assertion: https://github.com/tikv/raft-engine/blob/745e494bf9acafd88703bf3351cf93040d059ebb/src/memtable.rs#L116-L121

tabokie (commented):
I don't think we should add a config to silently drop middle corruption, as its recovery is non-trivial.

Also, I'd like to point out a few things about working with GitHub:
1. ~We open issue mainly to encourage discussion (e.g. a unresolved bug report, an early idea for potential optimization, a development task that requires help from community). Generally, if you already submit the code as PR, there is no need to create an accompanying issue. Or in other cases, you create an issue and expect feedbacks from project maintainer, you shouldn't submit any codes patches so fast.~
2. Commits messages in PR should be meaningful, so you shouldn't reuse the same message for every commit.
3. You should write "Fix #114" in the PR description, and when you do that, GitHub will identify the issue number and list it in the "Linked issue" sidebar. This way when PR is merged, the linked issue is automatically closed as well.

tabokie:
Closing this in favor of #140

PR #120, opened 2021-10-19
run asan in CI and stabilize write barrier test

tabokie:

Run ASAN in linux CI pipeline. The write barrier test is unstable when ran under asan/miri, fixing it. Also some housekeeping work: remove build step in CI, remove useless code.

TSAN still can't run because we use crossbeam channel for parallel recovery.


skyzh:
> Crossbeam channel will definitely cause race condition, as a feature. So this won’t work.

FYI https://github.com/rayon-rs/rayon/issues/812

PR #131, opened 2021-10-27
rollback to a consistent state after fsync error

MrCroxx:
Make `PipeLog` try to rollback to a consistent state after fsync error.

According to [PostgreSQL's fsync() surprise](https://lwn.net/Articles/752063/), `fsync` is not retriable. We must make pipe log rollback to where it last synced.

## What's changed?

**invariant: non-active log files are always synced and safe.**

Updated:

After discussion, the final decisions of this pr are listed below:

1. Panics when there is any `fsync` error.
2. Truncate active log file when there is `write` error during buffer writing log file.
3. Panics when there is any error during rotating log file (truncate/sync/create).

~Introduce `pre_write` and `post_write` for `PipeLog` to acquire or release a `WriteContext`, which can be used to track the sync state.~ 

~If `fsync` raises an error, log will rollback to last synced state, and set errors for the writers in the current write group. But if there are bytes rollbacked but in writers in the previous write groups, RaftEngine will panic, because that means some data that the user thinks is safe but lost actually.~

~`new_log_file` is devided into 2 phase. First truncate and sync the old log file, then rotate to a new log file. The second phase is safe to retry if `sync_dir` fails, because the new log will overwrite the failed one.~

TODO:
- [x] Unit tests with failpoints.


src/file_pipe_log.rs
```diff
@@ -668,9 +713,41 @@ impl<B: FileBuilder> PipeLog for FilePipeLog<B> {
...
+        if let Err(e) = match ctx.syncable {
+            Syncable::DontNeed => Ok(()),
+            Syncable::NeedSync => self.mut_queue(queue).sync(),
+            Syncable::NeedRotate => self.mut_queue(queue).truncate_and_sync(),
+        } {
+            self.mut_queue(queue).active_file.rollback()?;
+            return Err(Error::Fsync(ctx.synced, e.to_string()));
+        }
+        if ctx.syncable == Syncable::NeedRotate {
+            self.mut_queue(queue).rotate()?;
```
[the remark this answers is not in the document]
MrCroxx: The impl split rotation to 2 parts, first call `truncate_and_sync` for the current log, then call `rotate` to create and move the writer to the new log file. So calling `rotate` here means the old log file has already been truncated and synced.

src/errors.rs
```diff
@@ -15,8 +15,8 @@ pub enum Error {
     Corruption(String),
     #[error("IO Error: {0}")]
     Io(#[from] IoError),
-    #[error("Fsync Error: {0}")]
-    Fsync(String),
+    #[error("Fsync Error: reason: {0}")]
+    Severe(String),
```
tabokie, line 19: There is no need for making another error type, if:
1) We panic when fsync failed, no matter the error type is fatal or else
2) We don't propagate the error to multiple writers, so don't need a cloneable error type

tabokie (commented):
Needs unit tests. I've added some IO failpoints in #139, every one of them needs to be tested. You can use `catch_unwind_silent` to test the panics.

Cargo.toml
```diff
@@ -58,3 +58,9 @@ members = [ "stress" ]
 name = "bench_recovery"
 harness = false
 required-features = ["failpoints"]
+
+[[test]]
```
tabokie, line 62: You should send another PR to move all existing failpoint tests to this folder.

src/file_pipe_log.rs
```diff
@@ -160,9 +161,6 @@ impl<W: Seek + Write> ActiveFile<W> {
     }
 
     fn rotate(&mut self, fd: Arc<LogFd>, writer: W) -> Result<()> {
```
tabokie, line 163: maybe rename it to `reset`

failpoints/io_error_test.rs
```diff
@@ -0,0 +1,336 @@
...
+            dir: dir.path().to_str().unwrap().to_owned(),
+            bytes_per_sync: ReadableSize::kb(1024),
+            target_file_size: ReadableSize::kb(1024),
+            batch_compression_threshold: ReadableSize(0),
+            ..Default::default()
+        };
+
+        // b0 (ctx); b1 success; b2 fail, truncate; b3 success
+        let mut hook = FailpointsHook::new();
+        hook.register_pre_append_action(2, || {
```
tabokie, line 257: I don't particularly like this approach. Could you try using `cfg_callback`? might be easier.

Cargo.toml
MrCroxx: OK. I'd like to move io error test to this folder first in this PR, then send another one to move the existing test.

failpoints/io_error_test.rs
```diff
@@ -0,0 +1,341 @@
...
+            }
+        }
+    }
+
+    #[test]
+    fn test_open_error() {
+        fail::cfg("log_fd::open::err", "return").unwrap();
+        assert!(panic::catch_unwind(|| {
+            let dir = tempfile::Builder::new()
+                .prefix("handle_io_error")
```
tabokie, line 154:
```suggestion
                .prefix("test_open_error")
```

src/file_pipe_log.rs
```diff
@@ -712,10 +702,30 @@ impl<B: FileBuilder> PipeLog for FilePipeLog<B> {
...
-        if let Some(fd) = self.get_queue(queue).get_active_fd() {
-            fd.sync()?;
+    fn maybe_sync(&self, queue: LogQueue, force: bool) -> Result<()> {
+        let mut manager = match queue {
+            LogQueue::Append => self.appender.write(),
+            LogQueue::Rewrite => self.rewriter.write(),
+        };
+
+        if manager.active_file.written >= manager.rotate_size {
+            // need rotate
```
tabokie, line 712: These comments can be removed.

tabokie (commented):
Tests are failing.

MrCroxx:
> Tests are failing.

Yep. It works fine with my env and I’m looking for the reason.

failpoints/io_error_test.rs
```diff
@@ -0,0 +1,341 @@
...
+        fail::cfg("log_fd::write::err", "off").unwrap();
+    }
+
+    #[test]
+    fn test_concurrent_write_error() {
+        // b1 success; b2 fail, truncate; b3 success
+        let timer = AtomicU64::new(0);
+        fail::cfg_callback("engine::write::pre", move || {
+            match timer.fetch_add(1, std::sync::atomic::Ordering::SeqCst) {
+                2 => fail::cfg("log_fd::write::post_err", "return").unwrap(),
```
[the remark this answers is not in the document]
MrCroxx: This is for checking if the following appends correctly overwrite the pervious ones.

tests/failpoints/test_io_error.rs
```diff
@@ -0,0 +1,361 @@
...
+        for t in self.ths.drain(..) {
+            t.join().unwrap();
+        }
+    }
+}
+
+#[test]
+fn test_open_error() {
+    fail::cfg("log_fd::open::err", "return").unwrap();
+    assert!(catch_unwind_silent(|| {
```
tabokie, line 174: This is not a good way to test panic, I can't tell which part is actually panicking from this code.

tests/failpoints/test_io_error.rs
```diff
@@ -0,0 +1,361 @@
...
+    fail::cfg("log_fd::open::err", "off").unwrap();
+}
+
+#[test]
+fn test_rotate_error() {
+    // panic when truncate
+    fail::cfg("active_file::truncate::force", "return").unwrap();
+    fail::cfg("log_fd::truncate::err", "return").unwrap();
+    assert!(catch_unwind_silent(|| {
+        write_tmp_engine(ReadableSize::kb(1024), ReadableSize::kb(4), 1024, 1, 4)
```
tabokie, line 200: `write_tmp_engine` seems an overkill.

src/log_file.rs
```diff
@@ -122,6 +122,9 @@ impl LogFd {
             written += bytes;
             offset += bytes;
         }
+        fail_point!("log_fd::write::post_err", |_| {
```
tabokie, line 125: maybe remove `write::err`.

tests/failpoints/test_io_error.rs
```diff
@@ -0,0 +1,361 @@
...
+    })
+    .is_err());
+    fail::cfg("log_fd::write::err", "off").unwrap();
+}
+
+#[test]
+fn test_concurrent_write_error() {
+    // b1 success; b2 fail, truncate; b3 success
```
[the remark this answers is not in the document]
MrCroxx: 'b' for 'batch'. But `writer` here is better. I'll replace them.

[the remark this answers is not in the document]
MrCroxx: `writer` => `write`

src/file_pipe_log.rs
```diff
@@ -712,10 +702,28 @@ impl<B: FileBuilder> PipeLog for FilePipeLog<B> {
...
-            fd.sync()?;
+    fn maybe_sync(&self, queue: LogQueue, force: bool) -> Result<()> {
+        let mut manager = match queue {
+            LogQueue::Append => self.appender.write(),
+            LogQueue::Rewrite => self.rewriter.write(),
+        };
+
+        if manager.active_file.written >= manager.rotate_size {
+            if let Err(e) = manager.rotate() {
+                panic!(
```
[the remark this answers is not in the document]
MrCroxx: I ran the coverage test locally and it said these lines are covered. And I checked manually and made sure that it is covered. 

image


I've also checked the CI script, which seems fine. I'm not sure if it's a bug of the CI coverage test.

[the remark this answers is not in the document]
MrCroxx: Just  found where the problem lays, fixing.

MrCroxx:
stress --regions 1000 --write-sync true --time 600

[current]
Throughput(QPS) = 10119.08
Latency(μs) min = 40, avg = 83.03, p50 = 61, p90 = 161, p95 = 187, p99 = 230, p99.9 = 288, max = 24607
Fairness = 100.0%
Write Bandwidth = 7.5MiB/s


[master]
Throughput(QPS) = 10249.05
Latency(μs) min = 42, avg = 80.07, p50 = 59, p90 = 155, p95 = 174, p99 = 214, p99.9 = 265, max = 16639
Fairness = 100.0%
Write Bandwidth = 7.7MiB/s

PR #163, opened 2021-12-01
Replace app builder with derive macros in stress tool

### LINKED ISSUE #155: Replace app builder with derive macros in stress tool
Similar to [ctl](https://github.com/tikv/raft-engine/tree/master/ctl) tool, we can replace the use of clap with [structopt](https://docs.rs/structopt/0.3.25/structopt/) in [stress](https://github.com/tikv/raft-engine/tree/master/stress). It can greatly simplify the code.

ztelur:
replace structopt with clap-v3.0.0-beta.5 in ctl and stress tool referred to #155

Close #155.


stress/src/main.rs
```diff
@@ -47,6 +49,214 @@ const DEFAULT_WRITE_REGION_COUNT: u64 = 5;
...
+    path: String,
+
+    #[clap(
+        long = "time",
+        takes_value = true,
+        default_value = formatcp!("{}", DEFAULT_TIME.as_secs()),
+        value_name = "time[s]",
+        about = "Set the stress test time"
+    )]
+    time: String,
```
tabokie, line 76: maybe u64? so we don't need to manually parse the string.

stress/src/main.rs
```diff
@@ -378,260 +588,51 @@ impl EventListener for WrittenBytesHook {
...
+    args.purge_interval = Duration::from_millis(opts.purge_interval.parse::<u64>().unwrap());
+
+    if let Some(ttl) = opts.compact_ttl {
+        args.compact_ttl = Duration::from_millis(ttl.parse::<u64>().unwrap());
     }
-    if let Some(s) = matches.value_of("write_region_count") {
-        args.write_region_count = s.parse::<u64>().unwrap();
+
+    if args.compact_ttl.as_millis() > 0 {
+        println!("Not supported");
```
tabokie, line 614: This can be put into validator as well.

ctl/Cargo.toml
```diff
@@ -6,4 +6,5 @@ authors = ["The TiKV Authors"]
 
 [dependencies]
 structopt = "0.3"
```
tabokie, line 8: structopt can be removed.

tabokie (commented):
Could you remove `Cargo.lock` from you changeset? After that you can also add "Cargo.lock" to gitignore.

stress/src/main.rs
ztelur: I am a little confused about this validation. it seems that `compact_ttl` should always be zero . Should we just remove it from stress's paramters?

stress/src/main.rs
ztelur: To avoid parsing manually, I change type of most parameters, except `write_sync` and `reuse_data` because it will trigger 'default_value is meaningless for bool' error

ztelur:
> Could you remove `Cargo.lock` from you changeset? After that you can also add "Cargo.lock" to gitignore.

Done

ztelur:
BTW, running stress with default parameter value will lead to error "Write region count must be smaller than region-count / write-threads". Should we change the default value of those three parameters
`
target/debug/stress --path tmp
`

stress/src/main.rs
tabokie: Yes, it's a little hard to implement. You can remove it for now.

tabokie:
> BTW, running stress with default parameter value will lead to error "Write region count must be smaller than region-count / write-threads". Should we change the default value of those three parameters `target/debug/stress --path tmp`

@ztelur Please do.

stress/src/main.rs
```diff
@@ -47,6 +49,214 @@ const DEFAULT_WRITE_REGION_COUNT: u64 = 5;
...
+    write_region_count: u64,
+
+    #[clap(
+        long = "write-sync",
+        value_name = "sync",
+        takes_value = true,
+        default_value = formatcp!("{}", DEFAULT_WRITE_SYNC),
+        about = "Whether to sync write raft logs"
+    )]
+    write_sync: String,
```
tabokie, line 203: They can all be typed. Let's remove any `parse` from the code.

tabokie (approved):
Rest looks good.

stress/src/main.rs
ztelur: changing `write_sync` and `reuse_data` to bool type will leading to compile error 'default_value is meaningless for bool'. it look like a clap problem same with structopt mentioned at https://github.com/TeXitoi/structopt/issues/93

tabokie: I see. That's because is a "is_present" flag. You can change it from `write_sync` to `write_without_sync` instead and remove default value setting.

PR #164, opened 2021-12-05
Bump nix, rand, and more

PsiACE:
Signed-off-by: Chojan Shang <psiace@outlook.com>

### DIFF
--- a/src/file_pipe_log/log_file.rs
+++ b/src/file_pipe_log/log_file.rs
@@ -30,19 +30,14 @@ const FILE_ALLOCATE_SIZE: usize = 2 * 1024 * 1024;
 pub struct LogFd(RawFd);
 
 fn from_nix_error(e: nix::Error, custom: &'static str) -> std::io::Error {
-    match e {
-        nix::Error::Sys(no) => {
-            let kind = std::io::Error::from(no).kind();
-            std::io::Error::new(kind, custom)
-        }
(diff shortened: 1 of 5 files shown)


src/file_pipe_log/log_file.rs
```diff
@@ -30,19 +30,14 @@ const FILE_ALLOCATE_SIZE: usize = 2 * 1024 * 1024;
...
 fn from_nix_error(e: nix::Error, custom: &'static str) -> std::io::Error {
-    match e {
-        nix::Error::Sys(no) => {
-            let kind = std::io::Error::from(no).kind();
-            std::io::Error::new(kind, custom)
-        }
-        e => std::io::Error::new(std::io::ErrorKind::Other, format!("{}: {:?}", custom, e)),
-    }
+    let kind = std::io::Error::from(e).kind();
+    std::io::Error::new(kind, custom)
```
tabokie, line 34: Handle the case where kind == Other?

tabokie (approved):
Thanks! Rest looks good.

src/file_pipe_log/log_file.rs
PsiACE: In nix 0.23, all error types have been unified to Errno, and it appears that the `Other` kind no longer needs to be handled

tabokie: I see.

PR #167, opened 2021-12-06
Support dump subcommand in ctl

### LINKED ISSUE #151: Support `dump` subcommand in ctl
As #148 has been merged, this issue aim to support `dump` subcommand, that is 

Dump: dump out all operations in log files
```
ctl dump --file /path/to/file [--raft id1,id2,id3]
ctl dump --path /path path [--raft id1,id2,id3]
```

For more detail, please see #144

yuqi1129:
Add subcommand 'dump' in ctl utils.

Close #151


src/log_dumper.rs
```diff
@@ -0,0 +1,35 @@
+use crate::file_pipe_log::ReplayMachine;
+use crate::log_batch::{LogItem, LogItemBatch};
+use crate::{FileId, LogQueue};
+
+use std::collections::HashMap;
+
+pub struct LogDumer {
+    // raft_groups_id to LogItem vec
+    pub(crate) logs: HashMap<u64, Vec<LogItem>>,
```
tabokie, line 9: Use `LogItemReader` instead:
https://github.com/tikv/raft-engine/blob/e32e3e78604b7115621f543dea1231132abd8778/src/file_pipe_log/mod.rs#L48
You should wrap it into a `std::iter::Iter` type.

yuqi1129: 🆗

src/engine.rs
```diff
@@ -350,13 +352,47 @@ where
...
+            )));
+        }
+
+        let mut reader = if path.is_dir() {
+            LogItemReader::new_directory_reader(file_builder, path).unwrap()
+        } else {
+            LogItemReader::new_file_reader(file_builder, path).unwrap()
+        };
+
+        let mut log_map = HashMap::<u64, Vec<LogItem>>::default();
```
tabokie, line 381: No, I mean return an iterator to user, instead of reading all the items into vector.

yuqi1129: Return an iterator to user instead of a vector is for the possibility that data may be too large and we can afford to allocate so many memory? @tabokie

tabokie: Yes, and user might not need all of the items.

yuqi1129: Got it

src/engine.rs
```diff
@@ -350,13 +351,19 @@ where
...
 
     /// Dumps all operations.
-    #[allow(unused_variables)]
-    pub fn dump_with_file_builder(
-        path: &Path,
-        raft_groups: &[u64],
-        file_builder: Arc<B>,
-    ) -> Result<Vec<LogItem>> {
-        todo!()
+    pub fn dump_with_file_builder(path: &Path, file_builder: Arc<B>) -> Result<LogItemReader<B>> {
```
tabokie, line 354: Better implement https://doc.rust-lang.org/std/iter/trait.Iterator.html for `LogItemReader`

tabokie: And add an unit test for dump file, dump directory, and dump invalid path.

yuqi1129: done

src/engine.rs
```diff
@@ -298,7 +299,10 @@ impl Engine<DefaultFileBuilder, FilePipeLog<DefaultFileBuilder>> {
         )
     }
 
-    pub fn dump(path: &Path, raft_groups: &[u64]) -> Result<Vec<LogItem>> {
+    pub fn dump<'a>(
+        path: &Path,
+        raft_groups: &'a [u64],
```
tabokie, line 304: Actually, we can remove this parameter, and do the filtering in "ctl". So that you don't need to store it in iterator anymore.

src/file_pipe_log/mod.rs
```diff
@@ -221,7 +222,51 @@ pub mod debug {
                     }
                 }
             }
-            assert!(reader.next().is_none())
+
+            //dump dir with raft groups. 8 element with raft groups 7 and 2 elements with raft groups 8
+            let raft_groups_ids = &[7u64; 1];
+            let mut dump_it = Engine::dump(dir.path(), raft_groups_ids).unwrap();
+            let mut total = 0;
+            while let Some(_) = dump_it.next() {
```
tabokie, line 230: You can iterate an `Iterator` by
```rust
for item in dump_it {
  // do things
}
```

src/engine.rs
yuqi1129: @tabokie If we remove this parameter, method `dump` seems do nothing but convert `LogItemReader.next()` to a normal  iterator, is that OK?

tabokie: Yes, ctl still has this parameter.

yuqi1129: 👌

src/file_pipe_log/mod.rs
```diff
@@ -221,7 +222,51 @@ pub mod debug {
                     }
                 }
             }
-            assert!(reader.next().is_none())
+
```
tabokie, line 225: Put the test in engine.rs. You are not supposed to use `Engine` in a component crate.

src/engine.rs
```diff
@@ -372,28 +365,21 @@ where
...
-            raft_grous_ids: raft_groups,
-            item_reader,
-        })
+        Ok(DumpIterator { item_reader })
     }
 }
 
-pub struct DumpIterator<'a, B: FileBuilder> {
-    raft_grous_ids: &'a [u64],
+pub struct DumpIterator<B: FileBuilder> {
```
tabokie, line 372: You can directly implement it for `LogItemReader` now (in file_pipe_log::debug).

tabokie (commented):
Rest looks good.

src/engine.rs
yuqi1129: OK

src/file_pipe_log/mod.rs
```diff
@@ -52,6 +53,17 @@ pub mod debug {
         items: VecDeque<LogItem>,
     }
 
+    impl<B: FileBuilder> Iterator for LogItemReader<B> {
+        type Item = LogItem;
```
tabokie, line 57: Should be `Result<LogItem>`, or the dump will stop once encountering the first corruption.

src/engine.rs
```diff
@@ -1126,4 +1151,112 @@ mod tests {
...
+        batch2.put(8, b"key3".to_vec(), b"value".to_vec());
+        batch2
+            .add_entries::<Entry>(8, &generate_entries(5, 15, Some(&entry_data)))
+            .unwrap();
+        batches.push(vec![batch, batch2]);
+
+        for bs in batches.iter_mut() {
+            let file_path = file_id.build_file_path(dir.path());
+            // Write a file.
+            let mut writer =
```
tabokie, line 1187: You shouldn't use writer here. Refer to other tests in this file, use `Engine::write` instead.

src/engine.rs
```diff
@@ -1126,4 +1151,112 @@ mod tests {
...
+        //dump dir with raft groups. 8 element with raft groups 7 and 2 elements with raft groups 8
+        let raft_groups_ids = &[7u64; 1];
+        let dump_it = Engine::dump(dir.path()).unwrap();
+        let mut total = 0;
+        for item in dump_it {
+            if raft_groups_ids.is_empty() || raft_groups_ids.contains(&item.raft_group_id) {
+                total += 1;
+            }
+        }
+        assert!(total == 8);
```
tabokie, line 1214: No need to involve `raft_groups_ids` here as well. Verify the total count is enough.

src/file_pipe_log/mod.rs
yuqi1129: @tabokie  We should dump all the log and omit error in the process? and should we dump the error info?

src/engine.rs
yuqi1129: 🆗

src/file_pipe_log/mod.rs
tabokie: We should print the error and continue.

src/engine.rs
tabokie: @yuqi1129 Please address this comment.

src/engine.rs
```diff
@@ -350,13 +351,34 @@ where
...
+}
+
+impl<B: FileBuilder> Iterator for DumpIterator<B> {
+    type Item = LogItem;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        if let Some(Ok(item)) = self.item_reader.next() {
+            return Some(item);
+        }
+        None
```
tabokie, line 381: Those are not needed anymore.

src/engine.rs
yuqi1129: @tabokie , I have move the test to file `engine.rs` and you mean i should put it in `mod.rs`

> use Engine::write instead 

OK, I miss it.

tabokie: > you mean i should put it in `mod.rs`

No, the placement is fine.

src/engine.rs
yuqi1129: @tabokie  We do not need this iterator any more ?

tabokie: You already implemented it for `LogItemReader`.

yuqi1129:
@tabokie All discussions have been resolved, please take time to look again.

src/file_pipe_log/mod.rs
```diff
@@ -6,6 +6,8 @@ mod pipe;
 mod pipe_builder;
 mod reader;
 
+pub use format::FileNameExt;
+pub use log_file::{build_file_writer, LogFd};
```
tabokie, line 10: No need to export them now?

src/engine.rs
```diff
@@ -1126,4 +1135,76 @@ mod tests {
...
+        let cfg = Config {
+            dir: dir.path().to_str().unwrap().to_owned(),
+            ..Default::default()
+        };
+
+        let engine =
+            RaftLogEngine::open_with_file_builder(cfg, Arc::new(DefaultFileBuilder)).unwrap();
+        for bs in batches.iter_mut() {
+            for batch in bs.iter_mut() {
+                let _ = engine.write(batch, false);
```
tabokie, line 1172: You should do this. In test you should unwrap the result to make sure no error occurred.

src/engine.rs
```diff
@@ -1126,4 +1135,76 @@ mod tests {
...
+            }
+
+            let _ = engine.sync();
```
tabokie, line 1175: ditto

src/engine.rs
```diff
@@ -1126,4 +1135,76 @@ mod tests {
...
+
+        //dump dir with raft groups. 8 element with raft groups 7 and 2 elements with raft groups 8
+        let dump_it = Engine::dump(dir.path()).unwrap();
+        let mut total = 0;
+        for item in dump_it {
+            if item.is_ok() {
+                total += 1;
+            }
+        }
+        assert!(total == 10);
```
tabokie, line 1186: We can use `dump_it.count()`

tabokie (approved):
Rest looks good.

src/file_pipe_log/mod.rs
yuqi1129: No,  Because `FileNameExt` and `LogFd`  will be used in test code of `engine.rs`

yuqi1129: as refer to `build_file_writer`, we can safely remove it

src/engine.rs
yuqi1129: Should we need to check the item is Ok? as iterator will return Ok(v) or Err(msg)?

src/engine.rs
yuqi1129: 👌

src/file_pipe_log/mod.rs
yuqi1129:
```
        let dump_it = Engine::dump(file_id.build_file_path(dir.path()).as_path()).unwrap();
```
This will use `LogFd` and `FileNameExt `

src/engine.rs
tabokie: Oh, that makes sense. You can call `dump_it.map(|i| i.unwrap()).count()`.

src/file_pipe_log/mod.rs
tabokie: You can keep `FileNameExt`, `LogFd` can be removed?

yuqi1129: I think not
```rust


        //dump file
        let file_id = FileId {
            queue: LogQueue::Rewrite,
            seq: 1,
        };
        let dump_it = Engine::dump(file_id.build_file_path(dir.path()).as_path()).unwrap();
        assert!(0 == dump_it.map(|i| i.unwrap()).count());

        //dump dir that does not exists
        assert!(Engine::dump(Path::new("/not_exists_dir")).is_err());
```

`FileId` is used

yuqi1129: Yes, you are so careful 👍 
By the way, why `cargo clippy` and `cargo build` can't detect unused import ?

tabokie: Because it's a public export, even though we don't use it internally, rust thinks other users might use it from outside this library.

yuqi1129: 😃

src/engine.rs
```diff
@@ -1126,4 +1135,74 @@ mod tests {
...
+
+        let engine =
+            RaftLogEngine::open_with_file_builder(cfg, Arc::new(DefaultFileBuilder)).unwrap();
+        for bs in batches.iter_mut() {
+            for batch in bs.iter_mut() {
+                engine.write(batch, false).unwrap();
+            }
+
+            engine.sync().unwrap();
+        }
```
tabokie, line 1176: Close the engine before you dump it. Or there will be some uninitialized space at the end of file.

src/engine.rs
```diff
@@ -1126,4 +1135,74 @@ mod tests {
...
+        //dump file
+        let file_id = FileId {
+            queue: LogQueue::Rewrite,
+            seq: 1,
+        };
+        let dump_it = Engine::dump(file_id.build_file_path(dir.path()).as_path()).unwrap();
+        let total = dump_it
+            .map(|i| {
+                i.as_ref().unwrap();
+            })
```
tabokie, line 1196:
```suggestion
            .map(|i| i.unwrap())
```

yuqi1129: if i directly follow your advice，`cargo clippy` will output the following result:

image

tabokie: Oh, change it to `inspect` is enough.

yuqi1129: Yes,  I have changed it one place and forget to change all
```
        let total = dump_it
            .inspect(|i| {
                i.as_ref().unwrap();
            })
            .count();
        assert!(total == 10);
```

Consider the following message, `as_ref()` can't be remove

yuqi1129:
``` 
move occurs because `*i` has type `std::result::Result<log_batch::LogItem, errors::Error>`, which does not implement the `Copy` trait
     |                 help: consider borrowing the `Result`'s content: `i.as_ref()`
```

tabokie: Yes, only map allows that.

src/engine.rs
yuqi1129: seems simply use ` drop(engine)` can't fix test error, the test failed again

tabokie: It seems to be a bug. Could you implement `Drop` for the `SinglePipe`? In which we properly close the file by calling `writer.close()`.

yuqi1129: @tabokie ， OK, let me have a try

PR #175, opened 2021-12-13
support `truncate` subcommand

yuqi1129:
Signed-off-by: yuqi1129 <yuqi4733@gmail.com>

### DIFF
--- a/ctl/src/main.rs
+++ b/ctl/src/main.rs
@@ -1,9 +1,10 @@
 // Copyright (c) 2017-present, PingCAP, Inc. Licensed under Apache-2.0.
 
 use std::path::Path;
+use std::str::FromStr;
 
 use clap::{crate_authors, crate_version, AppSettings, Parser};
-use raft_engine::{Engine, Error, LogQueue, Result as EngineResult};
+use raft_engine::{Engine, Error, LogQueue, Result as EngineResult, TruncateMode};
 
 #[derive(Debug, clap::Parser)]
 #[clap(
@@ -112,7 +113,12 @@ impl ControlOpt {
         queue: &str,
         raft_groups: &[u64],
     ) -> EngineResult<()> {
-        Engine::unsafe_truncate(Path::new(path), mode, convert_queue(queue), raft_groups)
+        Engine::unsafe_truncate(
+            Path::new(path),
+            TruncateMode::from_str(mode)?,
+            convert_queue(queue),
+            raft_groups,
+        )
     }
 
     fn check(&self, path: &str) -> EngineResult<()> {
--- a/ctl/src/main.rs
+++ b/ctl/src/main.rs
@@ -1,10 +1,9 @@
 // Copyright (c) 2017-present, PingCAP, Inc. Licensed under Apache-2.0.
 
 use std::path::Path;
-use std::str::FromStr;
 
 use clap::{crate_authors, crate_version, AppSettings, Parser};
-use raft_engine::{Engine, Error, LogQueue, Result as EngineResult, TruncateMode};
+use raft_engine::{Engine, Error, LogQueue, Result as EngineResult};
 
 #[derive(Debug, clap::Parser)]
 #[clap(
@@ -115,7 +114,7 @@ impl ControlOpt {
     ) -> EngineResult<()> {
         Engine::unsafe_truncate(
             Path::new(path),
-            TruncateMode::from_str(mode)?,
+            mode.parse()?,
             convert_queue(queue),
             raft_groups,
         )
--- a/ctl/src/main.rs
+++ b/ctl/src/main.rs
@@ -89,14 +89,17 @@ impl ControlOpt {
         }
     }
 
-    fn dump(&self, path: &str, raft_groups: &[u64]) -> EngineResult<()> {
-        let r = Engine::dump(Path::new(path), &raft_groups.to_vec())?;
-
-        if r.is_empty() {
-            println!("No data");
-        } else {
-            println!("Raft entrys are as follows:\n");
-            r.iter().for_each(|entry| println!("{:?}", entry));
+    fn dump<'a>(&self, path: &str, raft_groups: &'a [u64]) -> EngineResult<()> {
+        let it = Engine::dump(Path::new(path))?;
+        for item in it {
+            if let Ok(v) = item {
+                if raft_groups.is_empty() || raft_groups.contains(&v.raft_group_id) {
+                    println!("{:?}", v)
+                }
+            } else {
+                // output error message
+                println!("{:?}", item)
+            }
         }
 
         Ok(())
--- a/src/consistency.rs
+++ b/src/consistency.rs
@@ -7,6 +7,7 @@ use std::sync::Arc;
 use crate::file_pipe_log::ReplayMachine;
 use crate::log_batch::{LogItemBatch, LogItemContent};
 use crate::pipe_log::{FileId, LogQueue};
+use crate::truncate::TruncateQueueParamter;
 use crate::{FileBuilder, Result};
 
 /// A `ConsistencyChecker` scans for log entry holes in a log queue. It will
@@ -70,7 +71,12 @@ impl ReplayMachine for ConsistencyChecker {
         Ok(())
     }
 
-    fn end<B: FileBuilder>(&mut self, _path: &Path, _builder: Arc<B>) -> Result<()> {
+    fn end<B: FileBuilder>(
+        &mut self,
+        _path: &Path,
+        _builder: Arc<B>,
+        _truncate_params: &TruncateQueueParamter,
+    ) -> Result<()> {
         Ok(())
     }
 }
--- a/src/consistency.rs
+++ b/src/consistency.rs
@@ -1,11 +1,13 @@
 // Copyright (c) 2017-present, PingCAP, Inc. Licensed under Apache-2.0.
 
 use hashbrown::HashMap;
+use std::path::Path;
+use std::sync::Arc;
 
 use crate::file_pipe_log::ReplayMachine;
 use crate::log_batch::{LogItemBatch, LogItemContent};
 use crate::pipe_log::{FileId, LogQueue};
-use crate::Result;
+use crate::{FileBuilder, Result};
 
 /// A `ConsistencyChecker` scans for log entry holes in a log queue. It will
 /// return a list of corrupted raft groups along with their last valid log index.
@@ -65,6 +67,10 @@ impl ReplayMachine for ConsistencyChecker {
         for (id, last_index) in rhs.corrupted.drain() {
             self.corrupted.entry(id).or_insert(last_index);
         }
+        Ok(())
+    }
+
+    fn end<B: FileBuilder>(&mut self, _path: &Path, _builder: Arc<B>) -> Result<()> {
         Ok(())
     }
 }
--- a/src/consistency.rs
+++ b/src/consistency.rs
@@ -1,14 +1,11 @@
 // Copyright (c) 2017-present, PingCAP, Inc. Licensed under Apache-2.0.
 
 use hashbrown::HashMap;
-use std::path::Path;
-use std::sync::Arc;
 
 use crate::file_pipe_log::ReplayMachine;
 use crate::log_batch::{LogItemBatch, LogItemContent};
 use crate::pipe_log::{FileId, LogQueue};
-use crate::truncate::TruncateQueueParameter;
-use crate::{FileBuilder, Result};
+use crate::Result;
 
 /// A `ConsistencyChecker` scans for log entry holes in a log queue. It will
 /// return a list of corrupted raft groups along with their last valid log index.
@@ -68,15 +65,6 @@ impl ReplayMachine for ConsistencyChecker {
         for (id, last_index) in rhs.corrupted.drain() {
             self.corrupted.entry(id).or_insert(last_index);
         }
-        Ok(())
-    }
-
-    fn end<B: FileBuilder>(
-        &mut self,
-        _path: &Path,
-        _builder: Arc<B>,
-        _truncate_params: &TruncateQueueParameter,
-    ) -> Result<()> {
         Ok(())
     }
 }
--- a/src/consistency.rs
+++ b/src/consistency.rs
@@ -7,7 +7,7 @@ use std::sync::Arc;
 use crate::file_pipe_log::ReplayMachine;
 use crate::log_batch::{LogItemBatch, LogItemContent};
 use crate::pipe_log::{FileId, LogQueue};
-use crate::truncate::TruncateQueueParamter;
+use crate::truncate::TruncateQueueParameter;
 use crate::{FileBuilder, Result};
 
 /// A `ConsistencyChecker` scans for log entry holes in a log queue. It will
@@ -75,7 +75,7 @@ impl ReplayMachine for ConsistencyChecker {
         &mut self,
         _path: &Path,
         _builder: Arc<B>,
-        _truncate_params: &TruncateQueueParamter,
+        _truncate_params: &TruncateQueueParameter,
     ) -> Result<()> {
         Ok(())
     }
--- a/src/consistency.rs
+++ b/src/consistency.rs
@@ -1,6 +1,7 @@
 // Copyright (c) 2017-present, PingCAP, Inc. Licensed under Apache-2.0.
 
 use hashbrown::HashMap;
+use std::path::Path;
 
 use crate::file_pipe_log::ReplayMachine;
 use crate::log_batch::{LogItemBatch, LogItemContent};
@@ -24,7 +25,7 @@ impl ConsistencyChecker {
 }
 
 impl ReplayMachine for ConsistencyChecker {
-    fn replay(&mut self, item_batch: LogItemBatch, _file_id: FileId) -> Result<()> {
+    fn replay(&mut self, item_batch: LogItemBatch, _file_id: FileId, _path: &Path) -> Result<()> {
         for item in item_batch.iter() {
             if let LogItemContent::EntryIndexes(ents) = &item.content {
                 if !ents.0.is_empty() {
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -2,7 +2,6 @@
 
 use std::marker::PhantomData;
 use std::path::Path;
-use std::str::FromStr;
 use std::sync::Arc;
 use std::time::Instant;
 
@@ -291,7 +290,7 @@ impl Engine<DefaultFileBuilder, FilePipeLog<DefaultFileBuilder>> {
 
     pub fn unsafe_truncate(
         path: &Path,
-        mode: &str,
+        mode: TruncateMode,
         queue: Option<LogQueue>,
         raft_groups: &[u64],
     ) -> Result<()> {
@@ -347,7 +346,7 @@ where
     #[allow(unused_variables)]
     pub fn unsafe_truncate_with_file_builder(
         path: &Path,
-        mode: &str,
+        mode: TruncateMode,
         queue: Option<LogQueue>,
         raft_groups: &[u64],
         file_builder: Arc<B>,
@@ -369,7 +368,7 @@ where
 
         builder.recover_or_truncate::<TruncateMachine>(Some(TruncateQueueParameter {
             queue,
-            truncate_mode: TruncateMode::from_str(mode)?,
+            truncate_mode: mode,
             raft_groups_ids: raft_groups,
         }))?;
 
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -13,13 +13,15 @@ use crate::consistency::ConsistencyChecker;
 use crate::event_listener::EventListener;
 use crate::file_builder::*;
 use crate::file_pipe_log::debug::LogItemReader;
-use crate::file_pipe_log::{get_lock_file, FilePipeLog, FilePipeLogBuilder, ReplayMachineBuilder};
+use crate::file_pipe_log::{
+    get_lock_file, DefaultMachineFactory, FilePipeLog, FilePipeLogBuilder, TruncateMachineFactory,
+};
 use crate::log_batch::{Command, LogBatch, MessageExt};
 use crate::memtable::{EntryIndex, MemTableAccessor, MemTableRecoverContext};
 use crate::metrics::*;
 use crate::pipe_log::{FileBlockHandle, FileId, LogQueue, PipeLog};
 use crate::purge::{PurgeHook, PurgeManager};
-use crate::truncate::{TruncateMachine, TruncateMode, TruncateQueueParameter};
+use crate::truncate::{TruncateMode, TruncateQueueParameter};
 use crate::write_barrier::{WriteBarrier, Writer};
 use crate::{Error, GlobalStats, Result};
 
@@ -79,7 +81,7 @@ where
         let mut builder = FilePipeLogBuilder::new(cfg.clone(), file_builder, listeners.clone());
         builder.scan()?;
         let (append, rewrite) =
-            builder.recover(&ReplayMachineBuilder::<MemTableRecoverContext>::default())?;
+            builder.recover(&DefaultMachineFactory::<MemTableRecoverContext>::default())?;
         let pipe_log = Arc::new(builder.finish()?);
         rewrite.merge_append_context(append);
         let (memtables, stats) = rewrite.finish();
@@ -335,7 +337,7 @@ where
         let mut builder = FilePipeLogBuilder::new(cfg, file_builder, Vec::new());
         builder.scan()?;
         let (append, rewrite) =
-            builder.recover(&ReplayMachineBuilder::<ConsistencyChecker>::default())?;
+            builder.recover(&DefaultMachineFactory::<ConsistencyChecker>::default())?;
         let mut map = rewrite.finish();
         for (id, index) in append.finish() {
             map.entry(id).or_insert(index);
@@ -376,9 +378,7 @@ where
             lock_file: Arc::new(get_lock_file(path.to_str().unwrap())?),
         };
 
-        let mut machine_builder = ReplayMachineBuilder::<TruncateMachine>::default();
-        machine_builder.truncate_params = Some(parms);
-        builder.recover(&machine_builder)?;
+        builder.recover(&TruncateMachineFactory::new(Some(parms)))?;
 
         Ok(())
     }
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -13,7 +13,7 @@ use crate::consistency::ConsistencyChecker;
 use crate::event_listener::EventListener;
 use crate::file_builder::*;
 use crate::file_pipe_log::debug::LogItemReader;
-use crate::file_pipe_log::{FilePipeLog, FilePipeLogBuilder};
+use crate::file_pipe_log::{FilePipeLog, FilePipeLogBuilder, ReplayMachineBuilder};
 use crate::log_batch::{Command, LogBatch, MessageExt};
 use crate::memtable::{EntryIndex, MemTableAccessor, MemTableRecoverContext};
 use crate::metrics::*;
@@ -78,7 +78,10 @@ where
         let start = Instant::now();
         let mut builder = FilePipeLogBuilder::new(cfg.clone(), file_builder, listeners.clone());
         builder.scan()?;
-        let (_append, _rewrite) = builder.recover_or_truncate::<MemTableRecoverContext>(None)?;
+        let machine_builder = ReplayMachineBuilder {
+            truncate_params: None,
+        };
+        let (_append, _rewrite) = builder.recover::<MemTableRecoverContext>(&machine_builder)?;
         let pipe_log = Arc::new(builder.finish()?);
         let rewrite = _rewrite.unwrap();
         let append = _append.unwrap();
@@ -332,7 +335,10 @@ where
         };
         let mut builder = FilePipeLogBuilder::new(cfg, file_builder, Vec::new());
         builder.scan()?;
-        let (append, rewrite) = builder.recover_or_truncate::<ConsistencyChecker>(None)?;
+        let machine_builder = ReplayMachineBuilder {
+            truncate_params: None,
+        };
+        let (append, rewrite) = builder.recover::<ConsistencyChecker>(&machine_builder)?;
         let mut map = rewrite.unwrap().finish();
         for (id, index) in append.unwrap().finish() {
             map.entry(id).or_insert(index);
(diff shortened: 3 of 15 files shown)


yuqi1129:
@tabokie Can you help to review this PR , unit test will be added later.  I have some problem about this issue 
1. Is this PR is on the right way ? if not, can you give me some advice, help wanted
2. Parameter `raft_groups` have not been used in `unsafe_truncate_with_file_builder`, i'm thinking how to use it and i have no idea whether we should use it in the `replay` method.
3. In the replay method, I'am using rename original file to tmp and move new file to the original file. this seems to be not atomic and we can't guarantee both is ok
4. how to deal with the original file if we successfully truncate it? just remove it?

yuqi1129:
5. I can't clearly understand the truncate mode `front`, if a file corrupted at beginning, we can't get the position where corrupt contents ends and normal logs starts

tabokie:
`truncate` recovers the data files to a bootable, consistent state. For a raft group with a missing log entry, we can accomplish that either by removing all the entries older than it ("front" mode), or removing all entries newer than it ("back" mode), or removing the whole raft group ("all" mode).

And yes, raft group id is needed, because the user might adopt different modes for different groups.

The code you just submit isn't good enough. You have to identify the "hole" in log entries by maintaining a map (id -> last_index) in `TruncateMachine`.

About the file rename issue, you can directly rename from the new filename to the original filename. This operation is atomic, and the old file is deleted.

yuqi1129:
@tabokie 

> truncate recovers the data files to a bootable, consistent state. For a raft group with a missing log entry, we can accomplish that either by removing all the entries older than it ("front" mode), or removing all entries newer than it ("back" mode), or removing the whole raft group ("all" mode).

The unit is a file or a raft groups? , If we meet corruption in a file and we choose `front` mode, Does this mean that we should drop all the message in the file that older than the corrupted message?
Another issue. `Truncate` action needs to guarantee that files names stay the same that is if there are 2 files 
A.write
B.write
After we execute the truncate command. there should be 2 files and one is A.write thats contains the content of `A.write` before? or we can merge this two file into one says C.write. I'am very puzzled at this point.

> And yes, raft group id is needed, because the user might adopt different modes for different groups.

If user only interested in some raft groups,  that means we only need to care about that groups? what if corruptions lies in other raft groups when we visited log files, how do we handle this scenarios

> The code you just submit isn't good enough. You have to identify the "hole" in log entries by maintaining a map (id -> last_index) in TruncateMachine.

You mean we should check whether the log index is sequential，what if some log has lost. For example, log index of a raft groups is 1, 2,3, 5 and 4 is lost, if the mode is `front`, we should drop 1,2,3 ?


I'am looking forward your replay, thanks

tabokie:
- We scan for all data files, but only modify specific raft groups. Only log entries in the corrupted raft group can be dropped.
- File name must be the same, because we assume file numbers are consecutive during recovery. This mean we will even write an empty file if all entries inside are truncated.
- Physical corruptions are already handled in `replay` method (via `TolerateAnyCorruption` mode), logical corruptions in unrelated raft groups will not be detected or handled.

yuqi1129:
>  Physical corruptions are already handled in replay method (via TolerateAnyCorruption mode), logical corruptions in unrelated raft groups will not be detected or handled.

That is exactly what makes me confused, as you have mention, Physical corruption have been handle,  so `truncate` subcomand only needs to handle logical corruption.  Because if there is physical corruption in a file, we can't do nothing but drop the rest of it.  As for logical corruption, say the log of one raft groups in a file is:
1,2, 3 ,4 6, 7 ,8,

- if mode is `front`, we should keep 1,2,3,4
- if mode is `back` we should keep 6,7,8
- if `mode` is `all`, we would drop all logs in the file of a specific raft groups 

That' is quite clear,  @tabokie Is there anything wrong with my unstanding?

tabokie:
@yuqi1129 Your understanding is correct.

yuqi1129:
@tabokie Can you take time to review this PR?

src/truncate.rs
```diff
@@ -0,0 +1,210 @@
...
+}
+
+#[derive(Clone, Copy)]
+pub struct TruncateQueueParameter<'a> {
+    pub queue: Option<LogQueue>,
+    pub truncate_mode: TruncateMode,
+    pub raft_groups_ids: &'a [u64],
+}
+
+impl Default for TruncateMachine {
```
tabokie, line 52: Refactor and export the file lock inside "file_pipe_log" module, then lock the directory when creating new truncate machine.

src/truncate.rs
```diff
@@ -0,0 +1,210 @@
...
+}
+
+impl TruncateMachine {
+    fn flush_to_new_file<B: FileBuilder>(
+        &mut self,
+        log_writer: &mut LogFileWriter<B>,
+        log_reader: &mut LogFileReader<B>,
+        truncate_params: &TruncateQueueParameter,
+    ) -> crate::Result<()> {
+        let mut batch = LogBatch::default();
```
tabokie, line 100: Limit the size of one log batch.

src/truncate.rs
```diff
@@ -0,0 +1,210 @@
...
+                        log_batch::OpType::Del => batch.delete(raft_id, kv.key.clone()),
+                    },
+                }
+            }
+        }
+
+        let raft_group_ids = truncate_params.raft_groups_ids;
+        let truncate_mode = &truncate_params.truncate_mode;
+
+        for (k, v) in region_id_to_index.iter() {
```
tabokie, line 140: I prefer to do the "corruption detection" inside `replay` function. Most of the file might not need rewriting at all. Also, you can assume there is no corruption inside one log item. This might help you simplify the code.

src/truncate.rs
yuqi1129: `LogBatch::default()` will make the size of log batch is `DEFAULT_LOG_ITEM_BATCH_CAP`(64),  Shall we need to change it explictly?

src/truncate.rs
```diff
@@ -0,0 +1,210 @@
...
+    Front = 0,
+    Back = 1,
+    All = 2,
+}
+
+impl FromStr for TruncateMode {
+    type Err = Error;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match s.to_lowercase().as_str() {
```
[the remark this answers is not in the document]
yuqi1129: 🆗

src/memtable.rs
```diff
@@ -769,6 +782,15 @@ impl ReplayMachine for MemTableRecoverContext {
         self.memtables.merge_newer_neighbor(&mut rhs.memtables);
         Ok(())
     }
+
+    fn end<B: FileBuilder>(
```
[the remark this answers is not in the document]
yuqi1129: I mistakenly think that there is no default implementation mechanism in rust trait like java `default` in interface. thanks

src/truncate.rs
tabokie: I'm referring to the actual written size of one log batch. E.g. we limit the batch under `MAX_REWRITE_BATCH_BYTES` when doing purge rewrite.

src/truncate.rs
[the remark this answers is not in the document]
yuqi1129: Seems we can only move the convert logic( string to TruncateMode) to `ctl.rs` and can't move the struct and related method to ctl.rs. Other logic rely on `TruncateMode` in `raft-engine`

src/truncate.rs
yuqi1129: 👌

src/truncate.rs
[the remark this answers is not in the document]
tabokie: That's exactly what I mean.

src/truncate.rs
yuqi1129: "corruption detection" logic can indeed do inside `replay` function. The reason that i do not do it is we lack the essential parameters in the method for example `raft_groups_ids` and `truncate_mode`, Changing method `replay` signature is costly.   About this problem,  do you have some ideas?

tabokie: As I commented earlier, you should pass in those necessary information when calling `TruncateMachine::new`.

src/memtable.rs
```diff
@@ -342,6 +344,17 @@ impl MemTable {
                 if allow_hole {
                     self.unsafe_truncate_back(first, 0, last);
                 } else {
+                    #[cfg(feature = "failpoints")]
```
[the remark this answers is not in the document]
yuqi1129: Sorry, i was the first time to use failpoint in rust. what's the right way to enable hole in the raft log? As far as i seen,  In real logic we would not make feature `failpoints ` enabled, So I was very puzzled that how this alter the existing code logic

src/truncate.rs
yuqi1129: Currently TruncateMachine and other struct that implements `ReplayMachine` was created by `default` method, how do we pass other information without change the `default` method? 
https://github.com/tikv/raft-engine/blob/89c3f39911b38dd2dcc9f9aa348187cc73ca1aaa/src/file_pipe_log/pipe_builder.rs#L221-L225

src/file_pipe_log/pipe_builder.rs
```diff
@@ -125,7 +133,10 @@ impl<B: FileBuilder> DualPipesBuilder<B> {
         Ok(())
     }
 
-    pub fn recover<S: ReplayMachine>(&mut self) -> Result<(S, S)> {
+    pub fn recover_or_truncate<S: ReplayMachine>(
```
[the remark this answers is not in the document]
yuqi1129: Can you give some advice, I know this is very ugly before i do it, but I was reluctant to add a new `truncate` method that has with almost the same code like `recover`

src/consistency.rs
```diff
@@ -1,11 +1,14 @@
 // Copyright (c) 2017-present, PingCAP, Inc. Licensed under Apache-2.0.
 
 use hashbrown::HashMap;
+use std::path::Path;
+use std::sync::Arc;
```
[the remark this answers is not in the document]
yuqi1129: 🆗

src/memtable.rs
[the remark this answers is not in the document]
tabokie: The purpose here is to create a hole in data files, but not in memtable. You can use failpoint "corrupted_items" just like I did in `test_consistency_tools`.

src/truncate.rs
tabokie: You are right, I forgot about this. Maybe pass in a replay machine factory:
```
pub trait ReplayMachineFactory<M: ReplayMachine> {
  fn new_machine(&self) -> M;
}
```

yuqi1129: This can create `ReplayMachine`  instance indeed, but i'm afraid that we should still add params `TruncateQueueParameter` which comes from caller to `new_machine` method to create `TruncateMachine` instance,  It's seems to be not very elegant. 

In other words, if we do not add a `truncate` method to replace `recover` ,  information about truncate action have to be given in `recovery` which is not very desirable. So what about add a new `truncate` method?

tabokie: You can pass the information to factory.

yuqi1129: @tabokie 
image

Is this elegant enough,  if it's OK, i will continue my work

tabokie: No it's still the same, What I meant before is you can pass in an actual factory (not just a generic type), which contains all the information.

yuqi1129:
@tabokie All discussion has been solved, can you help to view again?

src/engine.rs
```diff
@@ -77,8 +78,12 @@ where
         let start = Instant::now();
         let mut builder = FilePipeLogBuilder::new(cfg.clone(), file_builder, listeners.clone());
         builder.scan()?;
-        let (append, rewrite) = builder.recover::<MemTableRecoverContext>()?;
+        let (_append, _rewrite) =
+            builder.recover::<MemTableRecoverContext>(&ReplayMachineBuilder::default())?;
```
tabokie, line 82: First, only accept a builder, no need to pass in extra generic argument like `MemTableRecoverContext`.
```
pub trait ReplayMachineFactory {
  type Machine: ReplayMachine;
  fn new_machine(&self) -> Self::Machine;
}
// for FilePipeLogBuilder
fn recover<F: ReplayMachineFactory>(factory: F) {
  // details //
}
```

Second, add a default factory implementation for those can be constructed via default:
```
struct DefaultMachineFactory<T: ReplayMachine + Default>;
impl<T: ReplayMachine + Default> ReplayMachineFactory for DefaultMachineFactory {
  type Machine = T;
  fn new_machine(&self) -> T { T::default() }
}
// in engine.rs
let (append, rewrite) = builder.recover(DefaultMachineFactory<MemTableRecoverContext>)?;
```

src/log_batch.rs
```diff
@@ -90,14 +90,29 @@ impl EntryIndexes {
...
+        {
+            let enable_hole = || {
+                fail::fail_point!("memtable::enable_hole", |_| true);
+                false
+            };
+
+            if enable_hole() {
+                count -= 1;
+            }
+        }
```
tabokie, line 106: Why still add a new failpoint?

yuqi1129: You mean we can use the name `corrupted_items` instead of create a new name ? or we don't need to add a failpoint at all and can add some logic in the failpoint `corrupted_items` ?

tabokie: I mean why the failpoint "corrupted_items" isn't enough for your use case?

yuqi1129: On the one size `corrupted_items`  is intended for physical corruption and I think logical hole is different from that.  On the other hand. It's hard to modify byte array to create a logical hole as all content has been serialized to byte array (We may need to move the byte array content to remove the item that you want to delete) So i add a new failedpoint to mock the scenario that the end index of each `EntryIndexes` is not continuous.

Correct me if i'm wrong, thanks

tabokie: You can only write one log entry in one log batch, and corrupt that log batch. Also, corruption of multiple log entries (inside one log batch) still counts as a logical hole, I don't see why you can't test it.

yuqi1129:
@tabokie 
A new commit has been pushed,  if you have time, please to take a look to the latest code, thank you.

ctl/src/main.rs
```diff
@@ -112,7 +113,12 @@ impl ControlOpt {
         queue: &str,
         raft_groups: &[u64],
     ) -> EngineResult<()> {
-        Engine::unsafe_truncate(Path::new(path), mode, convert_queue(queue), raft_groups)
+        Engine::unsafe_truncate(
+            Path::new(path),
+            TruncateMode::from_str(mode)?,
```
tabokie, line 118: How about `parse(mode)?`

src/file_pipe_log/pipe_builder.rs
```diff
@@ -143,24 +191,40 @@ impl<B: FileBuilder> DualPipesBuilder<B> {
...
-                    LogQueue::Rewrite,
-                    self.file_builder.clone(),
-                    self.cfg.recovery_mode,
-                    rewrite_concurrency,
-                    self.cfg.recovery_read_block_size.0 as usize,
-                    &self.rewrite_files,
-                )
+                if truncate_params.is_none()
+                    || truncate_params.as_ref().unwrap().queue.is_none()
+                    || truncate_params.as_ref().unwrap().queue.unwrap() == LogQueue::Append
```
tabokie, line 214: Add interface like `fn buildable(queue: LogQueue) -> bool` to MachineBuilder, instead of these very specific conditions.

src/file_pipe_log/pipe_builder.rs
```diff
@@ -125,8 +169,12 @@ impl<B: FileBuilder> DualPipesBuilder<B> {
         Ok(())
     }
 
-    pub fn recover<S: ReplayMachine>(&mut self) -> Result<(S, S)> {
+    pub fn recover<S: ReplayMachine>(
+        &mut self,
+        machine_builder: &ReplayMachineBuilder<S>,
```
tabokie, line 174: Refer to https://github.com/tikv/raft-engine/pull/175#discussion_r772969268
```suggestion
    pub fn recover<F: ReplayMachineFactory>(
        &mut self,
        machine_builder: &F,
```

src/file_pipe_log/pipe_builder.rs
```diff
@@ -13,17 +14,74 @@ use crate::event_listener::EventListener;
...
+
+    fn init(&mut self, _params: Option<TruncateQueueParameter>) {}
+}
+
+pub trait ReplayMachineFactory: Sync {
+    type Machine: ReplayMachine;
+
+    fn new_machine(&self) -> Self::Machine;
+
+    fn parameters(&self) -> Option<TruncateQueueParameter>;
```
tabokie, line 42: Remove this, internal parameters should be exposed.

src/file_pipe_log/pipe_builder.rs
```diff
@@ -13,17 +14,74 @@ use crate::event_listener::EventListener;
...
+}
+
+impl<M: ReplayMachine> Default for ReplayMachineBuilder<M> {
+    fn default() -> Self {
+        ReplayMachineBuilder {
+            truncate_params: None,
+            _phantom: Default::default(),
+        }
+    }
+}
```
tabokie, line 65: I don't think you understood https://github.com/tikv/raft-engine/pull/175#discussion_r772969268, only `TruncateMachine` has a special builder, other machines that can be constructed by `default` share one default builder.

src/file_pipe_log/pipe_builder.rs
```diff
@@ -143,24 +206,34 @@ impl<B: FileBuilder> DualPipesBuilder<B> {
...
             || {
-                Self::recover_queue(
-                    LogQueue::Append,
-                    self.file_builder.clone(),
-                    self.cfg.recovery_mode,
-                    append_concurrency,
-                    self.cfg.recovery_read_block_size.0 as usize,
-                    &self.append_files,
-                )
+                if machine_builder.buildable(LogQueue::Append) {
```
tabokie, line 209: I still don't quite like the idea of returning `Option`. How about adding another interface `replay_queue` that does single queue recovery?

src/file_pipe_log/pipe_builder.rs
yuqi1129: OK, I have rename the method and return the boolean value that whether we should do truncate operation.

src/file_pipe_log/pipe_builder.rs
yuqi1129: Sounds reasonable if returning `Option` is not advised in Rust. I do this just for simplicity.

tabokie: It's not specific to rust. Returning `Option` is bug-prone, we tend to unwrapping it on caller side based on assumption that could be broken at some corner cases.

yuqi1129: @tabokie  OK, I have update the code, Is it elegant enough about this point ?

tabokie: Looks good, https://github.com/tikv/raft-engine/pull/175#discussion_r775937403 is still unresolved. I'll do a thorough review after you address that.

yuqi1129: @tabokie , I have thought this for a long time, can you help to confirm if the following code is what you have refer to?


```rust
impl ReplayMachineFactory for ReplayMachineBuilder<ConsistencyChecker> {
    type Machine = ConsistencyChecker;
}

impl ReplayMachineFactory for ReplayMachineBuilder<MemTableRecoverContext> {
    type Machine = MemTableRecoverContext;
}

impl ReplayMachineFactory for ReplayMachineBuilder<TruncateMachine> {
    type Machine = TruncateMachine;

    fn new_machine(&self) -> TruncateMachine {
        let mut m = TruncateMachine::default();
        m.init(self.truncate_params.clone());
        m
    }

    fn need_truncate(&self) -> bool {
        self.truncate_params.is_some()
    }

    fn buildable(&self, queue: LogQueue) -> bool {
        return self.truncate_params.is_none()
            || self.truncate_params.as_ref().unwrap().queue.is_none()
            || self.truncate_params.as_ref().unwrap().queue.unwrap() == queue;
    }
}
```

The following code seems more elegant, But there exists error when compiling.
```rust
impl<M: ReplayMachine> ReplayMachineFactory for ReplayMachineBuilder<M> {
    type Machine = M;
}

impl ReplayMachineFactory for ReplayMachineBuilder<TruncateMachine> {
    type Machine = TruncateMachine;

    fn new_machine(&self) -> TruncateMachine {
        let mut m = TruncateMachine::default();
        m.init(self.truncate_params.clone());
        m
    }

    fn need_truncate(&self) -> bool {
        self.truncate_params.is_some()
    }

    fn buildable(&self, queue: LogQueue) -> bool {
        return self.truncate_params.is_none()
            || self.truncate_params.as_ref().unwrap().queue.is_none()
            || self.truncate_params.as_ref().unwrap().queue.unwrap() == queue;
    }
```

tabokie: Like this:
```
pub type DefaultMachineFactory<M> = PhantomData<M>;

impl<M: Default + ReplayMachine> ReplayMachineFactory for DefaultMachineFactory<M> {
  type Machine = M;
  fn new_machine(&self) -> Self::Machine { M::default() }
}

pub struct TruncateMachineFactory {
  // parameters
}

impl ReplayMachineFactory for ReplayMachineBuilder {
    type Machine = TruncateMachine;

    fn new_machine(&self) -> TruncateMachine {
        TruncateMachine::new(self.params)
    }
}
```

And remove the use of `need_truncate` and `buildable`. You can add a new interface `recover_queue` to the PipeBuilder.

yuqi1129: trait `need_truncate` and `buildable` have been removed and I add the corresponding method in `DualPipesBuilder`, I'am not sure if this is what you thought

src/file_pipe_log/pipe_builder.rs
```diff
@@ -49,9 +175,18 @@ impl<B: FileBuilder> DualPipesBuilder<B> {
             listeners,
             append_files: Vec::new(),
             rewrite_files: Vec::new(),
+            extra_params: None,
         }
     }
 
+    pub fn set_truncate_params(&mut self, truncate_params: Option<TruncateQueueParameter>) {
```
tabokie, line 182: I'm a bit confused why is it so difficult to hide params from the builder. You can determine which queue to replay before entering `recover_queue`.

src/file_pipe_log/pipe_builder.rs
```diff
@@ -39,6 +67,104 @@ pub struct DualPipesBuilder<B: FileBuilder> {
 
     append_files: Vec<FileToRecover>,
     rewrite_files: Vec<FileToRecover>,
+    extra_params: Option<TruncateQueueParameter>,
+}
+
+trait RecoverQueue {
```
tabokie, line 73: Is it necessary? Trait only exists to help us write generic code, I don't see any use of it currently.

src/file_pipe_log/pipe_builder.rs
yuqi1129: > You can determine which queue to replay before entering recover_queue 

Yes, it can be determined before entering `recover_queue `, as `recover` return `Result<ReplayMachine>`, maybe we should write as follower

```
if should_recover_queue() {
}
else {
   Ok(machine)
}
```
I think it's not very nice to add the else block with simple return Ok,  so i put the logic into `recover_queue`


> I'm a bit confused why is it so difficult to hide params from the builder 

I can't get your idea clearly. you mean we should add `truncate_paramter` into default constructor of `Builder` or it's not wise to place the `truncate_paramter` in `DualPipesBuilder` ?

src/file_pipe_log/pipe_builder.rs
yuqi1129: No necessary, As you have commented, it's advise to add a `recover_queue` trait, so i do. before I push the commit, I have noticed that the trait have not used. so i ask for your help when i pushed the commit

src/file_pipe_log/pipe_builder.rs
tabokie: Builder shouldn't know about any parameters specific to the machine type. You should figure out the log queue and call corresponding `recover_queue` in `Engine::truncate`.

src/file_pipe_log/pipe_builder.rs
tabokie: Oh, I didn't mean trait by "interface". A method is enough.

src/file_pipe_log/pipe_builder.rs
```diff
@@ -13,17 +14,69 @@ use crate::event_listener::EventListener;
...
 use super::reader::LogItemBatchFileReader;
 
-pub trait ReplayMachine: Send + Default {
-    fn replay(&mut self, item_batch: LogItemBatch, file_id: FileId) -> Result<()>;
+pub trait ReplayMachine: Send + Default + Sync {
+    fn replay(&mut self, item_batch: LogItemBatch, file_id: FileId, path: &Path) -> Result<()>;
 
     fn merge(&mut self, rhs: Self, queue: LogQueue) -> Result<()>;
+
+    fn truncate<B: FileBuilder>(
```
tabokie, line 30: I presume this is unneeded?

src/file_pipe_log/pipe_builder.rs
```diff
@@ -13,17 +14,69 @@ use crate::event_listener::EventListener;
...
+}
+
+pub trait ReplayMachineFactory: Sync {
+    type Machine: ReplayMachine;
+
+    fn new_machine(&self) -> Self::Machine {
+        Self::Machine::default()
+    }
+
+    fn truncate<B: FileBuilder>(&self, _: &mut Self::Machine, _: Arc<B>) -> Result<()> {
```
tabokie, line 46: This too, you can hold all the necessary information in factory itself.

yuqi1129: @tabokie After we move the truncate logic to `TruncateMachine`,  the `DefaultMachineFactory` has no function and seems to be useless, It's wise to remove it and revert to original code ?

src/engine.rs
```diff
@@ -343,15 +346,85 @@ where
...
+            queue,
+            truncate_mode: mode,
+            raft_groups_ids: raft_groups.to_vec(),
+            lock_file: Arc::new(get_lock_file(path.to_str().unwrap())?),
+        };
+
+        let mut builder = FilePipeLogBuilder::new(cfg.clone(), file_builder.clone(), Vec::new());
+        builder.scan()?;
+
+        let threads = cfg.recovery_threads;
```
tabokie, line 379: I'm very confused about this part, why can't we reuse `recover_queue`?

tabokie: Okay, you can make it cleaner by removing the thread pool building code. We don't need all that parallelism for a debugging task like "truncate". Instead, you can call two `recover_queue` sequentially (and preferably passing in concurrency=1).

src/file_pipe_log/pipe_builder.rs
tabokie: I haven't reviewed your code for the truncate machine. But from the looks of it, you are currently stashing all the log items and truncating them at the end. But initially, I planned to do the truncate **while** reading log items, and doing it one file at a time. For that to happen, you need the truncate parameters during replay, and the customized machine factory.

tabokie: You don't need a new trait method to do the things I described. All parameters and all data can be accessed within the `replay` trait method. You can deduce the file boundary by looking at the `file_id` passed in through `replay` method.
And I don't see how this contradicts with the truncate logic. By "handle file one by one", I mean writing the truncated file immediately after processing the items in it.

yuqi1129: This the current logic that call replay method 
https://github.com/tikv/raft-engine/blob/517f3b7527412271d6adce969c966ac4567b0e2c/src/file_pipe_log/pipe_builder.rs#L231-L264

1. How can we get the information like queue name, `file_id` can only get the sequence number.  Of course We can put this information in the `TruncateMachine` when use `new_machine` in machine factory.

2. As far as i can see, `relay` method is used to handle one `LogItemBatch`,  As there would be a log of  `LogItemBatch` in one log file, How can we see which one is the last and then do the truncate logic.  The best way i think to do truncate logic in line 142(see above)

3. Handle file one by one, I think, there is still some problem. See the example I mentioned in past, the result will be 
- Front mode
A: 1, 2
B: 7

- Back mode
A:4, 5
B: 10

- All mode
A: (empty)
B: (empty)

Index of raft groups will still discontinuous in mode front and back. is it what we expected?

tabokie: 1. Yes, Putting truncate information into truncate machine was the initial plan.
2. "You can deduce the file boundary by looking at the file_id passed in through replay method." For sequential replay machine, you keep a record of last replayed file id, and when incoming file id == last replayed id + 1, you know it belongs to a new file.
3. There is no problem for back and all mode. For back mode you can maintain a record of last contiguous index, and delete all entries larger than it. But I do see a problem with front mode that I didn't notice before. You can't deduce the latest logical hole by scanning forward. How about we add a bool parameter `backward` to the `scan` method, so that we can scan the files backward?

yuqi1129: @tabokie 
> There is no problem for back and all mode

You mean result of mode back and all in https://github.com/tikv/raft-engine/pull/175#discussion_r779275658 is reasonable?, If so the only things left is implementation.

> But I do see a problem with front mode that I didn't notice before. You can't deduce the latest logical hole by scanning forward. How about we add a bool parameter backward to the scan method, so that we can scan the files backward?

Indeed, we need a method to traverse file reversely, let me have a try

tabokie: No, you can accomplish the old results (https://github.com/tikv/raft-engine/pull/175#discussion_r778799444) with new design. "For back mode you can maintain a record of last contiguous index, and delete all entries larger than it."

yuqi1129: @tabokie, Here is the situation
> 1. "You can deduce the file boundary by looking at the file_id passed in through replay method." For sequential replay machine, you keep a record of last replayed file id, and when incoming file id == last replayed id + 1, you know it belongs to a new file.

We may need another trait the flush the data of last file, As the content can only be wrote to file when handle the next file, we need a trait to save the last file. Is it element enough? or do you have any suggestions

2.  For `Back` mode we need a reverse iterator just as the `Front` mode, But for `All` mode, I'am afraid we should keep all data(items) in memory temporarily and check if there is a hole, Then we can safely save the items to log files. Is there any wrong with the point?

Looking forward your reply, thanks

tabokie:
Closed in favor of #182

PR #180, opened 2022-01-19
fix tombstone rewrite ordering

### LINKED ISSUE #142: Raft Group tombstone is rewritten in the wrong order
### Description

When a Raft Group is deleted (via `Clean` command), a tombstone is written to log file, and added to `MemTableAccessor`. During rewrite, log file holding that tombstone is purged, but tombstone in memtable still can be retrieved and rewritten to `Rewrite` queue.

However, consider a `Append` file containing these operations [append1 (..), clean, append2 (..)]. During rewrite, the tombstone is actually rewritten in the wrong order [append1 (..), append2 (..), clean]. This will cause the newly recreated Raft Group to disappear after restart.

TiKV guarantees to not recreate a region with the same ID, so this bug is of low severity.

tabokie:

Fix #142 

Fixed tombstone write ordering. Rewrite tombstone before rewriting incoming entries. Also fixed the issue where append tombstones are recorded twice (first by `apply` then by `merge_newer_neighbor`).


Little-Wallace (approved):
LGTM

PR #193, opened 2022-02-23
force rewrite memory tables when outdated log is not compact on time

glorv:
ref tikv/tikv#12011

### Problem Summary
In the current implementation, when raft-engine detect outdated log entries that will block log file purge, it will report these regions to tikv and expect tikv to do force compact. Since tikv only do force compact on region leader and ignores request on follower. There are cases that the raft log in leader store is rewrite by tikv but not on one or more follower. Then when the log on follower need to be compact, tikv will ignore the requests because it expect region leader to do so. If there are no newer log be written into this region, as leader's log is already rewritten, it will not trigger log purge. Thus, the old log on the follower will block log file purge.

### Changes
Add a HashMap to record each regions live generation after each purge phase. If a region lives after three times of force compact, raft engine will force rewrite this region to avoid blocking log file purges.


src/purge.rs
```diff
@@ -23,6 +24,7 @@ const REWRITE_RATIO: f64 = 0.7;
 // Only rewrite region with stale logs less than this threshold.
 const MAX_REWRITE_ENTRIES_PER_REGION: usize = 32;
 const MAX_REWRITE_BATCH_BYTES: usize = 128 * 1024;
+const MAX_COUNT_BEFORE_FORCE_WRITE: u32 = 3;
```
tabokie, line 27:
```suggestion
const MAX_COUNT_BEFORE_FORCE_REWRITE: u32 = 3;
```

src/purge.rs
```diff
@@ -35,7 +37,11 @@ where
     listeners: Vec<Arc<dyn EventListener>>,
 
     // Only one thread can run `purge_expired_files` at a time.
-    purge_mutex: Arc<Mutex<()>>,
+    //
+    // the hashmap(raft_group --> count) records raft groups that should be force compact.
+    // if target reft_group are still in the oldest 20% logs after `MAX_COUNT_BEFORE_FORCE_WRITE`
+    // times purge, it will be force rewrite so the logs can be purged.
+    pending_regions: Arc<Mutex<HashMap<u64, u32>>>,
```
tabokie, line 44: How about `force_rewrite_candidates`?

src/purge.rs
```diff
@@ -35,7 +37,11 @@ where
     listeners: Vec<Arc<dyn EventListener>>,
 
     // Only one thread can run `purge_expired_files` at a time.
-    purge_mutex: Arc<Mutex<()>>,
+    //
+    // the hashmap(raft_group --> count) records raft groups that should be force compact.
+    // if target reft_group are still in the oldest 20% logs after `MAX_COUNT_BEFORE_FORCE_WRITE`
+    // times purge, it will be force rewrite so the logs can be purged.
```
tabokie, line 43:
```suggestion
    // This table records Raft Groups that should be force compacted before. Those that are not compacted in time (after `MAX_EPOCH_BEFORE_FORCE_REWRITE` epochs) will be force rewritten.
```

tabokie (commented):
Two more things:
- Adjust the logic in `rewrite_rewrite_queue`: make Raft Groups with more than `MAX_REWRITE_ENTRIES_PER_REGION` rewrite entries be force compacted as well.
- Add a unit test in "engine.rs" to make sure laggers are indeed rewritten.

Oh, and you need to signed off your commit. Maybe `git commit --amend --signoff && git push -f`.

src/purge.rs
tabokie: Let's make it 9. 2m window is good enough.

src/engine.rs
```diff
@@ -966,6 +966,90 @@ mod tests {
...
+        // write some small entries to trigger purge, some of the should trigger rewrite
+        // but not compact.
+        for rid in 21..100 {
+            engine.append(rid, 1, 2, Some(&data));
+        }
+
+        let check_purge = |pending_regions: Vec<u64>| {
+            let res = engine.purge_expired_files();
+            assert!(res.is_ok());
+            let mut compact_regions = res.unwrap();
```
tabokie, line 1007:
```suggestion
            let mut compact_regions = engine.purge_expired_files().unwrap();
```

src/engine.rs
```diff
@@ -966,6 +966,90 @@ mod tests {
...
+        // purge 3rd.
+        check_purge(vec![1, 2]);
+        let span3 = engine.file_span(LogQueue::Append);
+        assert_eq!(span3, span);
+
+        // pruge the 4th time, it will trigger force rewrite and thus the result should
+        // be empty, and some log files should be purged.
+        check_purge(vec![]);
+        let span4 = engine.file_span(LogQueue::Append);
+        assert!(span4.0 > span.0 + 5);
```
tabokie, line 1037: This is a little verbose. Could you use a for loop that break when `purge_expired_files` returns empty list, and panic when the loop ends without encountering an empty list.

src/engine.rs
```diff
@@ -966,6 +966,90 @@ mod tests {
...
+
+        let engine =
+            RaftLogEngine::open_with_file_system(cfg, Arc::new(ObfuscatedFileSystem::new()))
+                .unwrap();
+        let data = vec![b'x'; 1024];
+        // write 50 small entries into region 1~3, it should trigger force compact
+        // report.
+        for rid in 1..4 {
+            for index in 0..50 {
+                engine.append(rid, index, index + 1, Some(&data[..20]));
```
tabokie, line 990: why use small entry?

src/engine.rs
```diff
@@ -966,6 +966,90 @@ mod tests {
...
+        for rid in 4..20 {
+            for index in 0..25 {
+                engine.append(rid, index, index + 1, Some(&data[..50]));
+            }
+        }
+        // write some small entries to trigger purge, some of the should trigger rewrite
+        // but not compact.
+        for rid in 21..100 {
+            engine.append(rid, 1, 2, Some(&data));
+        }
```
tabokie, line 1002: Maybe we can get the work done with only one type of region writes.

src/purge.rs
```diff
@@ -23,6 +24,7 @@ const REWRITE_RATIO: f64 = 0.7;
 // Only rewrite region with stale logs less than this threshold.
 const MAX_REWRITE_ENTRIES_PER_REGION: usize = 32;
 const MAX_REWRITE_BATCH_BYTES: usize = 128 * 1024;
+const MAX_COUNT_BEFORE_FORCE_REWRITE: u32 = 3;
```
tabokie, line 27:
```suggestion
const MAX_COUNT_BEFORE_FORCE_REWRITE: u32 = 9;
```

src/engine.rs
glorv: sparse region will trigger rewrite instead of compact.

tabokie: But `sparse` is determined by count of entries, not size of entries.

tabokie:
@glorv The last commit is not signed.

src/engine.rs
glorv: Yes. small entries is used for the convenience of triggering compact and rewrite as the total log size is smaller. I want each region takes about one log file.

PR #203, opened 2022-03-24
update global stats when memtable is dropped

glorv:
ref tikv/tikv#12158

### DIFF
--- a/src/memtable.rs
+++ b/src/memtable.rs
@@ -643,8 +643,7 @@ impl Drop for MemTable {
     fn drop(&mut self) {
         let mut append_kvs = 0;
         let mut rewrite_kvs = 0;
-        let kvs = std::mem::replace(&mut self.kvs, BTreeMap::default());
-        for (_k, (_v, id)) in kvs.into_iter() {
+        for (_k, (_v, id)) in &self.kvs {
             match id.queue {
                 LogQueue::Rewrite => rewrite_kvs += 1,
                 LogQueue::Append => append_kvs += 1,
--- a/src/memtable.rs
+++ b/src/memtable.rs
@@ -107,6 +107,7 @@ impl MemTable {
             );
             self.rewrite_count += rhs.rewrite_count;
             self.entry_indexes.append(&mut rhs.entry_indexes);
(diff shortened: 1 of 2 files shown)


src/memtable.rs
```diff
@@ -107,6 +107,7 @@ impl MemTable {
             );
             self.rewrite_count += rhs.rewrite_count;
             self.entry_indexes.append(&mut rhs.entry_indexes);
+            rhs.rewrite_count = 0;
```
tabokie, line 110: No need to change it.

src/memtable.rs
```diff
@@ -638,6 +639,27 @@ impl MemTable {
...
+    fn drop(&mut self) {
+        let mut append_kvs = 0;
+        let mut rewrite_kvs = 0;
+        let kvs = std::mem::replace(&mut self.kvs, BTreeMap::default());
+        for (_k, (_v, id)) in kvs.into_iter() {
+            match id.queue {
+                LogQueue::Rewrite => rewrite_kvs += 1,
+                LogQueue::Append => append_kvs += 1,
+            }
+        }
```
tabokie, line 652: I think you can directly `iter` through it without these maneuvers.

tabokie (commented):
Thanks a lot!

src/memtable.rs
tabokie: I see. Then I'm skeptical of why your change works. Because `Vec::append` will remove all entries in `rhs.entry_indexes`, that append count calculation inside drop incorrect.

glorv: Yes. Memable drop occurs in two scenario: 1) at bootstrap, memtables merge with each other. I this case, The drop is useless because the GlobalStats is not shared. 2) `Command::clean`, which will remote target memable. In this case we need to do this clean up by Memable::drop

tabokie: Ok. How about adding a `clear` method to MemTable instead of doing it in drop?

tabokie: Actually, forget it. It's not a good idea.

tabokie (approved):

Issue #205, opened 2022-03-31
Reduce memory footprint


tabokie:
Memtables can take up a lot of memory when log entries are not compacted in time.

- [x] Add memory tracing
- [ ] Reuse file block handle for entry indexes inside the same log batch.
- [x] Change entry offset and entry length to `u32`
- [x] Store entry offset in file instead of in memory
- [x] Implement a mmap-based allocator for memtable

PR #206, opened 2022-03-31
Reduce size of `EntryIndex`

tabokie:
Ref #205

Some enhancements:
- Do not store log index inside `EntryIndex`.
- Reduce entry offset and entry len from u64 to u32. It won't cause regression because the length is already limited by compression (i32::MAX).

Size for one `EntryIndex` is reduced from 64 bytes to 48 bytes (25%).

stress benchmarks shows 18.7% improvement (compared to results in #207, normalized with write bandwidth):
```
Memory: 7003MiB
[write]
Throughput(QPS) = 32322.92
Latency(μs) min = 8, avg = 16.10, p50 = 13, p90 = 26, p95 = 31, p99 = 45, p99.9 = 79, max = 4867
Fairness = 100.0%
Write Bandwidth = 24.2MiB/s
```

### DIFF
--- a/src/memtable.rs
+++ b/src/memtable.rs
@@ -53,6 +53,59 @@ impl Default for EntryIndex {
     }
 }
 
+impl EntryIndex {
+    fn from_thin(index: u64, e: ThinEntryIndex) -> Self {
+        Self {
+            index,
+            entries: e.entries,
+            compression_type: e.compression_type,
+            entry_offset: e.entry_offset,
+            entry_len: e.entry_len,
+        }
+    }
+}
+
+#[derive(Debug, Copy, Clone, PartialEq)]
+struct ThinEntryIndex {
+    entries: Option<FileBlockHandle>,
+    compression_type: CompressionType,
+    entry_offset: u32,
+    entry_len: u32,
+}
+
+impl Default for ThinEntryIndex {
+    fn default() -> ThinEntryIndex {
+        ThinEntryIndex {
+            entries: None,
+            compression_type: CompressionType::None,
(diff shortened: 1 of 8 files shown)


src/memtable.rs
```diff
@@ -615,13 +683,13 @@ impl MemTable {
...
         let mut seen_append = false;
-        let mut last_index = None;
+        // let mut last_index = None;
         for idx in self.entry_indexes.iter() {
             // Check 1: indexes are contiguous.
-            if let Some(last_index) = last_index {
-                assert_eq!(idx.index, last_index + 1);
-            }
-            last_index = Some(idx.index);
+            // if let Some(last_index) = last_index {
```
Connor1996, line 689: why comment out?

src/memtable.rs
```diff
@@ -36,9 +36,9 @@ pub struct EntryIndex {
     pub compression_type: CompressionType,
 
     /// The relative offset within its group of entries.
-    pub entry_offset: u64,
+    pub entry_offset: u32,
```
Connor1996, line 39: wrap `ThinEntryIndex` directly?

Connor1996 (commented):
Could you describe what you change to reduce the size

src/memtable.rs
tabokie: No need for this check because entry index doesn't store log index anymore.

src/memtable.rs
tabokie: No because `ThinEntryIndex` should be a private struct internal to `MemTable`.

PR #207, opened 2022-04-01
trace memory of MemTables

tabokie:

Ref #205 

Added a new method to flush internal metrics, including memory used by `MemTables`.

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -3,8 +3,9 @@
 use std::cell::{Cell, RefCell};
 use std::marker::PhantomData;
 use std::path::Path;
-use std::sync::Arc;
-use std::time::Instant;
+use std::sync::{mpsc, Arc, Mutex};
+use std::thread::{Builder as ThreadBuilder, JoinHandle};
+use std::time::{Duration, Instant};
 
 use log::{error, info};
 use protobuf::{parse_from_bytes, Message};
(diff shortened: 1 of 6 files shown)


tabokie:
I ran a benchmark to get a baseline memory usage:
```
# ./target/release/stress --path /data3/re_bench --time 60 --regions 1000 --write-threads 1 --purge-interval 0 --compact-count 0 --write-without-sync
Memory usage: 8003MiB
```

src/memtable.rs
```diff
@@ -902,6 +908,16 @@ impl MemTableAccessor {
         }
     }
 
+    pub(crate) fn flush_metrics(&self) {
```
Connor1996, line 911: let the user to call `flush_metrics` instead of doing itself?

src/engine.rs
```diff
@@ -92,6 +98,19 @@ where
             listeners.clone(),
         );
 
+        let (tx, rx) = mpsc::channel();
+        let stats_clone = stats.clone();
+        let memtables_clone = memtables.clone();
+        let metrics_flusher = ThreadBuilder::new()
+            .name("raft-engine-metrics".into())
+            .spawn(move || loop {
```
Connor1996, line 106: let the user to call flush_metrics instead of doing itself?

tabokie: Okay.

PR #211, opened 2022-04-08
support memory swap out

tabokie:

Ref #205

An allocator that switches to file-based mmap when memory usage is above certain watermark.

Micro-benchmarks:
```
test swappy_allocator::tests::bench_allocator_fast_path     ... bench:         295 ns/iter (+/- 165)
test swappy_allocator::tests::bench_allocator_slow_path     ... bench:      92,075 ns/iter (+/- 12,995)
test swappy_allocator::tests::bench_allocator_std_global    ... bench:         252 ns/iter (+/- 41)
```

buffered write (`./target/release/stress --path /data3/re_bench --time 120 --regions 1000 --write-threads 1 --write-without-sync`):
```
master:
Throughput(QPS) = 17314.75
Latency(μs) min = 19, avg = 41.10, p50 = 31, p90 = 49, p95 = 57, p99 = 81, p99.9 = 2289, max = 5151
Fairness = 100.0%
Write Bandwidth = 181.3MiB/s

swap memory-limit=+inf:
Throughput(QPS) = 17458.37
Latency(μs) min = 18, avg = 40.93, p50 = 31, p90 = 48, p95 = 56, p99 = 79, p99.9 = 2241, max = 5663
Fairness = 100.0%
Write Bandwidth = 182.8MiB/s

swap memory-limit=0:
Throughput(QPS) = 11312.92
Latency(μs) min = 17, avg = 71.75, p50 = 40, p90 = 64, p95 = 77, p99 = 133, p99.9 = 7819, max = 50303
Fairness = 100.0%
Write Bandwidth = 118.4MiB/s
```

write (`./target/release/stress --path /data3/re_bench --time 120 --regions 1000 --write-threads 1 --purge-interval 0 --compact-count 0`):
```
master:
Throughput(QPS) = 7010.16
Latency(μs) min = 77, avg = 123.41, p50 = 117, p90 = 153, p95 = 166, p99 = 201, p99.9 = 991, max = 4057
Fairness = 100.0%
Write Bandwidth = 73.4MiB/s

swap memory-limit=0:
Throughput(QPS) = 4748.54
Latency(μs) min = 85, avg = 191.08, p50 = 169, p90 = 256, p95 = 291, p99 = 520, p99.9 = 2339, max = 29887
Fairness = 100.0%
Write Bandwidth = 49.7MiB/s

swap memory-limit=1GB:
Throughput(QPS) = 6270.32
Latency(μs) min = 80, avg = 140.46, p50 = 124, p90 = 178, p95 = 201, p99 = 266, p99.9 = 2743, max = 23407
Fairness = 100.0%
Write Bandwidth = 65.6MiB/s
```

Test results on nvme disk:

### DIFF
--- a/src/memtable.rs
+++ b/src/memtable.rs
@@ -57,14 +57,21 @@ mod swap_conditional_imports {
 
 #[cfg(not(feature = "swap"))]
 mod swap_conditional_imports {
+    #[derive(Clone)]
+    pub struct DummyAllocator;
+
     pub trait AllocatorTrait: Clone + Send + Sync {}
-    impl AllocatorTrait for () {}
+    impl AllocatorTrait for DummyAllocator {}
 
-    pub type SelectedAllocator = ();
-    pub type VacantAllocator = ();
+    pub type SelectedAllocator = DummyAllocator;
+    pub type VacantAllocator = DummyAllocator;
 
-    pub fn new_vacant_allocator() -> VacantAllocator {}
-    pub fn new_allocator(_: &crate::Config) -> SelectedAllocator {}
+    pub fn new_vacant_allocator() -> VacantAllocator {
+        DummyAllocator
+    }
+    pub fn new_allocator(_: &crate::Config) -> SelectedAllocator {
+        DummyAllocator
+    }
 }
 
 use swap_conditional_imports::*;
--- a/src/memtable.rs
+++ b/src/memtable.rs
@@ -22,6 +22,7 @@ use crate::{Error, GlobalStats, Result};
 
 #[cfg(feature = "swap")]
 mod swap_conditional_imports {
+    use crate::swappy_allocator::SwappyAllocator;
     use log::error;
     use std::convert::TryFrom;
     use std::path::Path;
@@ -29,9 +30,8 @@ mod swap_conditional_imports {
     pub trait AllocatorTrait: std::alloc::Allocator + Clone + Send + Sync {}
     impl<T: std::alloc::Allocator + Clone + Send + Sync> AllocatorTrait for T {}
 
-    pub type SwappyAllocator = crate::swappy_allocator::SwappyAllocator<std::alloc::Global>;
-    pub type SelectedAllocator = SwappyAllocator;
     pub type VacantAllocator = std::alloc::Global;
+    pub type SelectedAllocator = SwappyAllocator<std::alloc::Global>;
 
     pub fn new_vacant_allocator() -> VacantAllocator {
         std::alloc::Global
@@ -40,31 +40,20 @@ mod swap_conditional_imports {
         let memory_limit =
             usize::try_from(cfg.memory_limit.map_or(u64::MAX, |l| l.0)).unwrap_or(usize::MAX);
         let path = Path::new(&cfg.dir).join("swap");
-        // Deletes the path if it exists.
-        if path.exists() {
-            if let Err(e) = std::fs::remove_dir_all(&path) {
-                error!(
-                    "Failed to clean up old swap directory: {:?}. \
-                    There might be obsolete swap files left in {}.",
-                    e,
-                    path.display()
-                );
-            }
-        }
         SwappyAllocator::new(&path, memory_limit)
     }
 }
 
 #[cfg(not(feature = "swap"))]
 mod swap_conditional_imports {
+    pub trait AllocatorTrait: Clone + Send + Sync {}
+
     #[derive(Clone)]
     pub struct DummyAllocator;
-
-    pub trait AllocatorTrait: Clone + Send + Sync {}
     impl AllocatorTrait for DummyAllocator {}
 
-    pub type SelectedAllocator = DummyAllocator;
     pub type VacantAllocator = DummyAllocator;
+    pub type SelectedAllocator = DummyAllocator;
 
     pub fn new_vacant_allocator() -> VacantAllocator {
         DummyAllocator
--- a/src/memtable.rs
+++ b/src/memtable.rs
@@ -2,12 +2,14 @@
 
 use std::borrow::BorrowMut;
 use std::collections::{BTreeMap, HashSet, VecDeque};
+use std::marker::PhantomData;
 use std::sync::Arc;
 
 use fail::fail_point;
 use hashbrown::HashMap;
 use parking_lot::{Mutex, RwLock};
 
+use crate::config::Config;
 use crate::file_pipe_log::ReplayMachine;
 use crate::log_batch::{
     Command, CompressionType, KeyValue, LogBatch, LogItemBatch, LogItemContent, LogItemDrain,
@@ -15,8 +17,44 @@ use crate::log_batch::{
 };
 use crate::metrics::MEMORY_USAGE;
 use crate::pipe_log::{FileBlockHandle, FileId, FileSeq, LogQueue};
-use crate::util::{hash_u64, slices_in_range};
+use crate::util::{hash_u64, Factory};
 use crate::{Error, GlobalStats, Result};
+
+#[cfg(feature = "swap")]
+mod swap_conditional_imports {
+    use std::convert::TryFrom;
+    use std::path::Path;
+
+    pub trait AllocatorTrait: std::alloc::Allocator + Clone {}
+    impl<T: std::alloc::Allocator + Clone> AllocatorTrait for T {}
+
+    pub type SwappyAllocator = crate::swappy_allocator::SwappyAllocator<std::alloc::Global>;
+    pub type SelectedAllocator = SwappyAllocator;
+    pub type VacantAllocator = std::alloc::Global;
+
+    pub fn new_vacant_allocator() -> VacantAllocator {
+        std::alloc::Global
+    }
+    pub fn new_allocator(cfg: &crate::Config) -> SelectedAllocator {
+        let memory_limit =
+            usize::try_from(cfg.memory_limit.map_or(u64::MAX, |l| l.0)).unwrap_or(usize::MAX);
+        SwappyAllocator::new(&Path::new(&cfg.dir).join("swap"), memory_limit)
+    }
+}
+
+#[cfg(not(feature = "swap"))]
+mod swap_conditional_imports {
+    pub trait AllocatorTrait {}
+    impl AllocatorTrait for () {}
+
+    pub type SelectedAllocator = ();
+    pub type VacantAllocator = ();
+
+    pub fn new_vacant_allocator() -> VacantAllocator {}
+    pub fn new_allocator(_: &crate::Config) -> SelectedAllocator {}
+}
+
+use swap_conditional_imports::*;
 
 /// Attempt to shrink entry container if its capacity reaches the threshold.
 const CAPACITY_SHRINK_THRESHOLD: usize = 1024 - 1;
@@ -88,12 +126,15 @@ impl From<&EntryIndex> for ThinEntryIndex {
 ///
 /// Each Raft Group has its own `MemTable` to store all key value pairs and the
 /// file locations of all log entries.
-pub struct MemTable {
+pub struct MemTable<A: AllocatorTrait> {
     /// The ID of current Raft Group.
     region_id: u64,
 
     /// Container of entries. Incoming entries are pushed to the back with
     /// ascending log indexes.
+    #[cfg(feature = "swap")]
+    entry_indexes: VecDeque<ThinEntryIndex, A>,
+    #[cfg(not(feature = "swap"))]
     entry_indexes: VecDeque<ThinEntryIndex>,
     /// The log index of the first entry.
     first_index: u64,
@@ -106,17 +147,34 @@ pub struct MemTable {
 
     /// Shared statistics.
     global_stats: Arc<GlobalStats>,
+
+    _phantom: PhantomData<A>,
 }
 
-impl MemTable {
-    pub fn new(region_id: u64, global_stats: Arc<GlobalStats>) -> MemTable {
+impl MemTable<VacantAllocator> {
+    #[allow(dead_code)]
+    fn new(region_id: u64, global_stats: Arc<GlobalStats>) -> MemTable<VacantAllocator> {
+        Self::with_allocator(region_id, global_stats, &new_vacant_allocator())
+    }
+}
+
+impl<A: AllocatorTrait> MemTable<A> {
+    fn with_allocator(
+        region_id: u64,
+        global_stats: Arc<GlobalStats>,
+        _allocator: &A,
+    ) -> MemTable<A> {
         MemTable {
             region_id,
+            #[cfg(feature = "swap")]
+            entry_indexes: VecDeque::with_capacity_in(CAPACITY_INIT, _allocator.clone()),
+            #[cfg(not(feature = "swap"))]
             entry_indexes: VecDeque::with_capacity(CAPACITY_INIT),
             first_index: 0,
             rewrite_count: 0,
             kvs: BTreeMap::default(),
             global_stats,
+            _phantom: PhantomData,
         }
     }
 
@@ -506,10 +564,9 @@ impl MemTable {
         let start_pos = (begin - first) as usize;
         let end_pos = (end - begin) as usize + start_pos;
 
-        let (first, second) = slices_in_range(&self.entry_indexes, start_pos, end_pos);
         let mut total_size = 0;
         let mut index = begin;
-        for idx in first.iter().chain(second) {
+        for idx in self.entry_indexes.range(start_pos..end_pos) {
             total_size += idx.entry_len;
             // No matter max_size's value, fetch one entry at least.
             if let Some(max_size) = max_size {
@@ -670,7 +727,7 @@ impl MemTable {
     }
 }
 
-impl Drop for MemTable {
+impl<A: AllocatorTrait> Drop for MemTable<A> {
     fn drop(&mut self) {
         let mut append_kvs = 0;
         let mut rewrite_kvs = 0;
@@ -690,7 +747,8 @@ impl Drop for MemTable {
     }
 }
 
-type MemTables = HashMap<u64, Arc<RwLock<MemTable>>>;
+type MemTables = HashMap<u64, Arc<RwLock<MemTable<SelectedAllocator>>>>;
+pub type MemTableHandle = Arc<RwLock<MemTable<SelectedAllocator>>>;
 
 /// A collection of [`MemTable`]s.
 ///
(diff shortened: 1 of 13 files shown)


tabokie:
@Connor1996 @BusyJay PTAL

src/swappy_allocator.rs
```diff
@@ -0,0 +1,442 @@
+// Copyright (c) 2017-present, PingCAP, Inc. Licensed under Apache-2.0.
+
+//! # An [`Allocator`] implementation that has a memory budget and can be
+//! swapped out.
+
+use std::alloc::{AllocError, Allocator, Global, Layout};
```
BusyJay, line 6: APIs of Allocator have been changed multiple times, this can be hard to maintain until they are stable. Why not fix the problem out side of allocator?

tabokie: I can't think of a better solution without hacking into `VecDeque`, which is quite complex in itself. The allocator, on the other hand, is easy to code and test.

BusyJay: I don't get it. For example, set a capacity limit on `VecDeque` and put half of it to disk whenever it's full.

tabokie: I can see it has several drawbacks:
- setting hard limit to individual vecs is not equivalent to a global memory limit.
- maintain a list of items by ourselves that behaves like a `VecDeque` is complex
- involves more overhead than allocator. Because we are adding checks to every container access (in contrary to only adding checks during allocation).

BusyJay: > setting hard limit to individual vecs is not equivalent to a global memory limit.

This can also be implemented on a global level limit just like entry cache has done.

> maintain a list of items by ourselves that behaves like a VecDeque is complex

I don't know what items you refer to. To me, they are just backlog pages similar to log entries.

> involves more overhead than allocator. Because we are adding checks to every container access (in contrary to only adding checks during allocation).

The check needs to be examined only when adding new entries. And the cost is smaller than memory copy.

The advantages of doing it without allocator are:
1. It's more flexible and the algorithm can be adjusted to suit best for different cases.
2. We don't rely on unstable APIs, upgrading rustc won't accidentally breaks build.
3. It's clear what has been put into disk, and what's remained in memory, and it should be easier to work with async fetch entries.
4. It's simpler and safer as we don't need to take care about the allocation and deallocation, which is error prone.

BusyJay: > They both need to provide VecDeque interfaces

The indexes stored in disk are just like log entries except the content are indexes, they don't need `VecDeque` interface.

> Also, what do you mean "than memory copy"?

What I mean is the check is so simple that its cost is negligible given there are so many copies during writes.

> ...maybe, but that's empty talk until we really need those cases...

Just like what has happen to the memory usage, right?

> I already have working, simple code lying in this PR.

I would argue if it's really as simple as it seems. It requires knowledge of `Allocator` trait, memmap, and ordering to reasoning if it's really correct and safe.

There is another benefit of the doing it out side of allocator, that it should have better and stable performance. As it can only put the old entries to the swap, which is unlikely to be accessed. While using allocator, it behaves the opposite way, that actively access memory is mapped while inactive segment is always kept in memory.

tabokie: First let me say, if needed, a swappable allocator can be used to implement your solution more easily by reusing standard containers.

> `Allocator` trait

Yes. But I would argue the cost of adjusting to new interface is low.

> ordering to reasoning if it's really correct and safe

Maybe you haven't read the code. But the ordering is actually quite simple. All atomics access is relaxed order. Because we use mutex when manipulating pages.

> There is another benefit of the doing it out side of allocator, that it should have better and stable performance.

It's true. We will have to see the performance impact of that. Maybe it will not be a big issue considering the disk write is buffered.

> that actively access memory is mapped while inactive segment is always kept in memory.

I'm not sure. Do you mean inactive segment of the same region? But all entries of the region will be mapped on reallocation.

src/util.rs
```diff
@@ -210,29 +209,28 @@ impl InstantExt for Instant {
...
-
 #[inline]
 pub fn crc32(data: &[u8]) -> u32 {
     let mut hasher = Hasher::new();
     hasher.update(data);
     hasher.finalize()
 }
 
+#[inline]
+pub fn hash_u64(mut i: u64) -> u64 {
```
Connor1996, line 220: please add a comment about the alogorithm

src/memtable.rs
```diff
@@ -100,17 +150,34 @@ pub struct MemTable {
...
+
+impl<A: AllocatorTrait> MemTable<A> {
+    fn with_allocator(
+        region_id: u64,
+        global_stats: Arc<GlobalStats>,
+        _allocator: &A,
+    ) -> MemTable<A> {
         MemTable {
             region_id,
+            #[cfg(feature = "swap")]
```
Connor1996, line 172: It's really necessary to use feature?

tabokie: `with_capacity_in` is a nightly interface. We already took some efforts to make raft-engine a stable crate, so `swap` feature is needed to gate these nightly-only stuff.

src/swappy_allocator.rs
tabokie: More screenshots are attached. Only read IOs are growing during memory pressure, and it's very likely caused by rewrite process reading out old entries.

BusyJay: Why write duration breakdown shows that apply time arise after memory limit is reached?

src/memtable.rs
```diff
@@ -675,52 +742,96 @@ impl Drop for MemTable {
...
 /// hashed region IDs.
 #[derive(Clone)]
-pub struct MemTableAccessor {
+pub struct MemTableAccessor<A: AllocatorTrait> {
     global_stats: Arc<GlobalStats>,
+    allocator: A,
 
     /// A fixed-size array of maps of [`MemTable`]s.
-    slots: Vec<Arc<RwLock<MemTables>>>,
+    slots: Vec<Arc<RwLock<MemTableMap<A>>>>,
```
Connor1996, line 759: add cache paddding?

src/memtable.rs
```diff
@@ -675,52 +742,96 @@ impl Drop for MemTable {
...
+        });
         memtable.clone()
     }
 
-    pub fn get(&self, raft_group_id: u64) -> Option<Arc<RwLock<MemTable>>> {
+    pub fn get(&self, raft_group_id: u64) -> Option<Arc<RwLock<MemTable<A>>>> {
         self.slots[Self::slot_index(raft_group_id)]
             .read()
             .get(&raft_group_id)
             .cloned()
```
Connor1996, line 831: can we return reference?

src/swappy_allocator.rs
tabokie: Updates to entry indexes will land on page cache. As mmap doesn't involves syscalls on fast path, I believe the amortized cost is optimal.

src/memtable.rs
tabokie: Afraid not. The reference requires the lock on `slots`.

src/swappy_allocator.rs
BusyJay: How about Max Value? I can see both write and apply matches in the end. If max value also matches, it may cause higher jitters.

tabokie: Write duration 999% shows the tail latency is relatively stable.

BusyJay: I have checked the monitor, apply can increase as large as writes in tail latency.

BusyJay:
Security is not considered in this PR. All writes to disk should be encrypted if encryption is enabled.

tabokie:
> Security is not considered in this PR. All writes to disk should be encrypted if encryption is enabled.

I have thought about that. And I think it's not a problem because entry index only contains file offsets but no user data.

src/swappy_allocator.rs
```diff
@@ -0,0 +1,1195 @@
...
+        SwappyAllocator(Arc::new(core))
+    }
+
+    #[inline]
+    pub fn memory_usage(&self) -> usize {
+        self.0.mem_usage.load(Ordering::Relaxed)
+    }
+
+    #[inline]
+    fn is_swapped(&self, ptr: NonNull<u8>, exaustive_check: bool) -> bool {
```
Connor1996, line 72: exhaustive?

src/swappy_allocator.rs
```diff
@@ -0,0 +1,1195 @@
...
+
+    #[inline]
+    fn allocate(&mut self, layout: Layout) -> Option<NonNull<[u8]>> {
+        unsafe {
+            let offset = self
+                .mmap
+                .as_ptr()
+                .add(self.tail)
+                .align_offset(layout.align());
+            if self.tail + offset + layout.size() > self.mmap.len() {
```
Connor1996, line 322: should be `offset + layout.size() > self.mmap.len()`?

src/swappy_allocator.rs
```diff
@@ -0,0 +1,1195 @@
...
+                        page.release(&self.0.path);
+                    }
+                    if pages.is_empty() {
+                        self.0.maybe_swapped.store(false, Ordering::Relaxed);
+                    }
+                    return;
+                }
+            }
+        }
+        self.0.mem_usage.fetch_sub(layout.size(), Ordering::Relaxed);
```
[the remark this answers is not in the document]
tabokie: address space should be equal to `usize::MAX`, so I think overflow is impossible.

src/swappy_allocator.rs
```diff
@@ -0,0 +1,1195 @@
...
+            );
+        }
+        debug_assert!(self.0.maybe_swapped.load(Ordering::Relaxed));
+        match pages.last_mut().unwrap().allocate(layout) {
+            None => {
+                pages.push(
+                    Page::new(
+                        &self.0.path,
+                        self.0.page_seq.fetch_add(1, Ordering::Relaxed),
+                        std::cmp::max(DEFAULT_PAGE_SIZE, layout.size()),
```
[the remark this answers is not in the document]
tabokie: A larger page will be created. And you may worry that the page will still not be large enough because of alignment padding. But I checked that the address returned from mmap will have the maximum alignment, i.e. it can always fulfil the request.

src/swappy_allocator.rs
```diff
@@ -0,0 +1,1195 @@
...
+        Self::new_over(path, budget, Global)
+    }
+}
+
+unsafe impl<A: Allocator> Allocator for SwappyAllocator<A> {
+    #[inline]
+    fn allocate(&self, layout: Layout) -> Result<NonNull<[u8]>, AllocError> {
+        // Always use mem_allocator to allocate empty pointer.
+        if layout.size() > 0
+            && self.0.mem_usage.fetch_add(layout.size(), Ordering::Relaxed) + layout.size()
```
Connor1996, line 124: if there is a alloc error, it won't be correct

src/swappy_allocator.rs
```diff
@@ -0,0 +1,1195 @@
+// Copyright (c) 2017-present, PingCAP, Inc. Licensed under Apache-2.0.
```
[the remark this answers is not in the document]
tabokie: I don't think it's right. Copyright should extend to the current date.

src/swappy_allocator.rs
```diff
@@ -0,0 +1,1195 @@
...
+            let offset = self
+                .mmap
+                .as_ptr()
+                .add(self.tail)
+                .align_offset(layout.align());
+            if self.tail + offset + layout.size() > self.mmap.len() {
+                None
+            } else {
+                let p = self.mmap.as_ptr().add(self.tail + offset);
+                self.tail += offset + layout.size();
```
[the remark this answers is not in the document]
tabokie: No, the offset is not "offset within the page", it's really the alignment padding.

src/swappy_allocator.rs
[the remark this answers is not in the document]
Connor1996: Oh, misread

src/swappy_allocator.rs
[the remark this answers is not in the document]
Connor1996: I'm just afraid of bug that somewhere forget or fail(due to race condition considering memory order) to add the `mem_usage`, then it will underflow

src/swappy_allocator.rs
```diff
@@ -0,0 +1,1195 @@
...
+    #[inline]
+    pub fn memory_usage(&self) -> usize {
+        self.0.mem_usage.load(Ordering::Relaxed)
+    }
+
+    #[inline]
+    fn is_swapped(&self, ptr: NonNull<u8>, exaustive_check: bool) -> bool {
+        // Ordering: `maybe_swapped` must be read after the pointer is available.
+        std::sync::atomic::fence(Ordering::Acquire);
+        self.0.maybe_swapped.load(Ordering::Relaxed)
```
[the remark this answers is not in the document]
tabokie: It's different. fence is more powerful than one atomic access: An acquire atomic load prevents subsequent R/W from being reordered before this load. An acquire fence, on the other hand, prevents subsequent R/W from being reordered before any load before this fence (in essence, it makes all loads before this fence appear as one big atomic load).

Here we use a fence to prevent `maybe_swapped.load()` being ordered before `ptr.load()` (which happens before the scope of this function).

p.s. this is a good writeup: https://preshing.com/20130922/acquire-and-release-fences/

src/swappy_allocator.rs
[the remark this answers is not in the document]
Connor1996: I see

src/swappy_allocator.rs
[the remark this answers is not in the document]
Connor1996: always struggle with memory order🥲

src/swappy_allocator.rs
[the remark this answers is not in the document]
tabokie: Assume there is a bug, the damage is not that much (only some performance overhead). But if we use wrapping sub here, it might be difficult to figure out what went wrong. So I think we should let it fail loudly.

BusyJay:
I built a [POC](https://github.com/tikv/raft-engine/compare/master...BusyJay:side-load-index?expand=1) of offloading partial entries instead of whole to disk. Compared to this PR, it has the following advantages:
1. Page management is very simple, it gets rid of all the complexity brought by allocator;
2. It's expected to offload applied entries only, so it can optimize under the assumption that offloaded entries will not be changed (rewritten). Though not all optimizations are implemented yet.
3. Accessing active entries like uncommitted entries or unapplied entries won't trigger disk access, so it won't bring extra IO in the raftstore thread in most cases, even there is a large log lag.
4. It takes advantage of Posix file system that a process can still write to file after it's deleted, so it's slightly safer.

It has one disadvantage though: it can't give a precise control on memory. However, because it offloads all applied entry indexes to disk, so there should be a limited index in memory. It can also choose to offload all committed entry indexes, which will only need to keep at most 100MiB in memory.

tabokie:
> I built a [POC](https://github.com/tikv/raft-engine/compare/master...BusyJay:side-load-index?expand=1) of offloading partial entries instead of whole to disk. Compared to this PR, it has the following advantages:

There are three weeks before code frozen. I don't trust the quality of your bloated data structure can be well tested. My allocator, due to its inherent simplicity (both in design and in interaction with other components), is already well tested (1000 LOC of tests + all existing tests).

I agree some of your points in general, but:
1. Your code is **not** simple. I think anyone can see that.
2. The assumption is wrong. As I pointed out in previous comment, most read IOs are caused by rewrite process, and hit primarily cold entries (applied entries). And your code will be even more complex to support that.

The points that I do agree, are also not that important. Performance (especially read) under memory throttling has never been a priority. Yes, better performance is good to have, only if I'm comfortable with the cost. But looking at your code, I'm not comfortable.

BusyJay:
>     1. Your code is **not** simple. I think anyone can see that.

I don't see why the code is not simple. The most complexity comes from making `BacklogIndex` and in memory index behave as a unified interface. And these part can be well tested by unit tests.

> As I pointed out in previous comment, most read IOs are caused by rewrite process, and hit primarily cold entries (applied entries). And your code will be even more complex to support that.

So why applied entries can be rewritten? Applied means it's immutable, any rewrite is wrong.

> But looking at your code, I'm not comfortable.

The same feeling I have about allocator. I strongly object to use an allocator.

tabokie:
> I don't see why the code is not simple.

You said it yourself. "behave as a unified interface", which echoes my comment emphasizing that implementing our own `VecDeque` is complex. It's not the interface that is complex of course, it's the internal states and branches that are complex, and this complexity cannot be hidden. Last time I spent half month of intensive working to reason and cover nearly every conditional branch in this codebase.

Allocator is simple, because it doesn't tamper with any of the data structure and their interactions. Its methods adhere to a strict calling protocol. The combination of method calling is finite.

> So why applied entries can be rewritten? Applied means it's immutable, any rewrite is wrong.

Rewrite logic hasn't changed much since I took over this repo. And I suspect any other implementations can do away with compactions.

> The same feeling I have about allocator. I strongly object to use an allocator.

I guess we can't persuade anyone with more discussions. You can put it to a vote if necessary.

BusyJay:
> Rewrite logic hasn't changed much since I took over this repo. And I suspect any other implementations can do away with compactions.

Now I see your point. Instead of introducing backlog index, index may need to be tracked base on file instead of region. That is, region maintains a list of log files that contain its entry, and every log file has a region indexes queue. So that rewrite is just updating log file list. Though it's certainly a big refactor.

tabokie:
> Now I see your point. Instead of introducing backlog index, index may need to be tracked base on file instead of region. That is, region maintains a list of log files that contain its entry, and every log file has a region indexes queue. So that rewrite is just updating log file list. Though it's certainly a big refactor.

So a more LSM-like approach. With a proper block cache it might work better than the current implementation. And it will cut a lot of complexity from memtable to make it easier to maintain. I'll file an issue later.

src/swappy_allocator.rs
```diff
@@ -0,0 +1,1267 @@
...
+        let mut pages = self.0.pages.lock();
+        if pages.is_empty() {
+            self.0.maybe_swapped.store(true, Ordering::Relaxed);
+            // Ordering: `maybe_swapped` must be set before the page is created.
+            std::sync::atomic::fence(Ordering::Release);
+            pages.push(
+                Page::new(
+                    &self.0.path,
+                    self.0.page_seq.fetch_add(1, Ordering::Relaxed),
+                    std::cmp::max(DEFAULT_PAGE_SIZE, layout.size()),
```
Connor1996, line 90: add a metrics to show how many memory are swapped

src/swappy_allocator.rs
```diff
@@ -0,0 +1,1267 @@
...
+                )
+                .ok_or(AllocError)?,
+            );
+        }
+        debug_assert!(self.0.maybe_swapped.load(Ordering::Relaxed));
+        match pages.last_mut().unwrap().allocate(layout) {
```
Connor1996, line 96: `pages.last_mut().map(|p| p.allocate(layout))`, then you can reuse pages.push() below

Connor1996 (approved):
LGTM

PR #213, opened 2022-04-13
offload in-memory `entry_len` field to disk

tabokie:

Ref #205 

Previously, we store the (offset, length) of each entry in memory. In this PR, the length field is encoded in on-disk data to reduce memory usage.

The size of `ThinEntryIndex` is now 40B compared to 64B in v0.1.0.


BusyJay:
Why this PR is merged without review?

PR #219, opened 2022-05-07
more robust corruption detection

tabokie:

By default, tail corruption, i.e. the corruption of end record in the last log file, is tolerated. Log header is treated as a special log record. This PR improves the corruption detection by making sure log header is sync-ed before the next log record is written.

Added tests for the following cases:
- `[corrupted header]`: Pass
- `[corrupted header][log records]`: Fail
- `[header][one corrupted record]`: Pass


Connor1996 (approved):
LGTM

Issue #220, opened 2022-05-18
Log batch decoding and encoding should consider format version


LykxSassinator:
As the title shown, the VERSION of log files does not make sense currently. Pratically, different VERSIONs should make difference on the behaviors of reading and writing on log files.
For the further development, we want to do the relevant implementation on it to make the VERSION significant.

PR #221, opened 2022-05-20
Make VERSION significant(#220)

LykxSassinator:
As the title shown, this commit is used for tackling the issue(#220).
This implementation is used for building the basic structure of binding the LogBatch `decoding` and `encoding` with `VERSION`.
Briefly, it introduces the trait `Header` for `LogFileHeader`, and make relevant implementations in `LogFileWriter` and `LogFileReader`.


src/file_pipe_log/pipe_builder.rs
```diff
@@ -227,7 +227,10 @@ impl<F: FileSystem> DualPipesBuilder<F> {
                     let is_last_file = index == chunk_count - 1 && i == file_count - 1;
```
tabokie, line 227: You can extend the file list to include a version field. Try use [par_chunks_mut](https://docs.rs/rayon/1.0.3/rayon/slice/trait.ParallelSliceMut.html#method.par_chunks_mut), decode the version here and update the file list.

src/file_pipe_log/pipe_builder.rs
```diff
@@ -315,7 +318,18 @@ impl<F: FileSystem> DualPipesBuilder<F> {
             LogQueue::Rewrite => &self.rewrite_files,
         };
         let first_seq = files.first().map(|f| f.seq).unwrap_or(0);
-        let files: VecDeque<Arc<F::Handle>> = files.iter().map(|f| f.handle.clone()).collect();
+        let files: VecDeque<FileHandler<F>> = files
+            .iter()
+            .map(|f| {
+                // As the version of each log file could not be captured from outsize, we
+                // need to do the `[build_file_header]` for parsing the `[Version]` at here.
+                let file_header = build_file_header(self.file_system.as_ref(), f.handle.clone());
```
tabokie, line 326: Refer to previous comment, no need to decode again.

src/engine.rs
```diff
@@ -149,8 +149,14 @@ where
                             .as_secs_f64(),
                     );
                     sync |= writer.sync;
-                    let log_batch = writer.get_payload();
+                    let log_batch = writer.get_mut_payload();
                     let res = if !log_batch.is_empty() {
+                        let (file_version, file_id) =
```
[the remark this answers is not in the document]
LykxSassinator: Got it.

src/config.rs
```diff
@@ -53,6 +53,12 @@ pub struct Config {
     ///
     /// Default: "4MB"
     pub bytes_per_sync: ReadableSize,
+
+    /// Version of the log file.
+    ///
+    /// Default: "V1"
+    pub file_version: String,
```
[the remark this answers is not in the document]
LykxSassinator: Surely, `String` might introduce unexpected `Version`.

src/engine.rs
```diff
@@ -458,12 +465,17 @@ where
 {
     BLOCK_CACHE.with(|cache| {
         if cache.key.get() != idx.entries.unwrap() {
+            let version = match idx.entries.as_ref() {
+                Some(handle) => pipe_log.fetch_file_version(handle.id),
+                None => Ok(Version::V1),
+            };
```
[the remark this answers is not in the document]
LykxSassinator: Ok, it was resolved.

src/file_pipe_log/format.rs
```diff
@@ -142,6 +184,15 @@ impl LogFileHeader {
     }
 }
 
+/// Basic trait for the successive LogXXX.
+pub trait Header {
```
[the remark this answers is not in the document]
LykxSassinator: In my perspective, the `LogFileWriter` and `LogFileReader` are accessors to LogFile for writing and reading. And each of them holds a `header` to represent the metadata for the related log file header. So, the `header` is proposed to provide the basic api for accessing and updating the `header` of `LogFileXXX`.

src/file_pipe_log/log_file.rs
```diff
@@ -110,25 +189,57 @@ impl<F: FileSystem> LogFileWriter<F> {
...
+///
+/// Attention please, the reader do not need a specified `[LogFileHeader]` from
+/// users.
+///
+/// * `[handle]`: standard handle of a log file.
+/// * `[reload_header]`: flag, whether to reload the log file header or not.
 pub(super) fn build_file_reader<F: FileSystem>(
     system: &F,
     handle: Arc<F::Handle>,
+    reload_header: bool,
```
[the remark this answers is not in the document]
LykxSassinator: The `reload_header` flag is used to specifiy whether it should load the header again if the caller wanted. By default, I expect that the `LogFileReader` should automatically load the header when it is built. However, there might exist other circumstances that it do not need to load the header, typically in the processing of build the `LogItemBatchFileReader`. So, this func is defined as this shown.

src/file_pipe_log/format.rs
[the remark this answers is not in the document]
tabokie: I think you can directly access the header field instead of using a method? Abstraction is only needed if there are some other unknown types that might implement this trait.

src/file_pipe_log/pipe_builder.rs
LykxSassinator: Yes! It can't be agreed any more. Thanks, i'll take a try by introducing `par_chunks_mut` to modify the ugly implementation.
Firstly, when i tried to extend it, i found an embarrassing fact that the `file_list` was built by the `scan`, whose main purpose was to build the list of log files without loading `file header` in advance. Then, the validation of `header` was done during the processing of `recover`, which just got a `read only` `file_list`. That is, i could not update the `version` of each file here. So, the final is implemented as the below shown.

src/file_pipe_log/log_file.rs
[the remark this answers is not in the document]
tabokie: Pass in a `Option<Version>` then, I'm thinking sth like this:
```
fn build_file_reader(fs, handle, mut version: Option<Version>) {
  let reader = fs.new_reader(handle);
  if version.is_none() {
    version = Some(Version::decode(reader.read(..))?);
  }
  LogFileReader::open(reader, version.unwrap())
}
```

src/file_pipe_log/format.rs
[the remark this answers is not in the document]
LykxSassinator: Yep, i just wanna import the `Header` as a common trait for `LogFileReader` and `LogFileWriter`. But it seems that it's not an appropriate design.
Your suggestion is a better choice for this.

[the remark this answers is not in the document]
LykxSassinator: Resolved

src/config.rs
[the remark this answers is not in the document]
LykxSassinator: Resolved.

src/engine.rs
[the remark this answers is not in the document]
LykxSassinator: Resolved.

src/file_pipe_log/log_file.rs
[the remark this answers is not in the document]
LykxSassinator: Cool, `Option<Header>` seems more resonable than `bool`.
Resolved.

src/file_pipe_log/log_file.rs
```diff
@@ -9,32 +9,110 @@ use crate::metrics::*;
...
+    pub handle: Arc<F::Handle>,
+    pub version: Version,
+}
+
+/// Function for reading the header of the log file, and return a
+/// `[LogFileHeader]`.
+///
+/// Attention please, to avoid to move the offset of the given `[handle]`, we
+/// use a copy of the `[handle]` to parse the header of the given file.
+pub(super) fn build_file_header<F: FileSystem>(
```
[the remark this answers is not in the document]
LykxSassinator: Done.

src/file_pipe_log/pipe_builder.rs
LykxSassinator: Done

src/log_batch.rs
```diff
@@ -1115,9 +1138,13 @@ mod tests {
...
             entries_handle.offset = LOG_BATCH_HEADER_LEN as u64;
             entries_handle.len = offset - LOG_BATCH_HEADER_LEN;
-            let decoded_item_batch =
-                LogItemBatch::decode(&mut &encoded[offset..], entries_handle, compression_type)
-                    .unwrap();
+            let decoded_item_batch = LogItemBatch::decode(
+                &mut &encoded[offset..],
+                entries_handle,
+                compression_type,
+                Version::V1,
```
tabokie, line 1145: Replace all `Version::V1` with `Version::default()`

src/file_pipe_log/log_file.rs
```diff
@@ -110,25 +183,44 @@ impl<F: FileSystem> LogFileWriter<F> {
...
 ) -> Result<LogFileReader<F>> {
     let reader = system.new_reader(handle.clone())?;
-    LogFileReader::open(handle, reader)
+    if let Some(version) = expected_version {
+        LogFileReader::open(handle, reader, version)
+    } else {
+        LogFileReader::open(
+            handle.clone(),
+            reader,
+            (build_file_header(system, handle)?).version(),
```
tabokie, line 206: `build_file_header` shouldn't recreate a fs reader. In most context the reader is ready. Here you should reuse the reader created by `system.new_reader(handle.clone())?;`.

src/file_pipe_log/pipe_builder.rs
```diff
@@ -172,62 +172,61 @@ impl<F: FileSystem> DualPipesBuilder<F> {
...
-        );
-
+        // As the `recover_queue` would update the `Version` of each log file in
+        // `apend_files` and `rewrite_files`, we just use the naive api `install`
+        // as the choice for parallelly recovering `LogQueue::Append` and
+        // `LogQueue::Rewrite`.
+        let append =
+            pool.install(|| self.recover_queue(LogQueue::Append, machine_factory, threads));
+        let rewrite =
+            pool.install(|| self.recover_queue(LogQueue::Rewrite, machine_factory, threads));
```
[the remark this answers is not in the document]
LykxSassinator: Sure, I'll supplement it.
On the one hand, an initial reason why i replaced `pool.join(_, _)` with seperated `install` is that `join` would cause `double-reference error` as the implementation of `recover_queue` introduces`&mut self`. 
On the other hand, each installed task would be excuted on the TheadPool by utilizing each sub-threads in it with the param `threads`.  And in my opinion, it seems also effective, compared to the original `join` with  param `rewrite_accuracy` and `append_accuracy` respectively ?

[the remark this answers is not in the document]
tabokie: Yes, I'm not saying the performance would be worse. It's just different.

src/config.rs
```diff
@@ -53,6 +53,12 @@ pub struct Config {
     ///
     /// Default: "4MB"
     pub bytes_per_sync: ReadableSize,
+
+    /// Version of the log file.
+    ///
+    /// Default: "1" => "Version::V1"
```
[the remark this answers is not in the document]
LykxSassinator: Done

src/config.rs
```diff
@@ -53,6 +53,12 @@ pub struct Config {
...
+    pub file_version: u64,
```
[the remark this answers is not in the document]
LykxSassinator: Done

src/engine.rs
```diff
@@ -149,8 +149,9 @@ where
                             .as_secs_f64(),
                     );
                     sync |= writer.sync;
-                    let log_batch = writer.get_payload();
+                    let log_batch = writer.get_mut_payload();
```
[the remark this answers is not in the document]
LykxSassinator: Done

src/file_pipe_log/format.rs
```diff
@@ -80,17 +80,64 @@ pub(super) fn lock_file_path<P: AsRef<Path>>(dir: P) -> PathBuf {
...
+            _ => Version::V1, // default
+        }
+    }
+
+    #[allow(dead_code)]
+    pub fn to_str(version: Version) -> Option<String> {
+        match version {
+            Version::V1 => Some(String::from("V1")),
+        }
+    }
```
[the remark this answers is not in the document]
LykxSassinator: Done

src/purge.rs
```diff
@@ -351,6 +351,12 @@ where
         if len == 0 {
             return self.pipe_log.maybe_sync(LogQueue::Rewrite, sync);
         }
+        let (file_version, file_id) = self.pipe_log.fetch_active_file(LogQueue::Rewrite);
+        // Signs a checksum, so-called `[signature]`, into the LogBatch. The
+        // `[signature]` is both generated by the given `[Version]` and
+        // `[FileId]`. That is, the final checksum of each LogBatch consists of
+        // this signature and the checksum of the contents.
+        log_batch.sign_checksum(file_version, file_id);
```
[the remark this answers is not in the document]
LykxSassinator: Done

src/filter.rs
```diff
@@ -287,6 +287,8 @@ impl RhaiFilterMachine {
                                     &entries_buf,
                                     ei.entries.unwrap(),
                                     ei.compression_type,
+                                    reader.header.version(),
+                                    // reader.get_file_header().version(),
```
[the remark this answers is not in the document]
LykxSassinator: Done

src/file_pipe_log/pipe_builder.rs
```diff
@@ -172,62 +172,61 @@ impl<F: FileSystem> DualPipesBuilder<F> {
...
             .unwrap();
-        let (append, rewrite) = pool.join(
-            || self.recover_queue(LogQueue::Append, machine_factory, append_concurrency),
-            || self.recover_queue(LogQueue::Rewrite, machine_factory, rewrite_concurrency),
-        );
-
+        // As the `recover_queue` would update the `Version` of each log file in
+        // `apend_files` and `rewrite_files`, we just use the naive api `install`
+        // as the choice for parallelly recovering `LogQueue::Append` and
+        // `LogQueue::Rewrite`.
```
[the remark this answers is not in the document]
LykxSassinator: Done

src/file_pipe_log/pipe_builder.rs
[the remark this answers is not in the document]
LykxSassinator: There still exists bugs when introducing `Criterion` crate for generating the performance report on MacOS.
I'll try to generate the comparison result in a later commit.

src/file_pipe_log/pipe_builder.rs
```diff
@@ -156,7 +151,12 @@ impl<F: FileSystem> DualPipesBuilder<F> {
                         files.clear();
                     } else {
                         let handle = Arc::new(self.file_system.open(&path)?);
-                        files.push(FileToRecover { seq, handle });
+                        files.push(FileHandler {
+                            seq,
+                            handle,
+                            version: Version::V1, /* default with V1, would be updated later in
```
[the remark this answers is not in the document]
LykxSassinator: Done.

src/log_batch.rs
```diff
@@ -762,6 +766,14 @@ impl LogBatch {
         }
     }
 
+    #[allow(unused_variables)]
+    pub(crate) fn sign_checksum(&mut self, file_version: Version, file_id: FileId) {
+        debug_assert!(matches!(self.buf_state, BufState::Sealed(_, _)));
+        // @TODO: lucasliang
+        // Supllement the details of signing checksum on each LogBatch with the
+        // specific `Version`.
+    }
```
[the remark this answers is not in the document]
LykxSassinator: Done.

src/log_batch.rs
LykxSassinator: Done

src/file_pipe_log/log_file.rs
LykxSassinator: Done.

src/file_pipe_log/pipe_builder.rs
```diff
@@ -181,53 +193,125 @@ impl<F: FileSystem> DualPipesBuilder<F> {
...
+    pub fn get_mut_file_list(&mut self, queue: LogQueue) -> &mut Vec<FileHandler<F>> {
+        if queue == LogQueue::Append {
+            &mut self.append_files
+        } else {
+            &mut self.rewrite_files
+        }
+    }
+}
+
+impl<F: FileSystem> DualPipesBuilder<F> {
```
tabokie, line 278: Is there a particular reason to separate this into a new block? It can be confusing at first glance.

src/file_pipe_log/pipe_builder.rs
```diff
@@ -181,53 +193,125 @@ impl<F: FileSystem> DualPipesBuilder<F> {
...
-                for (i, f) in chunk.iter().enumerate() {
+                for (i, f) in chunk.iter_mut().enumerate() {
                     let is_last_file = index == chunk_count - 1 && i == file_count - 1;
                     if let Err(e) = reader.open(
                         FileId { queue, seq: f.seq },
-                        build_file_reader(self.file_system.as_ref(), f.handle.clone())?,
+                        // Attention please, to make sure that the `[Err]` message coud be captured,
+                        // the reader is just a `[LogFileReader]` without loadin the file header
+                        // in advance, by passing `Version::V1` as the default param.
+                        build_file_reader(filesystem_ref, f.handle.clone(), Some(Version::default()))?,
```
tabokie, line 314: Let me clarify a bit about the use of `build_file_reader`. We only pass in a non-null `version` field into the reader if we already know the format version of the file. Other time use `None` and let reader do the parsing itself (not doing it in `LogItemBatchFileReader`).

src/file_pipe_log/pipe_builder.rs
```diff
@@ -156,7 +159,12 @@ impl<F: FileSystem> DualPipesBuilder<F> {
                         files.clear();
                     } else {
                         let handle = Arc::new(self.file_system.open(&path)?);
-                        files.push(FileToRecover { seq, handle });
+                        files.push(FileHandler {
+                            seq,
+                            handle,
+                            version: None, /* default with V1, would be updated later in
+                                            * the `recover_queue` progress */
```
tabokie, line 166:
```suggestion
                            version: None,
```

src/file_pipe_log/pipe_builder.rs
```diff
@@ -16,12 +16,12 @@ use crate::config::{Config, RecoveryMode};
...
 use crate::log_batch::LogItemBatch;
-use crate::pipe_log::{FileId, FileSeq, LogQueue};
+use crate::pipe_log::{FileId, LogQueue};
 use crate::util::Factory;
 use crate::{Error, Result};
 
-use super::format::{lock_file_path, FileNameExt, LogFileHeader};
-use super::log_file::build_file_reader;
+use super::format::{lock_file_path, FileNameExt, LogFileHeader, Version};
+use super::log_file::{build_file_reader, FileHandler};
```
tabokie, line 24: I don't like reusing it. The version field in `FileHandler` doesn't have to be `Option`. Make it optional is confusing.

src/file_pipe_log/log_file.rs
```diff
@@ -53,7 +132,8 @@ impl<F: FileSystem> LogFileWriter<F> {
         self.last_sync = 0;
         self.written = 0;
         let mut buf = Vec::with_capacity(LogFileHeader::len());
-        LogFileHeader::default().encode(&mut buf)?;
+        // LogFileHeader::default().encode(&mut buf)?;
```
tabokie, line 135: I feel like you can do a self-review next time to prune off these unneeded comments (it's easier to see them in GitHub PR view).

src/file_pipe_log/log_file.rs
```diff
@@ -110,25 +190,42 @@ impl<F: FileSystem> LogFileWriter<F> {
...
+/// Attention please, the reader do not need a specified `[LogFileHeader]` from
+/// users.
+///
+/// * `[handle]`: standard handle of a log file.
+/// * `[expected_version]`: if `[None]`, reloads the log file header and parse
+/// the relevant `Version` before building the `reader`.
 pub(super) fn build_file_reader<F: FileSystem>(
     system: &F,
     handle: Arc<F::Handle>,
+    expected_version: Option<Version>,
```
tabokie, line 204: I could mistake `expected_version` for doing verification etc.
```suggestion
    version: Option<Version>,
```

src/file_pipe_log/log_file.rs
```diff
@@ -6,35 +6,114 @@ use std::io::{Read, Seek, SeekFrom, Write};
...
+/// Build a file writer.
+///
+/// * `[handle]`: standard handle of a log file.
+/// * `[rewrite_flag]`:
+///     - `true` => rewrite the header with `[expected_hdr]`;
+///     - `false`=> use the original header from the log file.
 pub(super) fn build_file_writer<F: FileSystem>(
     system: &F,
     handle: Arc<F::Handle>,
+    expected_hdr: Option<LogFileHeader>,
```
tabokie, line 90: Two things here,
(1) use version like reader, in the future we might have more options in `LogFileHeader` that could affect file writing, but right now a version is good enough.
(2) Always provide a valid version to build file writer. It's because extracting version requires creating a file reader. It is sort-of a heavy operation, but we can't infer that information from the function name `build_file_writer`.
```suggestion
    version: Version,
```

src/file_pipe_log/log_file.rs
```diff
@@ -6,35 +6,114 @@ use std::io::{Read, Seek, SeekFrom, Write};
...
+/// Function for reading the header of the log file, and return a
+/// `[LogFileHeader]`.
+///
+/// Attention please, to avoid to move the offset of the given `[handle]`, we
+/// use a copy of the `[handle]` to parse the header of the given file.
+pub(super) fn build_file_header<F: FileSystem>(
+    system: &F,
+    handle: Arc<F::Handle>,
+    expected_reader: Option<&mut F::Reader>,
+) -> Result<LogFileHeader> {
```
tabokie, line 35: I think too many things are going on in this function. You can make this function a member of `LogFileReader`. That seems more natural to me.

src/file_pipe_log/log_file.rs
```diff
@@ -6,35 +6,114 @@ use std::io::{Read, Seek, SeekFrom, Write};
...
+    if file_size == 0 {
+        return Ok(LogFileHeader::default());
+    }
+    // [2] If the length lessed than the standard `LogFileHeader::len()`.
+    let header_len = LogFileHeader::len();
+    if file_size < header_len {
+        return Err(Error::Corruption("Invalid header of LogFile!".to_owned()));
+    }
+    // [3] Parse the header of the file.
+    let parse_file_header = |handle: Arc<F::Handle>, header_len: usize| -> Result<LogFileHeader> {
```
tabokie, line 52: BTW, why use a closure here?

src/file_pipe_log/format.rs
```diff
@@ -80,23 +80,44 @@ pub(super) fn lock_file_path<P: AsRef<Path>>(dir: P) -> PathBuf {
...
+        Self { version }
+    }
+
+    #[allow(dead_code)]
+    pub fn version(&self) -> Version {
+        self.version
+    }
 }
 
 impl LogFileHeader {
```
tabokie, line 123: You are breaking the implementations into two parts again.

src/file_pipe_log/pipe_builder.rs
LykxSassinator: It's just my code-sytle.
I just wanna divide `methods`(called by a specific `struct` object) and `associated functions`(called by the `struct` type) into seperated parts.

src/file_pipe_log/pipe_builder.rs
LykxSassinator: Yes, u'r right, it would make the implementation confusing.
The reason why it inroduce a default `Version` was intended to let the `LogItemBatchFileReader.open(_)` do the `build` its `header` as the previous codes do. And this implementation would follow the error capture of `if let Err(e)`.
It could be modified with a better way.

src/file_pipe_log/log_file.rs
LykxSassinator: Sure, it's a confusing and unnecessary annotation.

src/file_pipe_log/pipe_builder.rs
LykxSassinator: The reason why it defined with`Option` is that i wanna distinguish whether the `version` is initiated or not. At the begging, it should be `None` when the `SinglePipe` is not recovered.

src/file_pipe_log/log_file.rs
LykxSassinator: Oh, it's just a personal code-sytle. As the capture of variable are implemented as `moved`,  it actually made no difference to the direct implementation without the closure. I just wanna to integrate the implementation of `parsing the header` here.

src/file_pipe_log/log_file.rs
tabokie: You don't need to deal with file system when implementing member function for `LogFileReader`. It is a utility wrapper over the reader produced by file system.

src/file_pipe_log/log_file.rs
tabokie: You can define a function inside a function. Closure is only needed when we need to dynamically capture a variable.

src/file_pipe_log/log_file.rs
LykxSassinator: Got it, i did know what u meant.

src/engine.rs
```diff
@@ -387,16 +390,37 @@ where
...
         let factory = crate::filter::RhaiFilterMachineFactory::from_script(script);
         let mut machine = None;
         if queue.is_none() || queue.unwrap() == LogQueue::Append {
-            machine = Some(builder.recover_queue(LogQueue::Append, &factory, 1)?);
+            machine = Some(FilePipeLogBuilder::recover_queue(
+                RecoveryConfig {
+                    recovery_read_block_size,
+                    mode: recovery_mode,
+                    queue: LogQueue::Append,
+                    file_system: file_system.clone(),
```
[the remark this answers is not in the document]
LykxSassinator: Emm...Here, I'm concerned about a circumstance that there might need extra param in the `Config`, so i introduce the `RecoveryConfig` as the container to integrate all the related params.

src/file_pipe_log/log_file.rs
LykxSassinator: Got it.

src/engine.rs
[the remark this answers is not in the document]
tabokie: You can keep a separate a config. But leave file_system out of it. And preferably move concurrency into it.

src/log_batch.rs
```diff
@@ -479,7 +480,10 @@ impl LogItemBatch {
         buf: &mut SliceReader,
         entries: FileBlockHandle,
         compression_type: CompressionType,
+        file_version: Version,
     ) -> Result<LogItemBatch> {
+        //@TODO: decoding based on the version waited to be implemented
```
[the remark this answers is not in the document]
LykxSassinator: Done

src/file_pipe_log/reader.rs
```diff
@@ -44,12 +44,15 @@ impl<F: FileSystem> LogItemBatchFileReader<F> {
     pub fn open(&mut self, file_id: FileId, reader: LogFileReader<F>) -> Result<()> {
         self.file_id = Some(file_id);
         self.size = reader.file_size()?;
-        self.reader = Some(reader);
+        self.reader = Some(reader); // The LogFileReader is moved to this obj's reader.
```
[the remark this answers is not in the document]
LykxSassinator: Done

src/file_pipe_log/pipe_builder.rs
LykxSassinator: Clean-up the code-style.

src/file_pipe_log/pipe_builder.rs
LykxSassinator: Done

src/file_pipe_log/format.rs
LykxSassinator: Done

src/engine.rs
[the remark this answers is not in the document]
LykxSassinator: Got it.

src/pipe_log.rs
```diff
@@ -120,4 +121,13 @@ pub trait PipeLog: Sized {
...
     fn purge_to(&self, file_id: FileId) -> Result<usize>;
+
+    /// Returns the `[Version]` of the specific file.
+    fn fetch_file_version(&self, file_id: FileId) -> Result<Version>;
+
+    /// Returns the active `[FileId]` and related `[Version]` of the specific
+    /// log queue.
+    ///
+    /// Returns the `[Version]` and `[FileId]` of the related file.
+    fn fetch_active_file(&self, queue: LogQueue) -> (Version, FileId);
```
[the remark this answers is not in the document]
LykxSassinator: Done.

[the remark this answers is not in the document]
LykxSassinator: Ingore it.
I should follow the basics that `Each PR is used to tackle one issue.`
DONE.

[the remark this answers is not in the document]
tabokie: Oh, I didn't see `fetch_file_version` is already used. You can keep it, but remove `fetch_active_file`.

src/config.rs
```diff
@@ -132,6 +140,9 @@ impl Config {
             );
             self.recovery_threads = MIN_RECOVERY_THREADS;
         }
+        if !Version::is_valid(self.format_version) {
+            self.format_version = Version::default() as u64;
```
tabokie, line 144: Warning before changing the config value.

src/file_pipe_log/pipe_builder.rs
```diff
@@ -55,9 +55,12 @@ impl<M: ReplayMachine + Default> Factory<M> for DefaultMachineFactory<M> {
...
 
-struct FileToRecover<F: FileSystem> {
-    seq: FileSeq,
-    handle: Arc<F::Handle>,
+/// Container for basic settings on recovery.
+pub struct RecoveryConfig {
+    pub queue: LogQueue,
+    pub mode: RecoveryMode,
+    pub concurrency: usize,
+    pub recovery_read_block_size: u64,
```
tabokie, line 63:
```suggestion
    pub read_block_size: u64,
```

src/file_pipe_log/log_file.rs
tabokie: This comment is not addressed. I expect it to be changed to `LogFileReader::parse_header(&mut self)`.

src/file_pipe_log/pipe_builder.rs
tabokie: This one is not addressed. I believe `Pipe` is of higher priority here. It creates two issues: (1) adding seq to pipe's fileinfo creates multiple sources of truth; (2) using optional version doesn't convey the true meaning of the data (the version is already validated during recovery).

src/file_pipe_log/pipe.rs
```diff
@@ -35,6 +35,7 @@ struct ActiveFile<F: FileSystem> {
 pub(super) struct SinglePipe<F: FileSystem> {
     queue: LogQueue,
     dir: String,
+    target_file_header: LogFileHeader,
```
tabokie, line 38: Same as the old comment, use version for now.

src/file_pipe_log/pipe_builder.rs
```diff
@@ -87,6 +90,129 @@ impl<F: FileSystem> DualPipesBuilder<F> {
         }
     }
 
+    /// Manually reads through log items in all available log files of the
+    /// specified queue, and replays them to specific [`ReplayMachine`]s
+    /// that can be constructed via `machine_factory`.
+    pub fn recover_queue<M: ReplayMachine, FA: Factory<M>>(
```
tabokie, line 96: Could you keep the position unchanged? Previously it reflected the order of method calling (scan -> recover -> recover_queue).

src/file_pipe_log/log_file.rs
```diff
@@ -110,25 +126,51 @@ impl<F: FileSystem> LogFileWriter<F> {
     }
 }
 
+/// Build a file reader.
+///
+/// Attention please, the reader do not need a specified `[LogFileHeader]` from
+/// users.
+///
+/// * `[handle]`: standard handle of a log file.
+/// * `[expected_version]`: if `[None]`, reloads the log file header and parse
```
tabokie, line 135: Stale comment.

src/file_pipe_log/format.rs
```diff
@@ -140,6 +167,59 @@ impl LogFileHeader {
...
+    /// Function for reading the header of the log file, and return a
+    /// `[LogFileHeader]`.
+    ///
+    /// Attention please, to avoid to move the offset of the given `[handle]`,
+    /// we use a copy of the `[handle]` to parse the header of the given
+    /// file.
+    pub(super) fn build_file_header<F: FileSystem>(
+        system: &F,
+        handle: Arc<F::Handle>,
+        expected_reader: Option<&mut F::Reader>,
```
tabokie, line 180:
```suggestion
        reader: Option<&mut F::Reader>,
```

src/file_pipe_log/pipe.rs
```diff
@@ -63,8 +64,9 @@ impl<F: FileSystem> SinglePipe<F> {
         listeners: Vec<Arc<dyn EventListener>>,
         queue: LogQueue,
         mut first_seq: FileSeq,
-        mut fds: VecDeque<Arc<F::Handle>>,
+        mut fds: VecDeque<FileHandler<F>>,
     ) -> Result<Self> {
+        let expected_file_header = LogFileHeader::new(cfg.format_version);
```
tabokie, line 69:
```suggestion
        let default_version = cfg.format_version;
```

src/file_pipe_log/log_file.rs
LykxSassinator: Done

src/config.rs
LykxSassinator: Done.

src/file_pipe_log/pipe_builder.rs
LykxSassinator: Done

src/file_pipe_log/pipe_builder.rs
LykxSassinator: Done

src/file_pipe_log/log_file.rs
LykxSassinator: My fault, it's missed.

src/file_pipe_log/pipe.rs
LykxSassinator: Done

src/file_pipe_log/pipe_builder.rs
```diff
@@ -181,84 +198,120 @@ impl<F: FileSystem> DualPipesBuilder<F> {
...
-                            f.handle.truncate(0)?;
-                        } else {
-                            error!(
-                                "Failed to open log file due to broken header: {:?}:{}",
-                                queue, f.seq
-                            );
-                            return Err(e);
+                    // Attention please, to make sure that the `[Err]` message coud be captured,
+                    // the reader is just a `[LogFileReader]` without loadin the file header
+                    // in advance, by passing `Version::V1` as the default param.
```
tabokie, line 273: Stale comment.

src/file_pipe_log/pipe_builder.rs
```diff
@@ -181,84 +198,120 @@ impl<F: FileSystem> DualPipesBuilder<F> {
...
+                                return Err(e);
+                            }
+                        },
+                        Ok(file_reader) => {
+                            invalid_flag = false;
+                            reader.open(FileId { queue, seq: f.seq }, file_reader)?;
                         }
                     }
+                    // Update file version of each log file.
+                    f.version = Some( if !invalid_flag { reader.file_header().unwrap().version() } else { Version::default() });
```
tabokie, line 314: Some cleaning up:
```suggestion
                    match build_file_reader(file_system.as_ref(), f.handle.clone(), None) {
                        Err(e) => {
                            let is_local_tail = f.handle.file_size()? <= LogFileHeader::len();
                            if recovery_mode == RecoveryMode::TolerateAnyCorruption
                              || recovery_mode == RecoveryMode::TolerateTailCorruption
                                && is_last_file && is_local_tail {
                                warn!(
                                    "File header is corrupted but ignored: {:?}:{}, {}",
                                    queue, f.seq, e
                                );
                                f.handle.truncate(0)?;
                                f.version = Some(Version::default());
                                continue;
                            } else {
                                error!(
                                    "Failed to open log file due to broken header: {:?}:{}",
                                    queue, f.seq
                                );
                                return Err(e);
                            }
                        },
                        Ok(file_reader) => {
                            reader.open(FileId { queue, seq: f.seq }, file_reader)?;
                        }
                    }
                    f.version = Some(reader.file_header().unwrap().version());
```

src/file_pipe_log/format.rs
```diff
@@ -80,31 +80,56 @@ pub(super) fn lock_file_path<P: AsRef<Path>>(dir: P) -> PathBuf {
...
 
+/// In-memory representation of the log file header.
+#[derive(Clone, Default)]
+pub struct LogFileHeader {
+    version: Version,
+}
+
 impl LogFileHeader {
+    #[allow(dead_code)]
+    pub fn new(version: u64) -> Self {
```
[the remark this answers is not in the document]
LykxSassinator: Ok, done.

src/file_pipe_log/log_file.rs
```diff
@@ -6,35 +6,51 @@ use std::io::{Read, Seek, SeekFrom, Write};
...
+use super::format::{LogFileHeader, Version};
 use crate::env::{FileSystem, Handle, WriteExt};
 
 /// Maximum number of bytes to allocate ahead.
 const FILE_ALLOCATE_SIZE: usize = 2 * 1024 * 1024;
 
+/// Combination of `[Handle]` and `[Version]`, specifying a handler of a file.
+#[derive(Clone, Debug)]
+pub struct FileHandler<F: FileSystem> {
+    pub seq: FileSeq,
```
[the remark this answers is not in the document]
LykxSassinator: Done.

src/file_pipe_log/pipe.rs
```diff
@@ -184,7 +211,13 @@ impl<F: FileSystem> SinglePipe<F> {
 impl<F: FileSystem> SinglePipe<F> {
     fn read_bytes(&self, handle: FileBlockHandle) -> Result<Vec<u8>> {
         let fd = self.get_fd(handle.id.seq)?;
-        let mut reader = build_file_reader(self.file_system.as_ref(), fd)?;
+        // As the header of each log file already parsed in the processing of loading
+        // log files, we just need to build the `[LogFileReader]`.
```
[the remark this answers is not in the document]
LykxSassinator: Done

src/engine.rs
```diff
@@ -387,16 +390,37 @@ where
...
-            machine = Some(builder.recover_queue(LogQueue::Append, &factory, 1)?);
+            machine = Some(builder.recover_queue_on_type(
+                file_system.clone(),
+                RecoveryConfig {
+                    queue: LogQueue::Append,
+                    mode: recovery_mode,
+                    concurrency: 1,
+                    read_block_size,
+                },
+                LogQueue::Append,
```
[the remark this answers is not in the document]
LykxSassinator: My fault, it's truely duplicated on `LogQueue`.

src/file_pipe_log/log_file.rs
```diff
@@ -110,30 +126,64 @@ impl<F: FileSystem> LogFileWriter<F> {
...
-    LogFileReader::open(handle, reader)
+    if version.is_none() {
+        // Here, the caller expected that the given `handle` has pointed to
+        // a log file with valid format. Otherwise, it should return with
+        // `[Err]`.
+        if handle.file_size()? == 0 {
+            return Err(Error::Corruption(
+                "invalid format of file header".to_owned(),
+            ));
+        }
```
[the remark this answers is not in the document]
LykxSassinator: Done.

src/file_pipe_log/log_file.rs
```diff
@@ -6,35 +6,51 @@ use std::io::{Read, Seek, SeekFrom, Write};
...
 ) -> Result<LogFileWriter<F>> {
     let writer = system.new_writer(handle.clone())?;
-    LogFileWriter::open(handle, writer)
+    LogFileWriter::open(handle, writer, version)
 }
 
 /// Append-only writer for log file.
 pub struct LogFileWriter<F: FileSystem> {
+    /// header of file
+    pub header: LogFileHeader,
```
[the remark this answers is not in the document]
LykxSassinator: Both accpeted. I just wanna use `LogFileHeader` for furthur development. Keep it?

src/file_pipe_log/log_file.rs
```diff
@@ -164,6 +214,50 @@ impl<F: FileSystem> LogFileReader<F> {
...
+        let file_size: usize = match self.handle.file_size() {
+            Ok(size) => size,
+            Err(_) => {
+                return Err(Error::Corruption("Corrupted file!".to_owned())); // invalid file
+            }
+        };
+        // [1] If the file was a new file, we just return the default `LogFileHeader`.
+        if file_size == 0 {
+            return Ok(LogFileHeader::default());
+        }
```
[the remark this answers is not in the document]
LykxSassinator: Yep, it should be removed, which was a stale strategy for the judgement whether the caller had opened an empty file in the previou implementation.

src/file_pipe_log/pipe_builder.rs
```diff
@@ -326,6 +386,24 @@ impl<F: FileSystem> DualPipesBuilder<F> {
         )
     }
 
+    /// Manually reads through log items in all available log files of the
+    /// specified `[LogQueue]`, and replays them to specific
+    /// [`ReplayMachine`]s that can be constructed via `machine_factory`.
+    pub fn recover_queue_on_type<M: ReplayMachine, FA: Factory<M>>(
```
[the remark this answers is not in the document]
LykxSassinator: Done

src/file_pipe_log/log_file.rs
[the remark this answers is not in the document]
tabokie: I just think the naming is inappropriate: header is already written to file (which means we shouldn't copy it to memory), while version is a configuration that might apply to subsequent writes. So maybe rename `LogFileHeader` to `LogFileFormat` is better?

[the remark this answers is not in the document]
LykxSassinator: Yep, `LogFileHeader` also seems inconsistent with the configuration `format-version`.
 Cool！Introducing `LogFileFormat` is a better choice to keep the consistency with its raw definition to `LogFile`.

src/file_pipe_log/pipe_builder.rs
LykxSassinator: Done

src/file_pipe_log/mod.rs
```diff
@@ -46,7 +46,13 @@ pub mod debug {
             file_system.open(path)?
         };
         let fd = Arc::new(fd);
-        super::log_file::build_file_writer(file_system, fd)
+        let version = if create {
```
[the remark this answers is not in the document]
LykxSassinator: Done

src/file_pipe_log/log_file.rs
```diff
@@ -110,30 +125,62 @@ impl<F: FileSystem> LogFileWriter<F> {
...
+                handle,
+                reader,
+                // Set to an invalid offset to force a reseek at first read.
+                offset: u64::MAX,
+            }),
+            None => {
+                // Here, the caller expected that the given `handle` has pointed to
+                // a log file with valid format. Otherwise, it should return with
+                // `Err`.
+                if handle.file_size()? == 0 {
```
tabokie, line 168: This is already handled in `parse_format`.

src/file_pipe_log/log_file.rs
```diff
@@ -9,38 +9,53 @@ use crate::metrics::*;
...
 ) -> Result<LogFileWriter<F>> {
     let writer = system.new_writer(handle.clone())?;
-    LogFileWriter::open(handle, writer)
+    LogFileWriter::open(handle, writer, version)
 }
 
 /// Append-only writer for log file.
 pub struct LogFileWriter<F: FileSystem> {
+    /// header of file
+    pub header: LogFileFormat,
```
tabokie, line 41:
```suggestion
    format: LogFileFormat,
```

tabokie (commented):
Rest looks good.

src/file_pipe_log/pipe_builder.rs
```diff
@@ -181,84 +198,106 @@ impl<F: FileSystem> DualPipesBuilder<F> {
...
-                            f.handle.truncate(0)?;
-                        } else {
-                            error!(
-                                "Failed to open log file due to broken header: {:?}:{}",
-                                queue, f.seq
-                            );
-                            return Err(e);
+                    // Attention please, to make sure that the `Err` message coud be captured,
+                    // the reader is just a `LogFileReader` without loadin the file header
+                    // in advance, by passing `Version::V1` as the default param.
```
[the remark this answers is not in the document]
LykxSassinator: My fault, missed it.
It' already updated.

src/file_pipe_log/log_file.rs
LykxSassinator: Done

src/file_pipe_log/log_file.rs
LykxSassinator: Yes, missed it. It's a duplicated fault.

src/file_pipe_log/log_file.rs
```diff
@@ -164,8 +203,56 @@ impl<F: FileSystem> LogFileReader<F> {
...
+    /// `[LogFileFormat]`.
+    ///
+    /// Attention please, this function would move the `reader.offset`
+    /// to `0`, that is, the beginning of the file, to parse the
+    /// related `[LogFileFormat]`.
+    pub fn parse_format(&mut self) -> Result<LogFileFormat> {
+        // Here, the caller expected that the given `handle` has pointed to
+        // a log file with valid format. Otherwise, it should return with
+        // `Err`.
+        let file_size: usize = match self.handle.file_size() {
```
tabokie, line 216:
```suggestion
        let file_size = self.handle.file_size()?;
```

src/file_pipe_log/log_file.rs
```diff
@@ -164,8 +203,56 @@ impl<F: FileSystem> LogFileReader<F> {
...
+            match reader.read(buf) {
+                Ok(0) => {
+                    break;
+                }
+                Ok(n) => {
+                    buf = &mut buf[n..];
+                }
+                Err(e) => return Err(Error::Io(e)),
+            }
+        }
```
tabokie, line 244: Use `self.read_to`.

PR #227, opened 2022-06-22
Record perf contexts to the write procedure

sticnarf:
This PR tries to add RocksDB-like perf contexts to Raft Engine.

Now, I only add these statistics:

- `log_populating_nanos`
- `write_leader_wait_nanos`
- `log_write_nanos`
- `log_rotate_nanos`
- `log_sync_nanos`
- `write_apply_nanos`

I'm willing to know if there are other important metrics that is worth recording here.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 
 * Add `is_empty` to `Engine` API.
 * Add metadata deletion capability to `FileSystem` trait. Users can implement `exists_metadata` and `delete_metadata` to clean up obsolete metadata from older versions of Raft Engine.
+* Add `PerfContext` which records detailed time breakdown of the write process to thread-local storage.
 
 ## [0.2.0] - 2022-05-25
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,11 +6,14 @@
 
 * Unconditionally tolerate `fallocate` failures as a fix to its portability issue. Errors other than `EOPNOTSUPP` will still emit a warning.
 
+### New Features
+
+* Add `PerfContext` which records detailed time breakdown of the write process to thread-local storage.
+
 ### Public API Changes
 
 * Add `is_empty` to `Engine` API.
 * Add metadata deletion capability to `FileSystem` trait. Users can implement `exists_metadata` and `delete_metadata` to clean up obsolete metadata from older versions of Raft Engine.
-* Add `PerfContext` which records detailed time breakdown of the write process to thread-local storage.
 
 ## [0.2.0] - 2022-05-25
 
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -182,8 +182,9 @@ where
             let entered_time = writer.entered_time.unwrap();
             ENGINE_WRITE_PREPROCESS_DURATION_HISTOGRAM
                 .observe(entered_time.saturating_duration_since(start).as_secs_f64());
-            perf_context.write_leader_wait_duration +=
+            perf_context.write_wait_duration +=
                 entered_time.saturating_duration_since(before_enter);
+            debug_assert_eq!(writer.perf_context_diff.write_wait_duration, Duration::ZERO);
             perf_context += &writer.perf_context_diff;
             set_perf_context(perf_context);
             writer.finish()?
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -24,7 +24,7 @@ use crate::metrics::*;
 use crate::pipe_log::{FileBlockHandle, FileId, LogQueue, PipeLog};
 use crate::purge::{PurgeHook, PurgeManager};
 use crate::write_barrier::{WriteBarrier, Writer};
-use crate::{Error, GlobalStats, Result};
+use crate::{perf_context, Error, GlobalStats, Result};
 
 const METRICS_FLUSH_INTERVAL: Duration = Duration::from_secs(30);
 
@@ -142,9 +142,11 @@ where
         let len = log_batch.finish_populate(self.cfg.batch_compression_threshold.0 as usize)?;
         let block_handle = {
             let mut writer = Writer::new(log_batch, sync, start);
+            let mut perf_context = get_perf_context().with(|pc| pc.take());
             if let Some(mut group) = self.write_barrier.enter(&mut writer) {
                 let now = Instant::now();
-                let _t = StopWatch::new_with(&ENGINE_WRITE_LEADER_DURATION_HISTOGRAM, now);
+                let _t = StopWatch::new_with(&*ENGINE_WRITE_LEADER_DURATION_HISTOGRAM, now);
+                let append_watch = StopWatch::new_with(perf_context!(log_write_nanos), now);
                 for writer in group.iter_mut() {
                     ENGINE_WRITE_PREPROCESS_DURATION_HISTOGRAM.observe(
                         now.saturating_duration_since(writer.start_time)
@@ -166,6 +168,7 @@ where
                     };
                     writer.set_output(res);
                 }
+                drop(append_watch);
                 if let Err(e) = self.pipe_log.maybe_sync(LogQueue::Append, sync) {
                     panic!(
                         "Cannot sync {:?} queue due to IO error: {}",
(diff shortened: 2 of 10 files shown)


src/write_barrier.rs
```diff
@@ -165,6 +167,7 @@ impl<P, O> WriteBarrier<P, O> {
     /// the leader of a set of writers, returns a [`WriteGroup`] that contains
     /// them, `writer` included.
     pub fn enter<'a>(&self, writer: &'a mut Writer<P, O>) -> Option<WriteGroup<'_, 'a, P, O>> {
+        let _t = StopWatch::new(perf_context!(write_leader_wait_nanos));
```
tabokie, line 170: `ENGINE_WRITE_PREPROCESS_DURATION_HISTOGRAM` records the time from entering `write` to the data being written to pipe. And we treat it as the wait duration before.

But both two are incorrect. A better breakdown would be {compression} -> {wait} -> {writing write group} -> {apply}. Only the leader is aware of wait duration, so he must broadcast it to other writers (via an additional payload) so they can calculate their own write duration.

sticnarf: I improve the breakdown now.

Now, before entering the write group, the write thread first snapshot its current perf context.

Only the write leader will record `write_leader_wait_nanos`, `log_write_nanos`, `log_rotate_nanos` and `log_sync_nanos`. And the diff at the leader will be broadcast to each `Writer`.

After the write group exists, each writer thread will recover its perf context snapshot and add the diff to it.

tabokie:
@sticnarf It's getting a little messy, I'm thinking something like this:

image

It basically decouples perf context from prometheus with a new interface `fn observe(t: &mut Instant) -> Duration`.

src/metrics.rs
```diff
@@ -1,36 +1,140 @@
...
+}
+
+/// PerfContext records cumulative performance statistics of operations.
+///
+/// Raft Engine will update the data in the thread-local PerfContext whenever
+/// an opeartion is performed.
+#[derive(Debug, Clone, Default)]
+pub struct PerfContext {
+    /// Time spent encoding and compressing log entries.
+    pub log_populating_nanos: u64,
```
tabokie, line 48: Maybe `Duration`? No reason to copycat RocksDB here.

sticnarf: Ok

sticnarf:
> It basically decouples perf context from prometheus with a new interface `fn observe(t: &mut Instant) -> Duration`.

I think `StopWatch` is sometimes more convenient. Now I still keep but add an `observe_since` for cases suitable to pass in just an `Instant`.

sticnarf:
I think it important to pass the breakdown to the followers. For each thread, it still gives mostly accurate information about how long is spent inside raft-engine for each reason.

src/engine.rs
```diff
@@ -142,9 +142,12 @@ where
         let len = log_batch.finish_populate(self.cfg.batch_compression_threshold.0 as usize)?;
         let block_handle = {
             let mut writer = Writer::new(log_batch, sync, start);
+            // Snapshot and clear the current perf context temporarily, so the write group
+            // leader will collect the perf context diff later.
+            let mut perf_context = get_perf_context().with(|pc| pc.take());
```
tabokie, line 147: How about wrapping these to a function.

src/engine.rs
```diff
@@ -166,14 +169,26 @@ where
...
+                perf_context!(log_write_duration).observe_since(now);
                 if let Err(e) = self.pipe_log.maybe_sync(LogQueue::Append, sync) {
                     panic!(
                         "Cannot sync {:?} queue due to IO error: {}",
                         LogQueue::Append,
                         e
                     );
                 }
+                // Pass the perf context diff to all the writers.
+                get_perf_context().with(|pc| {
```
tabokie, line 181: ditto. Take the perf context and move loop outside.

tabokie (commented):
Update the changelog and add a simple test?

src/engine.rs
```diff
@@ -142,9 +142,12 @@ where
...
             let mut writer = Writer::new(log_batch, sync, start);
+            // Snapshot and clear the current perf context temporarily, so the write group
+            // leader will collect the perf context diff later.
+            let mut perf_context = get_perf_context().with(|pc| pc.take());
             if let Some(mut group) = self.write_barrier.enter(&mut writer) {
                 let now = Instant::now();
-                let _t = StopWatch::new_with(&ENGINE_WRITE_LEADER_DURATION_HISTOGRAM, now);
+                let _t = StopWatch::new_with(&*ENGINE_WRITE_LEADER_DURATION_HISTOGRAM, now);
                 for writer in group.iter_mut() {
                     ENGINE_WRITE_PREPROCESS_DURATION_HISTOGRAM.observe(
```
tabokie, line 152: write_leader_wait_duration should be recorded here. The leader wait duration doesn't equal followers' wait duration.

src/file_pipe_log/log_file.rs
```diff
@@ -111,7 +111,10 @@ impl<F: FileSystem> LogFileWriter<F> {
 
     pub fn sync(&mut self) -> Result<()> {
         if self.last_sync < self.written {
-            let _t = StopWatch::new(&LOG_SYNC_DURATION_HISTOGRAM);
+            let _t = StopWatch::new((
+                &*LOG_SYNC_DURATION_HISTOGRAM,
+                perf_context!(log_rotate_duration),
```
tabokie, line 116: sync not rotate. And the duration here isn't equivalent to the duration of `pipe::sync`. Because rotate will sync as well (both file and directory), which should be recorded by another counter.

sticnarf: I'll fix it

src/engine.rs
sticnarf: But I don't think it makes much sense to distinguish leader and followers' wait duration. The follower will not finish until a whole write group is done. And because of this, the `log_write_duration` of each writer is actually the total duration of all writers of the write group.

If we need to be more accurate and record each followers' wait duration, we also have to record separated `write_duration` for each writer, and also record how long it waits until other writers finish writing before syncing. This seems over-complicated to me...

tabokie: I agree with you that only the group's write time needs to be captured. So the end time to calculate wait_duration should be the same. But the start time is different. More specifically the follower wait time is longer than the leader wait time, in the worst case the difference will equal last write group's write time.

It's kind of hard to get a local perf event in your current implementation. I suggest changing the old `start_time` field in Writer to `end_time`, and recording the wait duration individually after exiting group.

sticnarf: Got it. Let me try to fix it.

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 
 * Add `is_empty` to `Engine` API.
 * Add metadata deletion capability to `FileSystem` trait. Users can implement `exists_metadata` and `delete_metadata` to clean up obsolete metadata from older versions of Raft Engine.
+* Add `PerfContext` which records detailed time breakdown of the write process to thread-local storage.
```
tabokie, line 13: Let's move it to new feature.

src/metrics.rs
```diff
@@ -1,36 +1,152 @@
...
+///
+/// Raft Engine will update the data in the thread-local PerfContext whenever
+/// an opeartion is performed.
+#[derive(Debug, Clone, Default)]
+pub struct PerfContext {
+    /// Time spent encoding and compressing log entries.
+    pub log_populating_duration: Duration,
+
+    /// Time spent waiting for becoming the write leader.
+    pub write_leader_wait_duration: Duration,
```
tabokie, line 50:
```suggestion
    /// Time spent waiting for the write group to form and get processed.
    pub write_wait_duration: Duration,
```

src/engine.rs
```diff
@@ -166,14 +165,27 @@ where
...
                 }
+                // Pass the perf context diff to all the writers.
+                let diff = get_perf_context();
+                for writer in group.iter_mut() {
+                    writer.perf_context_diff = diff.clone();
+                }
             }
+            let entered_time = writer.entered_time.unwrap();
+            ENGINE_WRITE_PREPROCESS_DURATION_HISTOGRAM
+                .observe(entered_time.saturating_duration_since(start).as_secs_f64());
```
tabokie, line 184: debug assert perf_context_diff.write_wait_duration == 0.

tabokie (commented):
Rest LG.

tabokie (approved):

PR #229, opened 2022-06-22
Add metadata deletion capability to `FileSystem` trait

tabokie:
User should be allowed to manage file deletion. For example, in TiKV, file encryption key can be deleted once the corresponding file is deleted.


Connor1996 (approved):
LGTM

PR #232, opened 2022-06-24
simplify `delete_metadata` trait method

tabokie:
There is no reason to return `exist` when we already have `exists_metadata` method.


Connor1996 (approved):
LGTM

PR #242, opened 2022-07-13
refactor: catch errno EINTR when reading and writing files instead of EAGAIN

guoxiangCN:
EAGAIN error code Appears only when the O_NONBLOCK option is set and there is no more content in the file to read.
If a read() is interrupted by a signal before it reads any data, it shall return -1 with errno set to [EINTR].

### DIFF
--- a/src/env/default.rs
+++ b/src/env/default.rs
@@ -100,7 +100,7 @@ impl LogFd {
             });
             let bytes = match pread(self.0, &mut buf[readed..], offset as i64) {
                 Ok(bytes) => bytes,
-                Err(e) if e == Errno::EAGAIN => continue,
+                Err(e) if e == Errno::EINTR => continue,
                 Err(e) => return Err(from_nix_error(e, "pread")),
             };
             // EOF
@@ -121,7 +121,7 @@ impl LogFd {
         while written < content.len() {
             let bytes = match pwrite(self.0, &content[written..], offset as i64) {
                 Ok(bytes) => bytes,
-                Err(e) if e == Errno::EAGAIN => continue,
+                Err(e) if e == Errno::EINTR => continue,
                 Err(e) => return Err(from_nix_error(e, "pwrite")),
             };
             if bytes == 0 {


tabokie:
Thanks for the attention.

According to [rust](https://doc.rust-lang.org/std/io/trait.Write.html#tymethod.write), a `Write::write` shouldn't write any bytes if an error is returned:
> If an error is returned then no bytes in the buffer were written to this writer.

So it appears we should remove the retry looping entirely. For consistency do the same for the `read` as well.

FYI we already have retry logic at a higher level, so this change won't affect the IO results.
https://github.com/tikv/raft-engine/blob/ee0f6cf0d9fd5c4839463bd84edbc009402dffda/src/file_pipe_log/log_file.rs#L203
https://github.com/tikv/raft-engine/blob/ee0f6cf0d9fd5c4839463bd84edbc009402dffda/src/file_pipe_log/log_file.rs#L107

LykxSassinator:
How about direclty add an extra option `OFlag::O_NONBLOCK` for the `open` and `create` api rather than modify the `Errno` ? I think the original design on it is to use `OFlag::O_NONBLOCK` as the default option for `fd`.

guoxiangCN:
Okay, I didn't notice the higher judgment.  However, according to the constraints of the write method you mentioned, retrying the entire while loop will also lead to a semi successful and semi failed state of data writing, that is, partial success and partial failure.
However, I'm not sure whether I need to cross the loop to really write. I don't know much about the underlying file system writing.

tabokie:
Having looked closely, I think the implementation here is not very important. We maintain a file offset and use it to write, so a partial successful write can be overwritten correctly afterwards:
https://github.com/tikv/raft-engine/blob/d2b41e0220edfe5498c52da677a37dec931e2131/src/env/default.rs#L215-L216

And the `pwrite` interface doesn't guarantee atomicity at all. (In fact non of the unix syscalls guarantees that, considering direct IO). So your current changes look okay.

tabokie:
> I think the original design on it is to use OFlag::O_NONBLOCK as the default option for fd.

@LykxSassinator No, non block is only useful for network fd.

tabokie (approved):

Issue #235, opened 2022-07-08
cargo +nightly clippy --all --all-features --all-targets -- -D clippy::all  error


jiangshifu:
![image](https://user-images.githubusercontent.com/79852712/178027447-caa7003e-fa49-4f01-a2c8-ff2c9d7a39ee.png)

commit 7a436eae40a6b62371123c96941e058b7fe52b63 (HEAD -> master, origin/master, origin/HEAD)
Author: Xinye Tao <xy.tao@outlook.com>
Date:   Tue Jun 28 14:09:49 2022 +0800

    support supplying file system to run ctl command (#231)
    
    Signed-off-by: tabokie <xy.tao@outlook.com>

cat /etc/centos-release 
CentOS Stream release 8

tabokie:
To avoid constantly rebuilding the cache, we use a specific nightly toolchain in our CI. That's why these issues are not caught early:
https://github.com/tikv/raft-engine/blob/7a436eae40a6b62371123c96941e058b7fe52b63/.github/workflows/rust.yml#L22

Feel free to fix these issues, and upgrade CI toolchain as well.

tabokie:
To be future-proof, I'll keep the issue here even though #244 will fix it temporarily. Contributions are always welcome if you find the CI toolchain is outdated.

PR #244, opened 2022-07-15
chore: update CI toolchain and clean up code

tabokie:
Ref #235

Update toolchain to 2022-07-13, add a makefile.

A lot of minor refactoring:
- Refactor `scan_messages` interface to share more code paths, and add tests for them.
- Hide `LogFileFormat` from outer crate, it should stay an implementation detail of file_pipe_log.
- Change `DataLayout` to `u64`, because otherwise it needs additional care to make sure `Alignment(0)` and `NoAlignment` don't coexist.
- Remove some tests that aren't very useful.


Connor1996 (approved):
LGTM

src/file_pipe_log/log_file.rs
```diff
@@ -148,59 +141,45 @@ impl<F: FileSystem> LogFileWriter<F> {
...
-                };
-                reader.parse_format()?;
-                Ok(reader)
-            }
+    fn open(handle: Arc<F::Handle>, reader: F::Reader) -> LogFileReader<F> {
+        Self {
+            handle,
+            reader,
+            // Set to an invalid offset to force a reseek at first read.
+            offset: u64::MAX,
```
LykxSassinator, line 166: I've got the purpose that u refactor the `LogFileReader` and its related method `build_file_reader`. But referring to the design on `LogFileWriter`, I still recommend to make the `parse_format` integrated into the processing of `LogFileReader::open()`, rather than `open()` -> manually `parse_format` by callers.

LykxSassinator (commented):
Rest LGTM

src/file_pipe_log/format.rs
```diff
@@ -93,127 +93,76 @@ pub(super) fn lock_file_path<P: AsRef<Path>>(dir: P) -> PathBuf {
...
-
-impl Default for LogFileFormat {
-    fn default() -> Self {
-        Self {
-            version: Version::default(),
-            data_layout: DataLayout::NoAlignment,
-        }
-    }
+    pub version: Version,
+    pub alignment: u64,
```
LykxSassinator, line 99: Recommend to supply a necessary annotations to the meaning of different value on `alignment`. Such as:
``` Rust
...
/// `0` => no alignment.
/// `_` => aligned block size.
/// Default: `0`.
pub alignment: u64,
...
```

src/pipe_log.rs
```diff
@@ -115,60 +116,28 @@ impl Default for Version {
...
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{}", self.to_u64().unwrap())
     }
 }
 
 #[derive(Debug, Clone)]
 pub struct LogFileContext {
     pub id: FileId,
-    pub format: LogFileFormat,
+    pub version: Version,
```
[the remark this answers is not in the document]
tabokie: > Hide `LogFileFormat` from outer crate, it should stay an implementation detail of file_pipe_log.

There's no need to expose it because only "file_pipe_log" needs to know about alignment.

src/file_pipe_log/log_file.rs
tabokie: My change removes the `format` stored inside reader, because reader doesn't need a format to function. The writer, on the other hand, needs to do format-specific initialization, so we can't refactor it like reader.

Makefile
```diff
@@ -0,0 +1,34 @@
...
+	cargo clippy --all --features all_stable --all-targets -- -D clippy::all
+else
+clippy:
+	cargo clippy --all --all-features --all-targets -- -D clippy::all
+endif
+
+## Run tests.
+ifeq ($(WITH_STABLE_TOOLCHAIN), true)
+test:
+	cargo test --all --features all_stable_except_failpoints ${EXTRA_CARGO_ARGS} -- --nocapture
```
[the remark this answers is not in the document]
tabokie: Do you encounter any problem using it? I think the default `auto` should work good enough.

src/file_pipe_log/reader.rs
```diff
@@ -42,9 +44,15 @@ impl<F: FileSystem> LogItemBatchFileReader<F> {
...
-        self.file_context = Some(LogFileContext::new(file_id, *reader.file_format()));
+    pub fn open(
+        &mut self,
+        file_id: FileId,
+        format: LogFileFormat,
+        reader: LogFileReader<F>,
+    ) -> Result<()> {
+        self.valid_offset = LogFileFormat::encode_len(format.version);
+        self.file_id = Some(file_id);
+        self.format = Some(format);
```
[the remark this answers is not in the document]
tabokie: Similar to my other comment, because the context no longer holds alignment, using it doesn't make sense in this context.

Makefile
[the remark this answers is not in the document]
LykxSassinator: Nop，it's work fine in my environment, also on MacOS.

Makefile
```diff
@@ -0,0 +1,34 @@
...
+format:
+	cargo fmt --all
+
+## Run clippy.
+ifeq ($(WITH_STABLE_TOOLCHAIN), true)
+clippy:
+	cargo clippy --all --features all_stable --all-targets -- -D clippy::all
+else
+clippy:
+	cargo clippy --all --all-features --all-targets -- -D clippy::all
```
LykxSassinator, line 22: Here, do we not need `+nightly` trait any more when `${WITH_STABLE_TOOLCHAIN} == false` ?

I've re-test the makefile on Centos:
`Linux version 3.10.0-862.14.4.el7.x86_64 (mockbuild@kbuilder.bsys.centos.org) (gcc version 4.8.5 20150623 (Red Hat 4.8.5-28) (GCC) ) #1 SMP Wed Sep 26 15:12:11 UTC 2018`

And it returns the following failure: 
```Rust
error[E0554]: `#![feature]` may not be used on the stable release channel
  --> src/lib.rs:16:34
   |
16 | #![cfg_attr(feature = "nightly", feature(test))]
   |                                  ^^^^^^^^^^^^^

error[E0554]: `#![feature]` may not be used on the stable release channel
  --> src/lib.rs:17:31
```

And if I add the `+nightly` trait, both `make test` and `make clippy` worked as expectation without errs, by setting `${WITH_STABLE_TOOLCHAIN} == false` .

tabokie: The idea is to assume user has set the correct default toolchain. But force setting sounds okay too, I'll add it later.

PR #245, opened 2022-07-15
pipe: reseek after write failure

### LINKED ISSUE #243: Interrupted write causes corruption
At this line, we use `write_all` to append some bytes:
https://github.com/tikv/raft-engine/blob/ee0f6cf0d9fd5c4839463bd84edbc009402dffda/src/file_pipe_log/log_file.rs#L107

If this write is interrupted, we directly bubble its error. But some portion of the data might already be written. In this case, the `self.written` is inconsistent with underlying writer's internal offset. A fractured write will remain as a phantom record.

tabokie:
Close #243


LykxSassinator (approved):
LGTM

Connor1996 (approved):

Issue #246, opened 2022-07-16
Support reading block based log format


tabokie:
Decode log batch from fixed-sized blocks. See a detailed design in [RocksDB wiki](https://github.com/facebook/rocksdb/wiki/Write-Ahead-Log-File-Format).

The write flow can be implemented later. Read is of higher priority to keep it in the format version V2.

This is a funding work for #168.

LykxSassinator:
Got it. The basic structure has been built, and I need to supply relevant uts for it.

PR #249, opened 2022-07-21
Support reading block based log format

LykxSassinator:
## PR Description
 
For closing #246, this pr is set. This pr is used for building basics for the new DIO feature, including:
* [New] DataLayout, representing the arrangement on data (records), in alignment or not.
* Refactor LogFileFormat / LogFileContext.
* Add extra support in `stress` tool for the new `alignment-mode` reading.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,7 +35,7 @@ fail = "0.5"
 fs2 = "0.4"
 hashbrown = "0.12"
 hex = "0.4"
-if_chain = "1.0.2"
+if_chain = "1.0"
 lazy_static = "1.3"
 libc = "0.2"
 log = { version = "0.4", features = ["max_level_trace", "release_max_level_debug"] }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,6 +35,7 @@ fail = "0.5"
 fs2 = "0.4"
 hashbrown = "0.12"
 hex = "0.4"
+if_chain = "1.0.2"
 lazy_static = "1.3"
 libc = "0.2"
 log = { version = "0.4", features = ["max_level_trace", "release_max_level_debug"] }
--- a/src/config.rs
+++ b/src/config.rs
@@ -3,7 +3,7 @@
 use log::warn;
 use serde::{Deserialize, Serialize};
 
-use crate::pipe_log::{DataLayout, Version};
+use crate::pipe_log::Version;
 use crate::{util::ReadableSize, Result};
 
 const MIN_RECOVERY_READ_BLOCK_SIZE: usize = 512;
@@ -37,7 +37,7 @@ pub struct Config {
     pub recovery_mode: RecoveryMode,
     /// Minimum I/O size for reading log files during recovery.
     ///
-    /// Default: "4KB". Minimum: "512B".
+    /// Default: "16KB". Minimum: "512B".
     pub recovery_read_block_size: ReadableSize,
     /// The number of threads used to scan and recovery log files.
     ///
@@ -59,11 +59,6 @@ pub struct Config {
     ///
     /// Default: 1
     pub format_version: Version,
-
-    /// Layout of data in log file.
-    ///
-    /// Default: "no-alignment"
-    pub format_data_layout: DataLayout,
 
     /// Target file size for rotating log files.
     ///
@@ -109,7 +104,6 @@ impl Default for Config {
             batch_compression_threshold: ReadableSize::kb(8),
             bytes_per_sync: ReadableSize::mb(4),
             format_version: Version::V1,
-            format_data_layout: DataLayout::NoAlignment,
             target_file_size: ReadableSize::mb(128),
             purge_threshold: ReadableSize::gb(10),
             purge_rewrite_threshold: None,
@@ -168,12 +162,6 @@ impl Config {
                 ));
             }
         }
-        if !self.format_data_layout.is_enabled() {
-            return Err(box_err!(
-                "format_data_layout: {:?} is not support to be enabled",
-                self.format_data_layout
-            ));
-        }
         #[cfg(not(feature = "swap"))]
         if self.memory_limit.is_some() {
             warn!("memory-limit will be ignored because swap feature is not enabled");
@@ -227,7 +215,6 @@ mod tests {
         assert_eq!(load.purge_threshold, ReadableSize::mb(3));
         assert_eq!(load.format_version, Version::V1);
         assert!(!load.enable_log_recycle);
-        assert_eq!(load.format_data_layout, DataLayout::NoAlignment);
     }
 
     #[test]
@@ -288,49 +275,5 @@ mod tests {
         assert!(toml::to_string(&load)
             .unwrap()
             .contains("tolerate-corrupted-tail-records"));
-    }
-
-    #[test]
-    fn test_format_data_layout() {
-        // "no-alignment" data_layout - default
-        {
-            let default_cfg = r#"
-                format-data-layout = "no-alignment"
-            "#;
-            let mut load: Config = toml::from_str(default_cfg).unwrap();
-            load.sanitize().unwrap();
-            assert_eq!(load.format_data_layout, DataLayout::NoAlignment);
-        }
-        // "align-with-integration"
-        {
-            let cfg = r#"
-                format-data-layout = "align-with-integration"
-            "#;
-            let mut load: Config = toml::from_str(cfg).unwrap();
-            load.sanitize().unwrap();
-            assert_eq!(load.format_data_layout, DataLayout::AlignWithIntegration);
-        }
-        // "align-with-fragments"
-        {
-            let cfg = r#"
-                format-data-layout = "align-with-fragments"
-            "#;
-            let mut load: Config = toml::from_str(cfg).unwrap();
-            assert!(load.sanitize().is_err());
-        }
-        // abnormal
-        {
-            let abnormal_cfg_1 = r#"
-                data-layout = "align"
-            "#;
-            let mut load: Config = toml::from_str(abnormal_cfg_1).unwrap();
-            load.sanitize().unwrap();
-            assert_eq!(load.format_data_layout, DataLayout::NoAlignment);
-
-            let abnormal_cfg_2 = r#"
-                format-data-layout = "aligned-with-fragment"
-            "#;
-            assert!(toml::from_str::<Config>(abnormal_cfg_2).is_err());
-        }
     }
 }
--- a/src/config.rs
+++ b/src/config.rs
@@ -3,7 +3,7 @@
 use log::warn;
 use serde::{Deserialize, Serialize};
 
-use crate::pipe_log::Version;
+use crate::pipe_log::{DataLayout, Version};
 use crate::{util::ReadableSize, Result};
 
 const MIN_RECOVERY_READ_BLOCK_SIZE: usize = 512;
@@ -60,6 +60,11 @@ pub struct Config {
     /// Default: 1
     pub format_version: Version,
 
+    /// Layout of data in log file.
+    ///
+    /// Default: "no-alignment"
+    pub format_data_layout: DataLayout,
+
     /// Target file size for rotating log files.
     ///
     /// Default: "128MB"
@@ -104,6 +109,7 @@ impl Default for Config {
             batch_compression_threshold: ReadableSize::kb(8),
             bytes_per_sync: ReadableSize::mb(4),
             format_version: Version::V1,
+            format_data_layout: DataLayout::NoAlignment,
             target_file_size: ReadableSize::mb(128),
             purge_threshold: ReadableSize::gb(10),
             purge_rewrite_threshold: None,
@@ -215,6 +221,7 @@ mod tests {
         assert_eq!(load.purge_threshold, ReadableSize::mb(3));
         assert_eq!(load.format_version, Version::V1);
         assert!(!load.enable_log_recycle);
+        assert_eq!(load.format_data_layout, DataLayout::NoAlignment);
     }
 
     #[test]
@@ -275,5 +282,41 @@ mod tests {
         assert!(toml::to_string(&load)
             .unwrap()
             .contains("tolerate-corrupted-tail-records"));
+    }
+
+    #[test]
+    fn test_setting_on_alignment_mode() {
+        // "no-alignment" data_layout - default
+        {
+            let default_cfg = r#"
+                format-data-layout = "no-alignment"
+            "#;
+            let mut load: Config = toml::from_str(default_cfg).unwrap();
+            load.sanitize().unwrap();
+            assert_eq!(load.format_data_layout, DataLayout::NoAlignment);
+        }
+        // "alignment"
+        {
+            let cfg = r#"
+                format-data-layout = "alignment"
+            "#;
+            let mut load: Config = toml::from_str(cfg).unwrap();
+            load.sanitize().unwrap();
+            assert_eq!(load.format_data_layout, DataLayout::Alignment);
+        }
+        // abnormal
+        {
+            let abnormal_cfg_1 = r#"
+                data-layout = "align"
+            "#;
+            let mut load: Config = toml::from_str(abnormal_cfg_1).unwrap();
+            load.sanitize().unwrap();
+            assert_eq!(load.format_data_layout, DataLayout::NoAlignment);
+
+            let abnormal_cfg_2 = r#"
+                format-data-layout = "alignments"
+            "#;
+            assert!(toml::from_str::<Config>(abnormal_cfg_2).is_err());
+        }
     }
 }
--- a/src/config.rs
+++ b/src/config.rs
@@ -168,6 +168,12 @@ impl Config {
                 ));
             }
         }
+        if !self.format_data_layout.is_enabled() {
+            return Err(box_err!(
+                "format_data_layout: {:?} is not support to be enabled",
+                self.format_data_layout
+            ));
+        }
         #[cfg(not(feature = "swap"))]
         if self.memory_limit.is_some() {
             warn!("memory-limit will be ignored because swap feature is not enabled");
@@ -285,7 +291,7 @@ mod tests {
     }
 
     #[test]
-    fn test_setting_on_alignment_mode() {
+    fn test_format_data_layout() {
         // "no-alignment" data_layout - default
         {
             let default_cfg = r#"
@@ -295,14 +301,22 @@ mod tests {
             load.sanitize().unwrap();
             assert_eq!(load.format_data_layout, DataLayout::NoAlignment);
         }
-        // "alignment"
+        // "align-with-integration"
         {
             let cfg = r#"
-                format-data-layout = "alignment"
+                format-data-layout = "align-with-integration"
             "#;
             let mut load: Config = toml::from_str(cfg).unwrap();
             load.sanitize().unwrap();
-            assert_eq!(load.format_data_layout, DataLayout::Alignment);
+            assert_eq!(load.format_data_layout, DataLayout::AlignWithIntegration);
+        }
+        // "align-with-fragments"
+        {
+            let cfg = r#"
+                format-data-layout = "align-with-fragments"
+            "#;
+            let mut load: Config = toml::from_str(cfg).unwrap();
+            assert!(load.sanitize().is_err());
         }
         // abnormal
         {
@@ -314,7 +328,7 @@ mod tests {
             assert_eq!(load.format_data_layout, DataLayout::NoAlignment);
 
             let abnormal_cfg_2 = r#"
-                format-data-layout = "alignments"
+                format-data-layout = "aligned-with-fragment"
             "#;
             assert!(toml::from_str::<Config>(abnormal_cfg_2).is_err());
         }
--- a/src/file_pipe_log/format.rs
+++ b/src/file_pipe_log/format.rs
@@ -192,8 +192,8 @@ impl LogFileFormat {
                 // Here, we mark this special err, that is, corrupted `payload`,
                 // with InvalidArgument.
                 Err(Error::InvalidArgument(format!(
-                    "invalid data_layout in the header, len: {}",
-                    buf_len - Self::header_len()
+                    "invalid dataload in the header, len: {}, expected len: {}",
+                    buf_len - Self::header_len(), Self::payload_len(version)
                 )))
             }
         }
--- a/src/file_pipe_log/format.rs
+++ b/src/file_pipe_log/format.rs
@@ -2,7 +2,6 @@
 
 //! Representations of objects in filesystem.
 
-use std::cmp::Ordering;
 use std::io::BufRead;
 use std::path::{Path, PathBuf};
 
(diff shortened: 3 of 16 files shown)


LykxSassinator:
More uts for covering the abnormal code paths wait to be supplemented.

LykxSassinator:
The pr should hold. 
Currently, this design is compatible to aligned records or raw records without alignment. However, it lacks compatibilities to fragmented records in alignment, which shoud be re-considered.

LykxSassinator:
Please hold on.
This pr considers the format of aligned records and raw records without alignment. However, it still lack the compatibility to fragmented records, which should be re-considered.

LykxSassinator:
> Please hold on. This pr considers the format of aligned records and raw records without alignment. However, it still lack the compatibility to fragmented records, which should be re-considered.

LykxSassinator:
> Please hold on. This pr considers the format of aligned records and raw records without alignment. However, it still lack the compatibility to fragmented records, which should be re-considered.

Done for it.

src/config.rs
```diff
@@ -60,6 +60,11 @@ pub struct Config {
     /// Default: 1
     pub format_version: Version,
 
+    /// Layout of data in log file.
+    ///
+    /// Default: "no-alignment"
+    pub format_data_layout: DataLayout,
```
tabokie, line 66: Why adding a configuration? Everything should be read from file header.

LykxSassinator: Yes, it's a necessary configuration to user to use which format to dump `LogBatch`s into log files.
For `write`:
* If `no-alignment`, the log content will be dumped into log files directly, without alignment or fragments.
* If `align-with-integration`, the log content will be aligned with default block size, and dumped into log files. [For future DIO]
* If `align-with-fragments`, the `LogBatch` will be aligned and divided into several parts, typed with `Full || First || Middle || Last` into the log file.

For `read`:
This context in the header will  be parsed and make the `Reader` how to read the context from the file, no aligned format, aligned format in integration or aligned format in fragments.

LykxSassinator: > Yes, it's a necessary configuration to user to use which format to dump `LogBatch`s into log files. For `write`:
> 
> * If `no-alignment`, the log content will be dumped into log files directly, without alignment or fragments.
> * If `align-with-integration`, the log content will be aligned with default block size, and dumped into log files. [For future DIO]
> * If `align-with-fragments`, the `LogBatch` will be aligned and divided into several parts, typed with `Full || First || Middle || Last` into the log file.
> 
> For `read`: This context in the header will be parsed and make the `Reader` how to read the context from the file, no aligned format, aligned format in integration or aligned format in fragments.
> 
> Meanwhile, this configuration is a compatible design to make the parsing strategy, when `recovery` or `get_entry`, on current format compatible to the future fragmented records in log file.

Ignore it. This configuration is over-designed. I'll change it.

tabokie (commented):
I think a bit more on this and I think RocksDB's format is not very efficient (many memcpy-s). And we require accessing raw entry data via a file block handle, so the data of one log batch should stay contiguous.

So there's no need to handling the `Type` as RocksDB does. The way I see it, the only work is to add a padding detection to the existing reader.

When reading a log batch, first `peek(offset, end_of_current_block)` and construct header. If this slice turns out to be padded with zeros, then go to the next block.

After the header is ready, we get the rest of the data just like before.

LykxSassinator:
Agree!
So I refactor the reading strategy and remove the over-designed structure in the previous commit.
> so the data of one log batch should stay contiguous.

About the padding detection, u menthioned before:
> the only work is to add a padding detection to the existing reader.
> When reading a log batch, first peek(offset, end_of_current_block) and construct header. If this slice turns out to be padded with zeros, then go to the next block.

Thanks for your suggestion, making me find a existed bug in the preivous detection strategy. And this bug has been fixed in this commit.

src/file_pipe_log/format.rs
```diff
@@ -19,6 +19,11 @@ const LOG_APPEND_SUFFIX: &str = ".raftlog";
 const LOG_REWRITE_SUFFIX: &str = ".rewrite";
 /// File header.
 const LOG_FILE_MAGIC_HEADER: &[u8] = b"RAFT-LOG-FILE-HEADER-9986AB3E47F320B394C8E84916EB0ED5";
+/// Mask of Version.
+const LOG_FILE_HEADER_VERSION_MASK: u64 = 0x00FFFFFFFFFFFFFF;
```
tabokie, line 23: Don't change the version field. Just add a new payload.

When reading the header, first call `let format = LogFileFormat::decode_version(peek(LogFileFormat::header_len())`, then call `format.decode_payload(peek(format.payload_len()))`.

src/file_pipe_log/log_file.rs
```diff
@@ -10,34 +10,45 @@ use log::warn;
...
+
 /// Combination of `[Handle]` and `[Version]`, specifying a handler of a file.
 #[derive(Debug)]
 pub struct FileHandler<F: FileSystem> {
     pub handle: Arc<F::Handle>,
     pub context: LogFileContext,
 }
 
+#[inline]
+pub(crate) fn get_system_block_size() -> usize {
```
tabokie, line 33: The alignment size should be encoded in file. `Alignment(u64)`. We can later worry about the case when the system block size is different than the alignment size specified by log header.

src/file_pipe_log/log_file.rs
```diff
@@ -198,15 +225,37 @@ impl<F: FileSystem> LogFileReader<F> {
     }
 
     pub fn read(&mut self, handle: FileBlockHandle) -> Result<Vec<u8>> {
-        let mut buf = vec![0; handle.len as usize];
-        let size = self.read_to(handle.offset, &mut buf)?;
+        let (start_offset, length) = {
```
tabokie, line 228: No need to change anything here. We are still using buffered IO.

src/file_pipe_log/reader.rs
```diff
@@ -107,38 +140,76 @@ impl<F: FileSystem> LogItemBatchFileReader<F> {
...
         debug_assert!(offset >= self.buffer_offset);
+        let fs_block_size = get_system_block_size();
         let reader = self.reader.as_mut().unwrap();
         let end = self.buffer_offset + self.buffer.len();
         if offset > end {
             self.buffer_offset = offset;
-            self.buffer
-                .resize(std::cmp::max(size + prefetch, self.read_block_size), 0);
-            let read = reader.read_to(self.buffer_offset as u64, &mut self.buffer)?;
+            let (calibrate_offset, calibrate_len) = {
```
tabokie, line 148: Don't change anything in this function as well.

src/file_pipe_log/reader.rs
```diff
@@ -64,17 +65,49 @@ impl<F: FileSystem> LogItemBatchFileReader<F> {
...
                 return Err(Error::Corruption(
                     "attempt to read file with broken header".to_owned(),
                 ));
             }
-            let (footer_offset, compression_type, len) = LogBatch::decode_header(&mut self.peek(
-                self.valid_offset,
-                LOG_BATCH_HEADER_LEN,
-                0,
-            )?)?;
+            let (footer_offset, compression_type, len) = {
```
tabokie, line 82: This code is ugly, how about separate the two implementations clearly like this:
```
match data_layout {
  DataLayout::NoAlignment => { old code }
  DataLayout::Alignment(size) => {
    new code
  }
}
```

src/file_pipe_log/reader.rs
```diff
@@ -64,17 +65,49 @@ impl<F: FileSystem> LogItemBatchFileReader<F> {
...
+                            // contain an valid header of the adjacent LogBatch or paddings
+                            // with `0`.
+                            round_up(self.valid_offset, fs_block_size) - self.valid_offset
+                                + LOG_BATCH_HEADER_LEN
+                        }
+                    },
+                    0,
+                )?) {
+                    Err(e) => {
+                        // In DataLayout::Alignment mode, tail data in the previous block may be
```
tabokie, line 98: Check it is indeed padding by looking at the first and last u64.

tabokie (commented):
You also need to test the decoding properly. Since the write flow is not implemented, you need to manually assemble a block-based log file content. Some cases need to be covered:
- [ batch padding ]
- [ batch1 batch2_header ][ batch2_middle ][ batch2_rest ]
- [ batch1 batch2_header_part1 ][ batch2_header_part2 ] (this should fail)
- [ batch non_zero_padding ] (fail too)

src/config.rs
```diff
@@ -37,7 +37,7 @@ pub struct Config {
     pub recovery_mode: RecoveryMode,
     /// Minimum I/O size for reading log files during recovery.
     ///
-    /// Default: "4KB". Minimum: "512B".
+    /// Default: "16KB". Minimum: "512B".
```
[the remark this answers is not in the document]
LykxSassinator: The default value is set with `16kb`, shown in `default()`, so i changed this annotation to keep coincident with the code.

https://github.com/tikv/raft-engine/blob/70d27ee04b9e92586b62cb44c3a82b2ea7dc75b3/src/config.rs#L102

src/file_pipe_log/reader.rs
LykxSassinator: Both accepted, current format is used to seperate the extraction of `(footer_offset, compression_type, len)` and share the following decoding strategy of LogBatch.

LykxSassinator: I perfer current one? Because the strategy for loading header of LogBatch are diverged when `DataLayout` are different, and the following processing on reading and parsing LogBatch content could be kept in same code paths.

LykxSassinator: Or how about sperate the decoding of header into divided parts, such as:
``` Rust
let (footer_offset, compression_type, len) = {
       match data_layout {
            DataLayout::NoAlignment => {
                ... /* get tuple of (offset, t, l) */
            }
            DataLayout::Alignment(fs_block_size) => {
                ... /* get tuple of (offset, t, l) */
            }
        }
    };
/* Old codes for decoding LogBatch. */
```

src/file_pipe_log/reader.rs
LykxSassinator: I think its no need for it. Because what we are concerned are valid headers of LogBatch, whatever the padding is filled with, we just need to fetch and get the next valid aligned offset, followed with parsing. So, we can just skip the rest bytes by `round_up(self.valid_offset, fs_block_size as usize)`.

LykxSassinator:
> You also need to test the decoding properly. Since the write flow is not implemented, you need to manually assemble a block-based log file content. Some cases need to be covered:
> 
> * [ batch padding ]
> * [ batch1 batch2_header ][ batch2_middle ][ batch2_rest ]
> * [ batch1 batch2_header_part1 ][ batch2_header_part2 ] (this should fail)
> * [ batch non_zero_padding ] (fail too)

Got it.
* The first two case have been tested in the `failpoints/test_engine.rs: test_build_engine_with_aligned_datalayout` in the previous commit. 
* The third abnormal cornor case waited to be supplemented.
* About the last corner case, I think it should be passed. Is necessary to check the padding whether it is filled with zeros? non-zero paddings can also be skipped as mentioned above.

src/file_pipe_log/reader.rs
tabokie: You can't just skip data whenever you want. They may contain user data that is corrupted by disk failure.

src/file_pipe_log/reader.rs
```diff
@@ -64,17 +65,59 @@ impl<F: FileSystem> LogItemBatchFileReader<F> {
...
+                        )?)?;
+                        (offset, t, l)
+                    }
+                    DataLayout::Alignment(fs_block_size) => {
+                        // In DataLayout::Alignment mode, tail data in this block maybe
+                        // contain an valid header of the adjacent LogBatch or paddings
+                        // with `0`.
+                        match LogBatch::decode_header(&mut self.peek(
+                            self.valid_offset,
+                            round_up(self.valid_offset, fs_block_size as usize) - self.valid_offset
```
tabokie, line 97: Why reading it differently? You should always read `LOG_BATCH_HEADER_LEN`, but the `hint` might be different.

A better way to write it would be:
```rust
fn padding_size() -> usize; // returns the number of bytes needed to pad the offset to given alignment.
let padding = padding_size(data_layout.alignment(), self.valid_offset);
let (..) = match LogBatch::decode_header(self.valid_offset, LOG_BATCH_HEADER_LEN, padding) {
  Err(e) if padding >= LOG_BATCH_HEADER_LEN {
    // check padding is zeros.
   continue;
  }
  Ok(r) => r,
  e @ Err(_) => return r,
}
```

src/file_pipe_log/log_file.rs
```diff
@@ -10,34 +10,44 @@ use log::warn;
...
+
 /// Combination of `[Handle]` and `[Version]`, specifying a handler of a file.
 #[derive(Debug)]
 pub struct FileHandler<F: FileSystem> {
     pub handle: Arc<F::Handle>,
     pub context: LogFileContext,
 }
 
+#[inline]
+pub(crate) fn get_system_block_size() -> usize {
```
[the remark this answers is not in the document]
LykxSassinator: It's an entrance to get the system block size for building the pipe if DIO was open.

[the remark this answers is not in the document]
LykxSassinator: Accept. I'll add extra annotations for it.

LykxSassinator:
Corner cases have been supplemented in this commit.
Thanks for reviewing.

src/file_pipe_log/reader.rs
```diff
@@ -64,17 +65,60 @@ impl<F: FileSystem> LogItemBatchFileReader<F> {
...
                     "attempt to read file with broken header".to_owned(),
                 ));
             }
-            let (footer_offset, compression_type, len) = LogBatch::decode_header(&mut self.peek(
-                self.valid_offset,
-                LOG_BATCH_HEADER_LEN,
-                0,
-            )?)?;
+            let (footer_offset, compression_type, len) = {
+                match data_layout {
```
tabokie, line 82: Use [if_chain](https://docs.rs/if_chain/latest/if_chain/) to reduce the nesting depth.
```rust
let r = LogBatch::decode_header(peek(valid_offset, HEADER_LEN, 0));
if_chain! {
    if r.is_err();
    if let DataLayout::Alignment(block_size) = data_layout;
    if block_size > 0;
    let aligned_next_offset = round_up(self.valid_offset, block_size);
    if valid_offset != aligned_next_offset;
    if is_padding(peek(valid_offset, aligned_next_offset - valid_offset, 0));
    then {
        continue;
    }
}
let (footer_offset, compression_type, len) = r?;
```

src/file_pipe_log/log_file.rs
```diff
@@ -236,16 +245,17 @@ impl<F: FileSystem> LogFileReader<F> {
...
+        let header_len = LogFileFormat::header_len();
         if file_size < header_len {
             return Err(Error::Corruption("Invalid header of LogFile!".to_owned()));
         }
-        // [2] Parse the header of the file.
-        let mut container = vec![0; header_len];
+        // [2] Parse the format of the file.
+        let mut container =
+            vec![0; LogFileFormat::header_len() + LogFileFormat::payload_len(Version::V2)];
         self.read_to(0, &mut container[..])?;
```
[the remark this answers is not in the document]
LykxSassinator: As discussed, `read_to` will return the actual fetched bytes without EOF error. So, this case won't happen.

src/file_pipe_log/format.rs
```diff
@@ -20,6 +21,43 @@ const LOG_REWRITE_SUFFIX: &str = ".rewrite";
 /// File header.
 const LOG_FILE_MAGIC_HEADER: &[u8] = b"RAFT-LOG-FILE-HEADER-9986AB3E47F320B394C8E84916EB0ED5";
 
+pub(crate) fn is_valid_paddings(buf: &[u8]) -> Result<bool> {
+    let len = buf.len();
```
tabokie, line 25: Let's change it to test the first and last u8 instead. return buf.is_empty() || buf[0] == 0 && buf[buf.len() - 1] == 0

src/file_pipe_log/reader.rs
```diff
@@ -64,17 +65,54 @@ impl<F: FileSystem> LogItemBatchFileReader<F> {
...
+                    DataLayout::NoAlignment | DataLayout::Alignment(0) => LogBatch::decode_header(
+                        &mut self.peek(self.valid_offset, LOG_BATCH_HEADER_LEN, 0)?,
+                    )?,
+                    DataLayout::Alignment(fs_block_size) => {
+                        // In DataLayout::Alignment mode, tail data in this block maybe
+                        // contain an valid header of the adjacent LogBatch or paddings
+                        // with `0`.
+                        let aligned_next_offset =
+                            round_up(self.valid_offset, fs_block_size as usize);
+                        let parser = LogBatch::decode_header(&mut self.peek(
```
tabokie, line 91: No, the point of using if_chain is to not match anymore. Two modes can share the decode_header (the hint is the same as well, in fact only in rare cases we need to access the end of block, we should optimize for common case where the decode_header succeeds).

src/file_pipe_log/pipe.rs
```diff
@@ -321,7 +339,19 @@ impl<F: FileSystem> SinglePipe<F> {
         let seq = active_file.seq;
         let writer = &mut active_file.writer;
 
-        let start_offset = writer.offset();
+        let start_offset = {
```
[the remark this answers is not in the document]
tabokie:
```rust
#[cfg(feature = "failpoint")]
{
  // padding logic
}
let start_offset = writer.offset();
```

src/file_pipe_log/pipe.rs
```diff
@@ -135,6 +136,24 @@ impl<F: FileSystem> SinglePipe<F> {
         mut fds: VecDeque<FileHandler<F>>,
         capacity: usize,
     ) -> Result<Self> {
+        let data_layout = {
+            let force_aligned_layout = || {
+                fail_point!("file_pipe_log::open::force_aligned_layout", |_| { true });
+                false
+            };
+            let force_set_fs_block_size = || {
```
tabokie, line 144: Why not merge these two? I see you always use them together.

src/file_pipe_log/reader.rs
LykxSassinator: Roger that.

Cargo.toml
```diff
@@ -35,6 +35,7 @@ fail = "0.5"
 fs2 = "0.4"
 hashbrown = "0.12"
 hex = "0.4"
+if_chain = "1.0.2"
```
tabokie, line 38:
```suggestion
if_chain = "1.0"
```

src/file_pipe_log/log_file.rs
```diff
@@ -100,6 +100,11 @@ impl<F: FileSystem> LogFileWriter<F> {
     }
 
     pub fn write(&mut self, buf: &[u8], target_size_hint: usize) -> Result<()> {
+        // @lucasliang, TODO:
```
tabokie, line 103: Remove these. It's too high level to be useful.

src/file_pipe_log/pipe_builder.rs
```diff
@@ -300,7 +306,7 @@ impl<F: FileSystem> DualPipesBuilder<F> {
                     let is_last_file = index == chunk_count - 1 && i == file_count - 1;
                     match build_file_reader(file_system.as_ref(), f.handle.clone(), None) {
                         Err(e) => {
-                            let is_local_tail = f.handle.file_size()? <= LogFileFormat::len();
+                            let is_local_tail = f.handle.file_size()? <= default_format_len;
```
tabokie, line 309: This is problematic. Opening with V2 and reading corrupted V1 file would not raise an error.

I think you can modify `build_file_reader` to return a `Result<Option<Reader>>`, a return value of `Ok(None)` means we hit EOF.

LykxSassinator: I prefer to use a special `Error::InvalidArguement` to mark this error, to avoid making batch code revisions on the relevant calling paths.

tabokie: Add a test.

LykxSassinator: Yes, it already been supplied in the previous commit, where test with `fail_points` on corrupted `DataLayout` has passed. It can be reviewed in `failpoints/test_engine.rs::test_tail_corruption`.

I'm still confused that why the code path on `Err(Error::InvalidArguement(...))` is not covered, but the code path on examing `Error::InvalidArguement` in `reader.rs` is covered. Also, the print on it also could prove that it already makes sense: 

``` Rust
// in file_pipe_log/format.rs
} else {
        // Here, we mark this special err, that is, corrupted `payload`,
        // with InvalidArgument.
        println!("invalid dataload in the header, len: {}, expected len: {}",
                buf_len - Self::header_len(), Self::payload_len(version));
        Err(Error::InvalidArgument(format!(
            "invalid dataload in the header, len: {}, expected len: {}",
            buf_len - Self::header_len(), Self::payload_len(version)
        )))
    }
```
And the `failpoints/test_engine.rs::test_tail_corruption` get the following prints with expectation:
> ``` invalid dataload in the header, len: 7, expected len: 8```

tabokie: Covering a specific line of code is different than covering a code path (the LOC you added could be encountered in many different situations, and we want to cover them all). What I want is to add a test that checks "Opening with V2 and reading corrupted V1 file would raise an error".

LykxSassinator: The relevant test cases have been supplemented. And the previous bug, checking log file with corrupted V1 header with `file_size() <= header_len() + payload_len()`, have been corrected by reverting the changes.

tests/failpoints/test_engine.rs
```diff
@@ -450,7 +450,75 @@ fn test_tail_corruption() {
...
+    // Version::V1 in header owns abnormal DataLayout.
+    {
+        let _f = FailGuard::new("log_file_header::force_abnormal_data_layout", "return");
+        let dir = tempfile::Builder::new()
+            .prefix("test_tail_corruption_4")
+            .tempdir()
+            .unwrap();
+        let cfg_err = Config {
+            dir: dir.path().to_str().unwrap().to_owned(),
+            recovery_mode: RecoveryMode::AbsoluteConsistency,
```
tabokie, line 464: Shouldn't use this mode. The test checks whether the code correctly detects tail corruption, by entering a condition of tail corruption (and engine opens successfully), and entering a condition of non tail corruption (and engine fails to open). But when you set to `AbsoluteConsistency` mode, engine will always fail in both cases, so it's hard to reason if the engine is behaving correctly.

src/pipe_log.rs
```diff
@@ -114,25 +115,55 @@ impl Default for Version {
...
+
+impl DataLayout {
+    pub fn from_u64(val: u64) -> Self {
+        match val {
+            0 => DataLayout::NoAlignment,
+            aligned => DataLayout::Alignment(aligned),
+        }
+    }
+
+    pub fn to_u64(self) -> u64 {
```
tabokie, line 136: debug_assert!(aligned > 0)

src/file_pipe_log/pipe_builder.rs
```diff
@@ -300,7 +303,12 @@ impl<F: FileSystem> DualPipesBuilder<F> {
                     let is_last_file = index == chunk_count - 1 && i == file_count - 1;
                     match build_file_reader(file_system.as_ref(), f.handle.clone(), None) {
                         Err(e) => {
-                            let is_local_tail = f.handle.file_size()? <= LogFileFormat::len();
+                            let mut is_local_tail = f.handle.file_size()? <= LogFileFormat::header_len();
```
tabokie, line 306: No need to use file_size anymore.

tabokie (commented):
Rest LG

tabokie (approved):

PR #251, opened 2022-07-25
fix append recovery bug

### LINKED ISSUE #250: overwrite panic during recovery
Panic message: `attempt to overwrite compacted entries in xx`.

This panic should be elided during parallel recovery. Consider this case:

file A: [index=1][index=2]
file B: [index=3][index=2][index=3]

It would raise a false panic if we recovery file B alone and try to merge it with file A afterwards.

Workaround:
1. Disable parallel recovery by setting `recovery_threads` to 1
2. Repair the log files with this script:

```rhai
fn filter_append(id, first, count, rewrite_count, queue, ifirst, ilast) {
  if id == rid && first + count - 1 >= ifirst {
    return 2;  // discard existing
  }
  0
}
fn filter_compact(id, first, count, rewrite_count, queue, compact_to) {
  0
}
fn filter_clean(id, first, count, rewrite_count, queue) {
  0
}
```

tabokie:
Close #250 

### DIFF
--- a/src/memtable.rs
+++ b/src/memtable.rs
@@ -408,28 +408,25 @@ impl<A: AllocatorTrait> MemTable<A> {
         }
     }
 
-    /// Appends some entries from rewrite queue. Assumes this table has no
-    /// append data.
+    /// Appends some entries from append queue. Assumes this table has no
+    /// rewrite data.
     ///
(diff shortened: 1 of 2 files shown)


Connor1996 (approved):
LGTM

PR #255, opened 2022-08-02
Unify the strategy of tolerating tail corruption when Version == V2.

LykxSassinator:
Supplemented bugfix for #249 , to unify the strategy on whether tolerating the tail corruption when `version == V2`.


tabokie:
Closing this, I'll fix it in #244

tabokie:
Actually, this is not possible anymore with file recycle. Because file_size is already larger than header_size even before the header is written.

Issue #258, opened 2022-08-10
Support multiple WALs


tabokie:
Write to multiple WALs concurrently to leverage hardware parallelism.

Raft Engine provides a `WriteToken` that maps to a WAL stream underneath. If `WriteToken` is not provided, it will write to the default WAL stream.

User guarantees there will be no concurrent writing to the same region using different tokens. But writes to a region are allowed to be sent to different WALs at different time.

There're two ways to establish total order between logs among muliple log streams,

(1) Raft Engine have awareness of Raft term of the log entries. Key-values use the term of latest log in the region.

(2) Raft Engine internally manages a sequence number for each region.

PR #260, opened 2022-08-13
Extend `LogQueue` to support multiple append channels

tabokie:
Ref #258

- Add `LogQueue(u8)` to reference different write channels
- Introduce `FilesView` to filter log entries older than a specific timepoint (see comments for `history_files_view`)
- Use `LogKind` in places where channel details are irrelevant

Most of the code is doing basically the same thing but renamed.


tabokie:
Let me finish this feature on my local branch first.

Issue #257, opened 2022-08-05
Support secondary log directory


tabokie:
It can be useful to configure raft engine on a separate (smaller) disk. And when that disk is full, we should provide a way for raft-engine to locate newly created log files on the main disk volume.

LykxSassinator:
U mean that we need a new configuration such as `sub-dir` or `assist-dir` to store logs on different disk?

With this pre-requisite, the main dir, that is, the current configuration `dir`, is still used as the default main directory to store logs. Only if the main `dir` is ***FULL*** or not accessible, we will use the `assist-dir` or `sub-dir` to store the incoming logs.

Meanwhile, about the abnormal circumstance:
* `sub-dir` is `None` or belongs to the same volumn of disk, if the disk is full, the newly file creation should fail.

PR #261, opened 2022-08-15
Support secondary directory for log writing.

LykxSassinator:
## Brief Introduction
This pr is used to support the `secondary directory` feature, referred to ISSUE: #257.
It contains:
* Add an extra configuration `sub-dir` to enable this feature. Default is `None`, and if set with `Some(...)`, this directory will be used if the main dir is full, specified by `dir` option.
* Add an extra class, named as `StorageInfo` in `pipe.rs`, to make this feature compatible to other existing features, i.e. recycle logs, recoverying and so on.


LykxSassinator:
It's a prototype for closing #257, and all reasonable suggestions are acceptable.

LykxSassinator:
It's a prototype for closing #257, and all reasonable suggestions are acceptable.

LykxSassinator:
@tabokie Please take a review, this is a prototype for implementing `secondary dir`. If this design is a bit confusing, we can briefly make a offline discussion later.

LykxSassinator:
Please hold on, the design of it should be refined and polished, to simplify the whole structure.

LykxSassinator:
Please hold on, the design of it should be refined and polished, to simplify the whole structure.

src/engine.rs
```diff
@@ -191,7 +191,18 @@ where
             debug_assert_eq!(writer.perf_context_diff.write_wait_duration, Duration::ZERO);
             perf_context += &writer.perf_context_diff;
             set_perf_context(perf_context);
-            writer.finish()?
+            // Retry if `writer.finish()` returns a special err, remarking there still
+            // exists free space for this `LogBatch`.
+            let ret = writer.finish();
```
tabokie, line 196: This is not enough. fsync failure will panic. You should let leader do the retry.

tabokie:
I think letting leader do all the retrying is simpler. When a write fails with NOSPC, the leader will close the current active file (it internally truncates un-sync-ed parts), and create a new file and retry the whole write group.

LykxSassinator:
I hold a different view on it.
As I do not expect the current writing group hanging on one LogBatch writing because of `NOSPC` err, I think the current leader should only be responsible for successfully written `LogBatch`s, and the followers, failed with `NOSPC`, should put them LogBatch into next writing group and wait for the next leader to do the `append` ops. And that's the core reason why I implemented the `retry` strategy.

LykxSassinator:
I hold a different view on it.
As I do not expect the current writing group hanging on one LogBatch writing because of `NOSPC` err, I think the current leader should only be responsible for successfully written `LogBatch`s, and the followers, failed with `NOSPC`, should put them LogBatch into next writing group and wait for the next leader to do the `append` ops. And that's the core reason why I implemented the `retry` strategy.

src/engine.rs
```diff
@@ -164,12 +174,35 @@ where
...
+                if force_rotate {
+                    if let Err(e) = self.pipe_log.rotate(LogQueue::Append) {
+                        panic!(
+                            "Cannot rotate {:?} queue due to IO error: {}",
+                            LogQueue::Append,
+                            e
+                        );
+                    }
+                }
                 if let Err(e) = self.pipe_log.maybe_sync(LogQueue::Append, sync) {
```
tabokie, line 206: It's still incorrect. The NOSPC error would happen here. You need to check its error as well.

src/engine.rs
```diff
@@ -191,7 +224,17 @@ where
...
-            writer.finish()?
+            // Retry if `writer.finish()` returns a special err, remarking there still
+            // exists free space for this `LogBatch`.
+            let ret = writer.finish();
+            if let Err(Error::Other(_)) = ret {
+                log_batch.reset_to_encoded_state();
+                // Here, we will retry this LogBatch `append` by appending this writer
+                // to the next write group, and the current write leader will not hang
+                // on this write and will return timely.
+                return self.write(log_batch, sync);
```
tabokie, line 235: This will probably enter infinite recursive calls if the disk is still full after rotate.

src/engine.rs
```diff
@@ -148,9 +148,19 @@ where
                 let now = Instant::now();
                 let _t = StopWatch::new_with(&*ENGINE_WRITE_LEADER_DURATION_HISTOGRAM, now);
                 let file_context = self.pipe_log.fetch_active_file(LogQueue::Append);
+                // Flag on whether force to flush the current active file or not.
+                let mut force_rotate = false;
```
tabokie, line 152: Two things here, (1) I don't want the engine to care about "rotate" details. We can at most implement a `fn retriable` for the error which is true for disk full (2) We need to assume all buffered data is lost if `fsync` fails. To tolerate disk full we need a way to detect "if there's any buffered write from last write group". If so we must panic regardless of whether it's a disk full error or not.

I haven't thought of an elegant abstraction to handle both of these cases, maybe we can rethink it next week.

src/file_pipe_log/pipe_builder.rs
```diff
@@ -99,108 +101,201 @@ impl<F: FileSystem> DualPipesBuilder<F> {
     /// Scans for all log files under the working directory. The directory will
     /// be created if not exists.
     pub fn scan(&mut self) -> Result<()> {
```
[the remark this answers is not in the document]
LykxSassinator: The modifications are not change, just spliting the procedure of `scan` into three parts. It's not approapriate to code it with two `fs::read_dir(path)?.for_each`s, as the definition of `cfg.secondary-dir` is `Option<String>`.

src/engine.rs
LykxSassinator: Just like your concerns in the following comments `Two things here ...`, I still wonder whether there exists other elegant ways to handle it.

src/engine.rs
LykxSassinator: About the second suggestion u mentioned above, I tried to figure out whether there exists an elegant solution to this corner case recently, but ended with failure. **I think it's hard to detect "if there's any buffered write from last write group".**

And I hold the view that if `fsync` is failed, there should be some hardware errs raised by IO device.
Refer to **https://wiki.postgresql.org/wiki/Fsync_Errors**, I think we also can take the same processing if `fsync` returns a unexpected error, **`panic`** is enough for it.

So, current `maybe_sync` is enought for `nospc` err:
- If `rotate` fails, it means:
  - There might have no free space or no stale files to reuse, should `panic`.
  - `fsync` is failed, as the previous annotations show, should `panic`.
- If `fsync` fails, just `panic`.

LykxSassinator:
PTAL cc @tabokie , several modifications have been supplied according to previous suggestions.

src/engine.rs
tabokie: The point is `fsync` can fail even if `pwrite` succeeds. But I agree in a practical sense this is unlikely to happen.

src/engine.rs
tabokie: This issue is not addressed.

src/log_batch.rs
```diff
@@ -555,6 +555,14 @@ enum BufState {
     /// # Invariants
     /// LOG_BATCH_HEADER_LEN <= buf.len()
     Sealed(usize, usize),
```
tabokie, line 557: Let's not add yet another state. Change `Sealed` to contain a third field "original checksum". During `prepare_write`, if the state is `Sealed`, we can use this field to re-sign checksum.

src/log_batch.rs
```diff
@@ -779,14 +790,17 @@ impl LogBatch {
...
+    /// Makes preparations for the write of `LogBatch`.
     #[inline]
     pub(crate) fn prepare_write(&mut self, file_context: &LogFileContext) -> Result<()> {
         match self.buf_state {
             BufState::Encoded(header_offset, entries_len) => {
                 LogItemBatch::prepare_write(&mut self.buf, file_context)?;
                 self.buf_state = BufState::Sealed(header_offset, entries_len);
             }
+            BufState::ReSealing(_, _) => {
+                LogItemBatch::prepare_write(&mut self.buf, file_context)?;
```
tabokie, line 802: This doesn't make sense, the signature will be `checksum^first_file^second_file`. I wonder why the tests didn't catch it.

src/file_pipe_log/pipe_builder.rs
```diff
@@ -99,108 +101,203 @@ impl<F: FileSystem> DualPipesBuilder<F> {
...
+            if let Some(secondary_dir) = self.cfg.secondary_dir.as_ref();
+            if let Ok(0) = parse_file_range(
+                    secondary_dir,
+                    StorageDirType::Secondary,
+                    &mut append_id_range,
+                    &mut rewrite_id_range,
+                    &mut append_file_dict,
+                    &mut rewrite_file_dict,
+                );
+            if let Ok(true) = from_same_dev(&self.cfg.dir, secondary_dir);
```
tabokie, line 238: This is wrong, user would've migrated the directory from another disk. Let's check the device during config validation instead.

src/log_batch.rs
LykxSassinator: `ReSealing` is a special state to handle the recursive `write` problem, remarking that the LogBatch has entered the `rewrite` state. 

With this special state, the log_batch will follow the state transformating path `Write(Open -> Encoded -> Sealed)` => `Rewrite(Sealed -> Resealing -> end with panic || end with successful "sync")`

So, if `rewrite` failed,  the next `return self.write(log_batch, sync)` would `panic` in the next checking of `log_batch.prepare_rewrite()`.

src/file_pipe_log/pipe_builder.rs
LykxSassinator: But I prefer doing dir checking here.

Here, only when "the `secondary-dir` had no valid log files" and "`from_same_dev` returned `true`", we could ignore the `secondary-dir`. And whether the migrated directory owns valid files has been checked by `if let Ok(0) = parse_file_range(...)`.

If we made dir checking during the config validation, we would not know _whether there exists valid log files in the `secondary-dir` when `from_same_dev(...)` returns `true`_. In this implementation,  we should check it again when scanning the two dir and make the decision whether we could ignore the `secondary-dir`.

tabokie: So what's the point of doing the check at all? We already allow a secondary directory on the same device in the case of "user migration". The way I see it, you either deny it or accept it, there shouldn't be any middle ground.

src/log_batch.rs
tabokie: It's too obscure to use this as a recursion limit. Using a loop to retry is more explicit and understandable.

src/file_pipe_log/pipe_builder.rs
LykxSassinator: No, we don't allow user use a `secondary-dir` on the same device, the strategy here is to check and ignore this case, by `reset cfg.secondary-dir = None` and logging with warning `"sub-dir ({}) and dir ({}) are on same device, ignore it"`.

Maybe I should add an extra specification to `secondary-dir`, which is only valid when it's on different device?

tabokie: > Here, only when "the secondary-dir had no valid log files" and "from_same_dev returned true", we could ignore the secondary-dir.

You said it before, this case is tolerated, i.e. the secondary directory on the same device can still be used.

LykxSassinator: Emmm.....more precisely, I expect and strongly recommend that users should use different dir on different devices respectively.

However, as u mentioned above, some users might migrate the `secondary-dir` from another device to the device, `dir` belonging to. And in this case, I think we should tolerate it, considering data integrity. And that's why I introduce this check here.

tabokie: My point is, if the code can still work (and is expected to work) if the secondary-dir is located on the same device, then there's no need to impose a restriction. By banning this case, you are in fact making it more difficult to test. That's why I said "you either deny it or accept it, there shouldn't be any middle ground".

Of course, you can still emit a warning (preferably during config validation, you can call `read_dir` there to detect if the directory is empty).

LykxSassinator:
After merging #269, activate this pr again.

Ping @tabokie .

src/file_pipe_log/pipe_builder.rs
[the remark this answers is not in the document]
tabokie: The change is still too big and I don't see a clear purpose. You can keep the `scan_dir` simple and short as before, and do the file handle initialization early, then sort the `FileToRecover` list afterwards.

src/file_pipe_log/pipe.rs
```diff
@@ -22,10 +23,91 @@ use crate::{perf_context, Error, Result};
...
+#[repr(u8)]
+#[derive(Clone, Copy, PartialEq, Eq, Debug, EnumIter)]
+pub enum StorageDirType {
+    Main = 0,
+    Secondary = 1,
+}
+
+/// Represents the info of storage dirs, including `main dir` and
+/// `secondary dir`.
+struct StorageInfo {
```
tabokie, line 35: StorageManager, and use path_id to refer to path instead of storage_type.

src/file_pipe_log/pipe.rs
```diff
@@ -242,9 +349,7 @@ impl<F: FileSystem> SinglePipe<F> {
     /// Synchronizes all metadatas associated with the working directory to the
     /// filesystem.
     fn sync_dir(&self) -> Result<()> {
```
tabokie, line 351: This is a heavy operation. Let's be specific about which directory to sync.

src/engine.rs
```diff
@@ -139,51 +139,111 @@ where
         let start = Instant::now();
         let len = log_batch.finish_populate(self.cfg.batch_compression_threshold.0 as usize)?;
         let block_handle = {
```
tabokie, line 141: How about
```rust
let block_handle = loop {
  attempt += 1;
  let res = do_some_writes();
  if res.is_ok() || attempt >= MAX_RETRY {
    break res;
  }
}
```

src/engine.rs
```diff
@@ -139,51 +139,111 @@ where
...
+                            ))));
+                            continue;
+                        }
+                        writer.set_output(res);
+                    }
+                    perf_context!(log_write_duration).observe_since(now);
+                    if force_rotate {
+                        // If the leader failed to `rotate` a new log for the un-synced LogBatches,
+                        // it means that it encounters unexpected errors, and should panic.
+                        if let Err(e) = self.pipe_log.rotate(LogQueue::Append) {
```
tabokie, line 205: Put the rotate inside `append`, which is the whole point of #269.

src/file_pipe_log/pipe_builder.rs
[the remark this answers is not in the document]
LykxSassinator: On my view, I think it's more clear than before.

`scan` has been splited into the following steps:
1. scanning dir to get file_seq_range of append logs and rewrite logs
  + `scan_dir` in **Main dir**;
  + `scan_dir` in **Secondary dir**, if it had been specified by `cfg.secondary-dir`;
2. Check and clear stale metadata of logs
  + `clear_stale_metadata` in **Main dir**;
  + `clear_stale_metadata` in **Secondary dir**, if it had been specified by `cfg.secondary-dir`;
3. Build file_list vector;

Still confusing?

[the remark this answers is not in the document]
tabokie: There're two things here, one is to keep diff as short as possible. I do refactor all the time, but they are often centered around a change of abstraction (i.e. interaction between different modules, usually involves changing function interface and data types). I don't often refactor the internal implementation unless it significantly improves readability.

Then let's go to the code itself, is the readability improved? IMO it is not. The LOC increased 100%. And you introduced several types for the sake of refactoring (adding glue types usually indicates a bad abstraction). In particular I don't think the `min_id`/`max_id` approach is suitable anymore.

src/file_pipe_log/pipe.rs
tabokie: path_id as in an integer. You can make the StorageManager manages a vector of directories, and you can also let it take care of directory lock as well. (Pass it to `Pipe::open`).

src/file_pipe_log/pipe_builder.rs
```diff
@@ -90,31 +91,125 @@ impl<F: FileSystem> DualPipesBuilder<F> {
...
+    /// Scans for all log files under the working directory.
     pub fn scan(&mut self) -> Result<()> {
-        let dir = &self.cfg.dir;
-        let path = Path::new(dir);
-        if !path.exists() {
-            info!("Create raft log directory: {}", dir);
-            fs::create_dir(dir)?;
-            self.dir_lock = Some(lock_dir(dir)?);
-            return Ok(());
+        // Scan main `dir` and `secondary-dir`, if `secondary-dir` is valid.
```
tabokie, line 102: Let me be more verbose this time:
```rust
struct PipeBuilder {
  // Only available after a successful `scan`.
  dir_manager: Option<Arc<DirectoryManager>>,
}
struct DirectoryManager {
  paths: Vec<PathBuf>,
  locks: Vec<File>,
}

fn scan(&self) {
  // setup directories 
  let mut dirs = DirectoryManager::new();
  dirs.add(self.cfg.dir)?;
  dirs.add(self.second_dir)?;
  for path_id in 0..dirs.len() {
    self.scan_dir(path_id);
  }
  self.rewrite_files.sort();
  self.append_files.sort();
  for queue, files in [] {
    if files.is_empty() { continue; }
    // check consecutiveness
    let mut invalid_files = 0;
    let mut current_seq = files[0].seq;
    for (i, f) in files.enumerate() {
      if f.seq > current_seq {
        warn!("hole");
        current_seq = f.seq + 1;
        invalid_files = i;
      } else if {
        f.seq < current_seq;
        return Error::InvalidArgument("Duplicate file");
      }
    }
    files.drain(..invalid_files);
    if files.is_empty() { continue; }
    // cleanup metadata
    let delete_start = {...}
    'cleanup: for seq in delete_start..files[0].seq {
      for path_id in 0..dirs.len() {
        if self.file_system.exists_metadata(dirs.path(path_id)) {
          if let Err(e) = self.file_system.delete_metadata() { ...; break 'cleanup; }
        }
      }
    }
  }

  self.dir_manager = Arc::new(dirs);
}

fn scan_dir(&self) {
  for f in fs::read_dir() {
    self.rewrite_files.push(...);
  }
}

fn build(self) {
  SinglePipe::new(self.dir_manager.clone(), ...);
}
```

src/util.rs
```diff
@@ -325,6 +325,56 @@ pub trait Factory<Target>: Send + Sync {
...
+    /// Check the given `dir` is valid or not. If `dir` not exists,
+    /// it would be created automatically.
+    pub(crate) fn validate_dir(dir: &str) -> IoResult<()> {
+        let path = Path::new(dir);
+        if !path.exists() {
+            std::fs::create_dir(dir)?;
+            return Ok(());
+        }
+        if !path.is_dir() {
+            return Err(std::io::Error::new(
```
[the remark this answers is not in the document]
LykxSassinator: Not returned with a std::io::Error? Or u means that the `ErrorKind` is not suitable ?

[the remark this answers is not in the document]
tabokie: The former. Actually I think it makes more sense to move this function to config.rs.

LykxSassinator:
This pr is based on a too obsolete `master` branch, which needs to be rebase.

So, I've reviewed the previous comment and re-built another pr for tackling this issue.

tabokie (commented):
(Some pending comments from months ago)

src/log_batch.rs
```diff
@@ -551,10 +529,10 @@ enum BufState {
     /// state only briefly exists between encoding and writing, user operation
     /// will panic under this state.
     /// # Content
-    /// (header_offset, entries_len)
+    /// (header_offset, entries_len, signature)
```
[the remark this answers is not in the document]
LykxSassinator: This change on `BufState::Sealed()` has been refactored in the new PR.

src/log_batch.rs
```diff
@@ -779,13 +760,35 @@ impl LogBatch {
...
+                self.buf_state = BufState::Sealed(header_offset, entries_len, new_signature);
+            }
+            BufState::Sealed(header_offset, entries_len, old_signature) => {
+                // Re-seal the LogBatch.
+                match (old_signature, new_signature) {
+                    (Some(old), Some(new)) => {
+                        sign_checksum(&mut self.buf, Some(old ^ new))?;
+                    }
+                    (Some(old), None) => {
+                        sign_checksum(&mut self.buf, Some(old))?;
```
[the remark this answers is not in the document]
LykxSassinator: Same as above.

src/config.rs
```diff
@@ -183,10 +214,31 @@ impl Config {
...
+fn validate_dir(dir: &str) -> Result<()> {
+    use std::path::Path;
+
+    let path = Path::new(dir);
+    if !path.exists() {
+        std::fs::create_dir(dir)?;
+        return Ok(());
+    }
+    if !path.is_dir() {
+        return Err(box_err!("Not directory: {}", dir));
```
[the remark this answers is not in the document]
LykxSassinator: Same as above.

src/engine.rs
```diff
@@ -154,14 +163,16 @@ where
...
                     let log_batch = writer.mut_payload();
                     let res = self.pipe_log.append(LogQueue::Append, log_batch);
                     writer.set_output(res);
                 }
                 perf_context!(log_write_duration).observe_since(now);
                 if sync {
-                    // As per trait protocol, this error should be retriable. But we panic anyway to
-                    // save the trouble of propagating it to other group members.
+                    // As per trait protocol, this error should be retriable. But we panic
+                    // anyway to save the trouble of propagating it to
```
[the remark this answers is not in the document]
LykxSassinator: Tackled in the new pr.

src/engine.rs
```diff
@@ -178,7 +189,27 @@ where
             debug_assert_eq!(writer.perf_context_diff.write_wait_duration, Duration::ZERO);
             perf_context += &writer.perf_context_diff;
             set_perf_context(perf_context);
-            writer.finish()?
+            // Retry if `writer.finish()` returns a special 'Error::Other', remarking that
```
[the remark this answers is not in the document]
LykxSassinator: Done.

PR #264, opened 2022-08-18
clean up file collection

tabokie:
* Putting mutable operations into methods
* Clean up tests
* Log recycling doesn't affect `purge_to` return value any more
* Disallow reading stale files via pipe_log
* Enable recycling for tests

### DIFF
--- a/src/file_pipe_log/pipe.rs
+++ b/src/file_pipe_log/pipe.rs
@@ -32,37 +32,79 @@ struct FileCollection<F: FileSystem> {
     /// Sequence number of the first file that is in use.
     first_seq_in_use: FileSeq,
     fds: VecDeque<FileWithFormat<F>>,
-    /// `0` => no capbility for recycling stale files
-    /// `_` => finite volume for recycling stale files
+    /// A hint to control the amount of stale files.
+    /// Unless `active_len() > capacity`, `len() <= capacity`.
     capacity: usize,
 }
 
+#[derive(PartialEq, Eq, Debug)]
+struct FileState {
+    first_seq: FileSeq,
+    first_seq_in_use: FileSeq,
+    total_len: usize,
+}
+
+/// Note: create a method for any mutable operations.
 impl<F: FileSystem> FileCollection<F> {
-    /// Recycles the first obsolete file and renewed with new FileId.
-    ///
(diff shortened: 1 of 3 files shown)


src/file_pipe_log/pipe.rs
```diff
@@ -814,79 +773,50 @@ mod tests {
...
-        pipe_log.purge_to(FileId { queue, seq: 3 }).unwrap();
-        assert_eq!(pipe_log.file_span(queue), (3, 3));
-
-        // fetch active file
-        let file_context = pipe_log.fetch_active_file(LogQueue::Append);
-        assert_eq!(file_context.version, Version::V2);
-        assert_eq!(file_context.id.seq, 3);
+            assert_eq!(reader.read(handle).unwrap(), content);
+            // Delete file so that it cannot be reused.
+            fs.delete(handle.id.build_file_path(path)).unwrap();
```
LykxSassinator, line 808: The `deletion` is redundant here, I think.
The implementation of `reuse` in ObfuscatedFileSystem has already do `delete(src)` and `create(dst)`. And I think this case should testing the following cases:
* Builds `pipe` with `DefaultFileSystem` for recycling, validates that `reuse` is just `rename` a stale file for `old_filepath` to `new_filepath`.
* Builds `pipe` with `ObfuscatedFileSystem` for recycling, validates that `reuse` will delete the old file and create a new file. The new file is empty.

LykxSassinator (commented):
Rest is LGTM.

src/file_pipe_log/pipe.rs
tabokie: It's not redundant. It is to make the next `reuse` fails.

There's no need to test the behavior of a specific file system because it can have arbitrary user implementation.

tabokie:
Ping @Connor1996

src/file_pipe_log/pipe.rs
```diff
@@ -239,12 +281,23 @@ impl<F: FileSystem> SinglePipe<F> {
         };
         let path = file_id.build_file_path(&self.dir);
         let fd = {
```
Connor1996, line 283: let fd = Arc::new({
    ...
});

Connor1996 (approved):
LGTM

PR #265, opened 2022-08-29
Default enable log recycling

### LINKED ISSUE #263: Default enable log recycling.
## Description
According to PR: #224 , we have make RaftEngine support log recycling with configuration `enable-log-recycle: true`. Also, the testing results shown in this pr have proved that this feature have positive effects to `write` workloads.
So, starting from `0.3` version, we want to make it `open` as default.

## Related changes
Default values to `format-version` and `enable-log-recycle` should be modified respectively:
* Default value of `format-version` should be changed to `2` (origin is `1`);
* Default value of `enable-log-recycle` should be changed to `true` (origin is `false`);

LykxSassinator:
## Description
Default enable log recycling. Close #263.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,9 +2,6 @@
 
 ## [Unreleased]
 
-### Behavior Changes
-* Enable log recycling by default.
-
 ### Bug Fixes
 
 * Unconditionally tolerate `fallocate` failures as a fix to its portability issue. Errors other than `EOPNOTSUPP` will still emit a warning.
@@ -21,6 +18,10 @@
 * Add metadata deletion capability to `FileSystem` trait. Users can implement `exists_metadata` and `delete_metadata` to clean up obsolete metadata from older versions of Raft Engine.
 * Add `Engine::scan_messages` and `Engine::scan_raw_messages` for iterating over written key-values. 
 * Add `Engine::get` for getting raw value.
+
+### Behavior Changes
+
+* Enable log recycling by default.
 
 ## [0.2.0] - 2022-05-25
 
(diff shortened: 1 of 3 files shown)


tests/failpoints/test_engine.rs
```diff
@@ -611,6 +611,7 @@ fn test_recycle_with_stale_logbatch_at_tail() {
         target_file_size: ReadableSize::kb(2),
         purge_threshold: ReadableSize::kb(4),
         enable_log_recycle: true,
+        format_version: Version::V1,
```
tabokie, line 614: Why?

LykxSassinator: Here, default Version has been updated to V2 to enable log recycling defaultly, this case should be compatible to this modification.

The original design of this case is to mock a case that RaftEngine does recycling on `V1` log files, and it cannot recognize the recycled part in the recycled file, as `V1` logs has no signatures.

tabokie (approved):

tabokie (commented):
Update the changelog under "Behavior changes".

CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## [Unreleased]
 
+### Behavior Changes
+* Enable log recycling by default.
```
tabokie, line 6: And put it at the end.
```suggestion
### Behavior Changes

* Enable log recycling by default.
```

tabokie: @LykxSassinator I updated the comment.

LykxSassinator: How about => 

### Bahavior Changes

* Change format version to 2 from 1 by default.
* Enable log recycling by default.

tabokie: Sure.

tabokie (approved):

PR #267, opened 2022-09-01
fix panic when reusing a was-empty batch

### LINKED ISSUE #266: Writing an empty batch causes panic
panic message reads: `assertion failed: self.buf_state == BufState::Open`

tabokie:
Close #266

An empty batch went through `finish_populate` but not `drain` (in memtable apply block). This will break internal state machine and panic.

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -1397,6 +1397,34 @@ mod tests {
         );
     }
 
+    #[ignore]
+    #[test]
+    fn test_empty_batch() {
+        let dir = tempfile::Builder::new()
+            .prefix("test_empty_batch")
+            .tempdir()
+            .unwrap();
+        let cfg = Config {
+            dir: dir.path().to_str().unwrap().to_owned(),
+            ..Default::default()
+        };
+        let engine =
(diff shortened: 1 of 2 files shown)


src/engine.rs
```diff
@@ -136,8 +136,13 @@ where
     /// bytes. If `sync` is true, the write will be followed by a call to
     /// `fdatasync` on the log file.
     pub fn write(&self, log_batch: &mut LogBatch, mut sync: bool) -> Result<usize> {
+        if log_batch.is_empty() {
+            return Ok(0);
+        }
         let start = Instant::now();
         let len = log_batch.finish_populate(self.cfg.batch_compression_threshold.0 as usize)?;
+        debug_assert!(!log_batch.is_empty());
```
LykxSassinator, line 144: Do it before `finish_populate` or remove it. The first validation `log_batch.is_empty()` has guaranteed that `log_batch.is_empty() == false`.

LykxSassinator (commented):
Rest LGTM

src/engine.rs
tabokie, line 144:
```suggestion
```

Connor1996 (approved):
LGTM

PR #269, opened 2022-09-02
hide more file writing details from engine

tabokie:
This is a preparing work for #258.

Changes:
- Move `fsync` to file handle instead of writer, so that we can call it concurrently in the future. In doing so, we have to remove the sync offset tracking and deprecate `bytes_per_sync` support
- Move `prepare_write` into pipe_log. Introduce a trait `ReactiveBytes` for this purpose.
- Rotate the file during write instead of doing it every write group, the timing is also delayed to the next write after exceeding limit (instead of the write that exceeds the limit).
- Differentiate two types of I/O errors: unrecoverable error from fsync, and retriable error from pwrite. Only bubble error for the latter case, panic early for unrecoverable ones.
- Also refactor the purge code, fix two cases where force-compact is not triggered

### DIFF
--- a/src/file_pipe_log/pipe.rs
+++ b/src/file_pipe_log/pipe.rs
@@ -471,14 +471,6 @@ impl<F: FileSystem> SinglePipe<F> {
         self.flush_metrics(current.total_len);
         Ok((current.first_seq_in_use - prev.first_seq_in_use) as usize)
     }
-
-    fn fetch_active_file(&self) -> LogFileContext {
-        let files = self.files.read();
-        LogFileContext {
-            id: FileId::new(self.queue, files.first_seq + files.fds.len() as u64 - 1),
-            version: files.fds.back().unwrap().format.version,
-        }
-    }
 }
 
 /// A [`PipeLog`] implementation that stores data in filesystem.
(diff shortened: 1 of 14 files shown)


src/file_pipe_log/pipe.rs
```diff
@@ -401,15 +413,11 @@ impl<F: FileSystem> SinglePipe<F> {
         Ok(handle)
     }
 
-    fn maybe_sync(&self, force: bool) -> Result<()> {
+    fn sync(&self) -> Result<()> {
```
LykxSassinator, line 416: In current implementation of `sync`, it still gets the `mutex` for avoiding concurrent racing at the beginning.

According to #258 , will it be refactored to support concurrent accessing in later work ? i.e. replace `mutex` with `Rwlock` ? or add an extra `F::Handle` in `SinglePipe`, making it be called concurrently ?

LykxSassinator (commented):
Rest is LGTM

src/file_pipe_log/pipe.rs
tabokie: Yes.

tabokie:
Ping @Connor1996 , this one is blocking #261

Connor1996 (approved):
LGTM

PR #272, opened 2022-09-19
Disable `enable-log-recycle` by default.

LykxSassinator:
### Description
Disable `enable-log-recycle` in raft-engine by default.

When opening `log recycling`,
> We observe on physical disk, overwriting existing log files will generate unnecessary read I/Os, and even though the fdatasync latency is reduced, the duration sum of fdatasync and pwrite has increased.

So, we disable it by default.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,7 +26,7 @@
 ### Behavior Changes
 
 * Change format version to 2 from 1 by default.
-* Enable log recycling by default.
+* Disable log recycling by default.
 
 ## [0.2.0] - 2022-05-25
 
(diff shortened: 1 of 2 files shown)


LykxSassinator:
cc @tabokie

tabokie (commented):
Update changelog.

CHANGELOG.md
```diff
@@ -26,7 +26,7 @@
 ### Behavior Changes
 
 * Change format version to 2 from 1 by default.
-* Enable log recycling by default.
+* Disable log recycling by default.
```
tabokie, line 29: 0.3.0 is already released, should open an Unreleased section.

tabokie (approved):

PR #278, opened 2022-11-01
Optimize the cold startup of RaftEngine when `enable-log-recycle` == true

### LINKED ISSUE #277: Make RaftEngine with Recycle Logs effective when starting in the cold
## Description
As the title shown, we wanna make log recycling effective when starting RaftEngine with `enable-log-recycle` in the cold state.
So, for this optimization, we will introduce extra operations to make it, i.e. preparing a bunch of stale files in advance for recycle.

LykxSassinator:
## Discription
This commit optimize the startup of Engine when `enable-log-recycle` == true, by preparing a bunch of logs named with special suffix ".fakelog", so-called "Fake logs", to curtail the side effect when there is no enough stale files for log recycling.

Related Issue: close #277


LykxSassinator:
@tabokie PTAL, thx

tabokie (commented):
Mixing multiple log queues is too ugly. I'm thinking something like this:

```rust
// builder.rs
let suffix = "append.reserved";
let reserved_files = ReservedFiles::scan_from_path(fs, path, suffix)?;
if capacity > active_files.len() + reserved_files.len() {
  reserved_files.generate_some(..);
}
let pipe = SinglePipe::build(reserved_files, active_files);
// pipe.rs
fn purge() {
  for f in purged_files {
    if f.format == V2 {
      self.reserved_files.add(f.handle);
    }
  }
  self.reserved_files.truncate(self.capacity - self.active_files.len());
}
```

LykxSassinator:
@tabokie PTAL, thx

src/file_pipe_log/pipe.rs
```diff
@@ -28,7 +28,7 @@ pub struct FileWithFormat<F: FileSystem> {
     pub format: LogFileFormat,
 }
 
-struct FileCollection<F: FileSystem> {
+pub struct FileCollection<F: FileSystem> {
```
tabokie, line 31: I don't understand the need to reuse this structure.

tabokie (commented):
As I mentioned before, keeping obsolete files across restart (i.e. do not delete them when Pipe is dropped) is of higher priority. Which means all obsolete files (not just "fake" logs) will be managed by `reserved_files`. They are renamed (this detail is preferably put inside `ReservedFiles`) before retired so that we won't waste time recovering them later.

src/file_pipe_log/file_mgr.rs
```diff
@@ -0,0 +1,536 @@
...
+    pub format: LogFileFormat,
+}
+
+#[derive(Clone, PartialEq, Eq, Debug)]
+pub struct FileState {
+    pub first_seq: FileSeq,
+    pub total_len: usize,
+}
+
+pub trait FileCollectionMgr<F: FileSystem> {
```
tabokie, line 31: There's no need for trait unless there's generic code.

src/file_pipe_log/file_mgr.rs
```diff
@@ -0,0 +1,536 @@
...
+        (prev.clone(), prev, VecDeque::<_>::default())
+    }
+}
+
+/// A collection of files for managing stale files.
+///
+/// Stale files contains two parts:
+/// - `Dummy` part, dummy log files named with `.raftlog.dummy` suffix.
+/// - `Expired` part, obsolete log files manually purged by callers, with
+///   `.raftlog` suffix.
```
tabokie, line 114: Why two different types? Why not always rename .raftlog to .reserved?

LykxSassinator: As explained, StaleFileCollection has two part of files:
+ dummy part => generated by engine when starting in cold.
+ expired part => when outer user periodically calls `purge_expired_log`, typically called by TiKV, the expired files will be appended into StaleFileCollection.

And for the problems:
1. why we do not `rename` these expired file when purging?
Here, I just wanna delay the `rename` when `Engine.drop()` to avoid the extra timecost on renaming expired files into dummy files.

2. when and how do we convert expired part into dummy part ?
As mentioned above, the converting progress is delayed to `Drop()`. That is, when `Engine.drop()`,  `LogQueue::Append` will convert all remained `expired` logs into `dummy` logs, with `.raftlog.dummy` suffix.

tabokie: > to avoid the extra timecost on renaming expired files into dummy files.

There is no need. The purge is already done in background.

LykxSassinator:
@tabokie PTAL, thx.

src/file_pipe_log/file_mgr.rs
```diff
@@ -0,0 +1,499 @@
...
+    Main = 0,
+}
+
+pub type Dirs = [String; 1];
+
+#[derive(Debug)]
+pub struct FileWithFormat<F: FileSystem> {
+    pub handle: Arc<F::Handle>,
+    pub format: LogFileFormat,
+    pub dir: DirType,
```
tabokie, line 34: pub type PathId = usize;

tabokie (commented):
I don't know why this change is so difficult for me to review... Maybe part of it has to do with the interactions between structures are not properly encapsulated. E.g. `prepare_dummy_logs_for_recycle` with so many arguments, `concat` that accepts files without checking if they have the proper names.

To improve, I think you can just keep thing even simpler, make `FileCollection` as simple as possible (as simple as a Vec):

```rust
pub struct FileList { seq, files };
impl FileList {
  fn append(&mut self, rhs: &mut FileList ) { ... }
  // Different than `VecDeque`, this function takes seqno as splitting point, and returns the first half.
  fn split_front(&mut self, seq: u64) -> Self {}
  // Caller must guarantee the file has the right seqno.
  fn push_back(&mut self, file) {..}
  fn pop_front(&mut self) -> Option<FileWithFormat> {...};
  fn span(&self) -> (usize, usize);
  fn len(&self) -> usize;
  fn get(&self, seq: usize) -> Option<&FileWithFormat> {...}
}
```

Then all the complicated logic can be put in-place, or wrapped into a bigger struct:

```rust
pub struct FileCollection { capacity, active_files: Mutex<FileList>, stale_files: Mutex<FileList>, fs }

impl FileCollection {
  fn rotate(&self) -> FileHandle {
    let f = self.stale_files.pop_front().or_else(|| fs.create(self.active_files.span().1 + 1));
    self.active_files.lock().push_back(f);
    f.handle
  }

  fn purge(&self, seq: usize) {
    let stale_files = self.active_files.split_front(seq);
    for f in stale_files {
      if f.version > 1 {
        self.fs.reuse(f, new_stale_file_name(self.stale_files.span().1 + 1));
        self.stale_files.lock().push_back(f);
      }
    }
  }

  fn initialize(&mut self) {
    for _ in self.stale_files.len()..capacity {
      let f = self.fs.create(new_stale_file_name(self.stale_files.span().1 + 1));
      self.stale_files.push_back(f);
    }
  }

  fn get(&self, seq: usize) -> Option<FileHandle> {
    self.active_files.lock().get(seq)
  }
}
```

LykxSassinator:
> 
> To improve, I think you can just keep thing even simpler, make `FileCollection` as simple as possible (as simple as a Vec):

Thx for your advices.
I've also take this design in previous draft, but I dropped it finally. For the reason why I didn't adopt the structure as u said, just like:
```Rust
struct FileCollection {
    stale_files: FileList<F>,
    active_files: FileList<F>,
    ...
}
```
is that:
In previous work, `SinglePipe` is the operator and manager of `FileCollection`. IMO, the Stale files and Active files, represent by `StaleFileCollection` and `ActiveFileCollection` respectively, are the minimal operating unit of `FileList`. So, each one of them should be designed as an individual object because the related operations has been encapsulated in the API of `SinglePipe`. So, from my views, `SinglePipe` is just the subject with same trait as the new `FileCollection` u mentioned above. For example, `SinglePipe.get_fd(...)` just called the `ActiveFileCollection.get_fd(...)` to return the expected `fd` to callers.

And the confusing parts, such as `prepare_dummy_logs_for_recycle`, will be polished.

tabokie:
@LykxSassinator It doesn't contradict with that I said.

> Then all the complicated logic can be put in-place, or wrapped into a bigger struct:

The code snippet I wrote refers to the way to "wrapped into a bigger struct". It can be done as "put in-place" as well.

LykxSassinator:
> @LykxSassinator It doesn't contradict with that I said.
> 
> > Then all the complicated logic can be put in-place, or wrapped into a bigger struct:
> 
> The code snippet I wrote refers to the way to "wrapped into a bigger struct". It can be done as "put in-place" as well.

Yes, i got what u meaned. I will polish the current codes.

src/engine.rs
```diff
@@ -698,10 +717,24 @@ mod tests {
         sub_dir.push("raft-engine");
         let cfg = Config {
             dir: sub_dir.to_str().unwrap().to_owned(),
+            enable_log_recycle: false,
```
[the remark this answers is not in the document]
LykxSassinator: One point, if `recycling` enabled here, `ObfuscatedFs` would cause the time cost of preparing stale files tool high. 

And I also add an extra `cfg_recycle` for testing starting Engine with `enable-log-recycle == true` here.

src/file_pipe_log/file_mgr.rs
```diff
@@ -0,0 +1,674 @@
...
+        file_id: FileId,
+        format: LogFileFormat,
+        target_file_size: usize,
+    ) -> Result<FileWithFormat<F>> {
+        let path_id = FileCollection::<F>::get_valid_path(paths, target_file_size);
+        let file_path = file_id.build_stale_file_path(&paths[path_id]);
+        let fd = Arc::new(file_system.create(&file_path)?);
+        let mut file = build_file_writer(file_system, fd.clone(), format, true)?;
+        let mut written = LogFileFormat::encoded_len(format.version);
+        let buf = vec![0; 4096];
```
tabokie, line 504: This is way too small. Try 16MB.

src/file_pipe_log/file_mgr.rs
```diff
@@ -0,0 +1,674 @@
...
+        file_system: &F,
+        paths: &Paths,
+        file_id: FileId,
+        format: LogFileFormat,
+        target_file_size: usize,
+    ) -> Result<FileWithFormat<F>> {
+        let path_id = FileCollection::<F>::get_valid_path(paths, target_file_size);
+        let file_path = file_id.build_stale_file_path(&paths[path_id]);
+        let fd = Arc::new(file_system.create(&file_path)?);
+        let mut file = build_file_writer(file_system, fd.clone(), format, true)?;
```
tabokie, line 502: Use file_system.new_writer(file_system.create) is enough.

src/file_pipe_log/file_mgr.rs
```diff
@@ -0,0 +1,674 @@
...
+        }
+        Ok(logical_purged_count)
+    }
+
+    /// Initialize current file collection by preparing several
+    /// stale files for later log recycling in advance.
+    ///
+    /// Attention, this function only makes sense when
+    /// `Config.enable-log-recycle` is true.
+    pub fn initialize(&mut self, target_file_size: usize) -> Result<()> {
```
tabokie, line 342: I don't know why this function is so long a gain. Why doesn't this work?
```rust
for _ in self.active_files.len() + self.stale_files.len() .. self.capacity {
  let f = build_stale_files(self.stale_files.span().1 + 1);
  self.stale_files.push_back(f);
}
```

LykxSassinator: For this, the whole design here follows the prerequisite --
+ `seqno` is consecutive in `stale_files` and `active_files`;
+ `stale_files.last_seq` + 1 == `active_files.first`.

So, if we start the engine on following special scenario:
```Rust
// Existed stale files: [16, ..., 21]
// Existed active files: [45, ..., 80]
```
we should calibrate the `seqno` in the existed stale files, to make all `seqno` consecutive:
```Rust
// After initializing =>
// stale files: [35, ..., 40][41, ..., 44] (this part is newly added by Initialize().)
// active files: [45, ..., 80]
```

tabokie: > stale_files.last_seq + 1 == active_files.first

No need.

src/file_pipe_log/pipe_builder.rs
```diff
@@ -398,29 +401,48 @@ impl<F: FileSystem> DualPipesBuilder<F> {
...
-        let files = match queue {
-            LogQueue::Append => &self.append_files,
-            LogQueue::Rewrite => &self.rewrite_files,
+        let (capacity, active_files, stale_files) = match queue {
+            LogQueue::Append => {
+                // Scan and get all existing stale files.
+                (
+                    self.cfg.recycle_capacity(),
+                    &self.append_files,
+                    FileCollection::scan_stale_files(self.file_system.as_ref(), &self.cfg.dir)?,
```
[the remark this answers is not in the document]
LykxSassinator: IMO, `scan_stale_files` is better to be executed later.

LykxSassinator:
ping @tabokie  PTAL, thx

src/file_pipe_log/pipe_builder.rs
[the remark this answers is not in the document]
tabokie: The code is mostly duplicate. And it makes sense to do all the scanning in function `scan`.

src/file_pipe_log/file_mgr.rs
```diff
@@ -0,0 +1,694 @@
...
+        // they can be ignored by clear them here.
+        Ok(FileWithFormat {
+            handle: fd,
+            format,
+            path_id,
+        })
+    }
+
+    /// Clear a log file by the given metadata.
+    fn clear_file(
```
tabokie, line 505: Doesn't make sense to wrap such a small function inside another function, especially when it's behavior is controlled by a boolean, which is a bad practice.

If you want to fail point the delete operation, just do it inside `FileSystem`.

src/file_pipe_log/file_mgr.rs
```diff
@@ -0,0 +1,694 @@
...
+        #[cfg(feature = "failpoints")]
+        {
+            let remove_skipped = || {
+                fail::fail_point!("file_pipe_log::remove_file_skipped", |_| true);
+                false
+            };
+            if remove_skipped() {
+                return Ok(());
+            }
+        }
```
tabokie, line 521: It is equivalent to:
```
fail_point!("name", |_| Ok(()));
```

src/file_pipe_log/file_mgr.rs
```diff
@@ -0,0 +1,694 @@
...
+        let path_id = FileCollection::<F>::get_valid_path(paths, target_file_size);
+        let file_path = file_id.build_stale_file_path(&paths[path_id]);
+        let fd = Arc::new(file_system.create(&file_path)?);
+        let mut writer = file_system.new_writer(fd.clone())?;
+        let mut written = 0_usize;
+        let buf = vec![0; std::cmp::min(LOG_STALE_FILE_BUF_SIZE, target_file_size)];
+        while written <= target_file_size {
+            writer.write_all(&buf).map_err(|e| {
+                writer
+                    .seek(SeekFrom::Start(written as u64))
```
tabokie, line 487: Why reseek?

src/file_pipe_log/file_mgr.rs
```diff
@@ -0,0 +1,694 @@
...
+            // Concurrent prepraring will bring more time consumption on racing. So, we just
+            // introduce a serial processing for preparing progress.
+            let mut stale_files = self.stale_files.write();
+            let (prepare_first_seq, prepare_stale_files_count) = (
+                stale_files.span().1 + 1,
+                stale_capacity.saturating_sub(stale_files.len()),
+            );
+            if stale_files.len() == 0 {
+                stale_files.first_seq = 1;
+            }
```
tabokie, line 387: Why? Isn't it already updated in `push_back`?

src/file_pipe_log/file_mgr.rs
```diff
@@ -0,0 +1,694 @@
...
+                    true, /* is_stale */
+                )?;
+            }
+            // If `self.capacity` is still not enough for storing files, we should check
+            // and remove redundant file from previous purged_files.
+            // Here, stale file with Version::V1 also should be marked and cleared.
+            let remains_capacity = self
+                .capacity
+                .saturating_sub(active_state.total_len + stale_files.len());
+            let redundant_list = purged_files.purge(remains_capacity);
```
tabokie, line 322: Ditto, remove use of `purge`.

src/engine.rs
```diff
@@ -1532,6 +1564,8 @@ mod tests {
                 dir: dir.path().to_str().unwrap().to_owned(),
                 target_file_size: ReadableSize(1),
                 purge_threshold: ReadableSize(1),
+                format_version: Version::V1,
+                enable_log_recycle: false,
```
tabokie, line 1568: Again, why change the config.

src/engine.rs
```diff
@@ -686,6 +693,18 @@ mod tests {
                 self.file_count(Some(LogQueue::Append)) + self.file_count(Some(LogQueue::Rewrite))
             }
         }
+
+        fn stale_file_count(&self, queue: Option<LogQueue>) -> usize {
```
tabokie, line 697: Don't expose the concept of stale. If you must, use `fs.file_count() - pipe.file_count(None)` instead.

src/engine.rs
```diff
@@ -364,6 +364,13 @@ where
...
     }
 
+    /// Returns the sequence number range of all stale log files in the specific
+    /// log queue.
+    /// For testing only.
+    pub fn stale_file_span(&self, queue: LogQueue) -> (u64, u64) {
+        self.pipe_log.stale_file_span(queue)
+    }
+
     pub fn get_used_size(&self) -> usize {
```
tabokie, line 374: Update the implementation to include stale files.

tabokie: Add a test for this behavior too while you are at it.

src/engine.rs
LykxSassinator: Because the default setting of `Config` on version is `Version::V2`, not expected V1 in `cfg_v1`.

As the annotations said, `cfg_v1` should be initialized with `Verion::V1` and `cfg_v2` should be initialized with `Version::V2` for testing.

src/file_pipe_log/file_mgr.rs
LykxSassinator: Redundant here.

src/file_pipe_log/file_mgr.rs
LykxSassinator: Redundant here.

src/file_pipe_log/file_mgr.rs
```diff
@@ -0,0 +1,584 @@
...
+            13,
+            new_file_handler(&path, FileId::new(LogQueue::Append, 13), Version::V2, false),
+        );
+        assert_eq!(active_files.span(), (12, 13));
+        // 11 12 | 13
+        let (state, files) = active_files.split_by(13);
+        assert_eq!(state.first_seq, 13);
+        assert_eq!(files.span(), (12, 12));
+        assert_eq!(
+            stale_files.append(files),
```
tabokie, line 505: Remove `fn append`. Use `push_back` for this line.

src/file_pipe_log/file_mgr.rs
```diff
@@ -0,0 +1,584 @@
...
+            }
+        }
+        let dir = Builder::new()
+            .prefix("test_file_collection")
+            .tempdir()
+            .unwrap();
+        let path = String::from(dir.path().to_str().unwrap());
+        let target_file_size = ReadableSize(1);
+        // | 12
+        let mut active_files = FileList::new(
```
tabokie, line 467: Either test `FileList`, or test `FileCollection`, Avoid testing two concepts in one case.

src/file_pipe_log/file_mgr.rs
```diff
@@ -0,0 +1,584 @@
...
+            handle: fd,
+            format,
+            path_id,
+        })
+    }
+
+    /// Returns a valid path for dumping new files.
+    ///
+    /// Default: 0 -> default index of the main directory.
+    pub fn get_valid_path(paths: &Paths, _target_file_size: usize) -> PathId {
```
tabokie, line 424: Replace it with `const DEFAULT_PATH_ID: PathId = 0`;

src/file_pipe_log/file_mgr.rs
```diff
@@ -0,0 +1,584 @@
...
+            self.stale_files.read().len(),
+        );
+        Ok(())
+    }
+
+    /// Generates a Fake log used for recycling.
+    ///
+    /// Attention, this function is only called when `Config.enable-log-recycle`
+    /// is true.
+    fn build_stale_file(
```
tabokie, line 393: Now you can make it a member function.

src/file_pipe_log/file_mgr.rs
```diff
@@ -0,0 +1,584 @@
...
+        let (active_state, mut purged_files) = self.active_files.write().split_by(file_seq);
+        if file_seq >= active_state.first_seq + active_state.total_len as FileSeq {
+            debug_assert_eq!(purged_files.len(), 0);
+            return Err(box_err!("Purge active or newer files"));
+        } else if purged_files.len() == 0 {
+            return Ok(0);
+        }
+        let logical_purged_count = purged_files.len();
+        {
+            let mut stale_files = self.stale_files.write();
```
tabokie, line 270:
```rust
let capacity = self.capacity.saturating_sub(active_state.total_len());
for _ in capacity..stale_files.len() {
  let _ = self.file_system.delete(stale_files.pop_front().unwrap());
}
while stale_files.len() < capacity && purged_files.len() > 0 {
  let f = purged_files.pop_front().unwrap;
  if f.has_log_signing() {
    self.file_system.rename(f);
    stale_files.push_back(f);
  } else {
    let _ = self.file_system.delete(f);
  }
}
```

LykxSassinator:
Ping @tabokie

src/file_pipe_log/pipe.rs
```diff
@@ -183,81 +76,59 @@ impl<F: FileSystem> SinglePipe<F> {
...
-            fds.push_back(FileWithFormat {
-                handle: fd,
-                format: LogFileFormat::new(cfg.format_version, alignment),
-            });
-            first_seq
-        } else {
-            first_seq + fds.len() as u64 - 1
+        let file_format = LogFileFormat::new(cfg.format_version, alignment);
+        // Update default file format in the file collection.
+        files.upd_file_format(file_format);
```
tabokie, line 81: (1) why not pass it in when creating the collection (2) this function name is bad

src/file_pipe_log/format.rs
```diff
@@ -17,8 +17,23 @@ const LOG_SEQ_WIDTH: usize = 16;
...
+const LOG_APPEND_STALE_COUNT_MAX: usize = 82;
+
+/// Returns the max limitation of the count of `.raftlog.stale`s.
+///
+/// In most common cases, the default capacity of `.raftlog`s
+/// will be set with `82`, decided by `cfg.purge_threshold` /
+/// `cfg.target_file_size` + 2. So, we can just set it as
+/// the max limitation of `.raftlog.stale`.
+#[inline]
+pub(crate) fn max_stale_log_count() -> usize {
```
tabokie, line 34: Why do this? You can create a warning if there're too many files, but don't prefill it halfway, it will cause jitter in halfway.

src/file_pipe_log/file_mgr.rs
```diff
@@ -0,0 +1,604 @@
...
+        Ok(FileWithFormat {
+            handle: fd,
+            format,
+            path_id,
+        })
+    }
+
+    /// Returns a valid path for dumping new files.
+    #[inline]
+    pub fn get_valid_path(paths: &Paths, _target_file_size: usize) -> PathId {
```
tabokie, line 413: remove this function. move the assertion to ctor.

src/file_pipe_log/file_mgr.rs
tabokie: You implementation would not fill the full capacity if there's a V1 purged file. (add a test for this case too).

I updated the previous comment for some edge cases, please just implement this function exactly like it. If there's a difficulty, raise it and let's discuss.

src/file_pipe_log/file_mgr.rs
LykxSassinator: IMO, It's better to keep it. 
This function is a basement for later `multi-dir` feature, which is expected to return a valid `Path` to user according to whether the free space in one `Path` is enough for the `target_file_size`.

src/file_pipe_log/pipe.rs
LykxSassinator: A compromised way to avoid too many arguments for FileCollection::new(...).

I'll use `allow(clippy::too_many_arguments)` here.

LykxSassinator:
Ping @tabokie

src/file_pipe_log/pipe_builder.rs
```diff
@@ -398,29 +438,78 @@ impl<F: FileSystem> DualPipesBuilder<F> {
...
+                    fail_point!("file_pipe_log::build_pipe::force_set_alignment", |_| {
+                        true
+                    });
+                    false
+                };
+                if force_set_aligned_layout() {
+                    alignment = 16;
+                }
+            }
+        }
```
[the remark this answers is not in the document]
LykxSassinator: About this part, why not keep the previous agreement that "codes about fail_point should not be contained in the release version"?

src/file_pipe_log/file_mgr.rs
```diff
@@ -0,0 +1,591 @@
...
+        let splitted_list =
+            if (self.first_seq..self.first_seq + self.fds.len() as u64).contains(&file_seq) {
+                let purged = file_seq.saturating_sub(self.first_seq) as usize;
+                let purged_file_list =
+                    FileList::new(self.first_seq, self.fds.drain(..purged).collect());
+                self.first_seq = file_seq;
+                purged_file_list
+            } else {
+                FileList::new(0, VecDeque::default())
+            };
```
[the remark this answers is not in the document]
LykxSassinator: It's incorrect.
As the api `split_off` shows:
> Returns a newly allocated VecDeque. self contains elements [0, at), and the returned deque contains elements [at, len).
```Rust
use std::collections::VecDeque;

let mut buf: VecDeque<_> = [1, 2, 3].into();
let buf2 = buf.split_off(1);
assert_eq!(buf, [1]);
assert_eq!(buf2, [2, 3]);
```
This function will return an inverse result => as the example shows, what we expect should be:
> After splitted, buf => [2, 3], buf2 => [1].

src/file_pipe_log/file_mgr.rs
```diff
@@ -0,0 +1,591 @@
...
+            return Ok(0);
+        }
+        let logical_purged_count = purged_files.len();
+        {
+            let mut stale_files = self.stale_files.write();
+            // If there exists several stale files out of space, contained in
+            // `self.stale_files` and `purged_files`, we should check and remove them
+            // to avoid the `size` of whole files beyond `self.capacity`.
+            let remains_capacity = self.capacity.saturating_sub(active_state.total_len);
+            while stale_files.len() > remains_capacity {
```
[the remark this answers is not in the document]
LykxSassinator: Not recommend. By this implementation, `stale_files` won't clear the `out of space` part.

For example, given that original `capacity` is `80`, and `stale_files.len() == 67`. If we restart engine with `capacity` == 50, extra stale files in `stale_files` list won't be cleared by each `purge` operation.

src/file_pipe_log/file_mgr.rs
```diff
@@ -0,0 +1,591 @@
...
+
+#[derive(Clone, PartialEq, Eq, Debug)]
+pub struct FileState {
+    pub first_seq: FileSeq,
+    pub total_len: usize,
+}
+
+pub struct FileList<F: FileSystem> {
+    /// Sequence number of the first file.
+    first_seq: FileSeq,
```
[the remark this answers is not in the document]
LykxSassinator: IMO, it's not recommended.

If we initialized `FileList` with `(1, VecDeq::default())`, related `FileState` is obvious -> `FileState { first_seq: 1, total_len: 0}`, but actually, it should be `FileState { first_seq: 0, total_len: 0 }`.

src/file_pipe_log/file_mgr.rs
```diff
@@ -0,0 +1,591 @@
...
+        Self { first_seq, fds }
+    }
+
+    #[inline]
+    pub fn len(&self) -> usize {
+        self.fds.len()
+    }
+
+    #[inline]
+    pub fn span(&self) -> (FileSeq, FileSeq) {
```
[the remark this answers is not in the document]
LykxSassinator: Why `Option` ? The meaning of `span()` is clear -- "Returns the `seqno` range of current file list." If the file list is empty, we should return (0, 0) to users.

src/file_pipe_log/file_mgr.rs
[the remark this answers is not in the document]
tabokie: std::mem::swap(self, &mut buf2).

src/file_pipe_log/file_mgr.rs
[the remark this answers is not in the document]
tabokie: Emmmm, just add it, what's so difficult about that? The point is to reduce lock footprint.
```
if state.total_len > capacity {
  stale_files.lock().remove_some();
} else {
  stale_files.lock().append(&mut to_add);
}
```

src/file_pipe_log/pipe_builder.rs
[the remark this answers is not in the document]
tabokie: Here `||{0}` is not a code per se, it's a const. For code like this:
```rust
#[cfg(feature = "failpoints")]
        {
            // Set header corrupted.
            let corrupted = || {
                fail::fail_point!("log_file_header::corrupted", |_| true);
                false
            };
            // Set abnormal DataLayout.
            let too_large = || {
                fail::fail_point!("log_file_header::too_large", |_| true);
                false
            };
            // Set corrupted DataLayout for `payload`.
            let too_small = || {
                fail::fail_point!("log_file_header::too_small", |_| true);
                false
            };
            if corrupted() {
                buf[0] += 1;
            }
            assert!(!(too_large() && too_small()));
            if too_large() {
                buf.encode_u64(0_u64)?;
            }
            if too_small() {
                buf.pop();
            }
        }
```
Lines like `buf[0] += 1;` and `buf.pop();` are codes that have side effect.

src/file_pipe_log/file_mgr.rs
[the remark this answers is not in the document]
LykxSassinator: Thx. What I've figured out is just like u mentioned, I still want to find an elegant way to reduce the cost of racing.

src/file_pipe_log/file_mgr.rs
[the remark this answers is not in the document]
tabokie: > first_seq + total_len consistently represent the next file.

This is the reason. It is consistent. (0,0) is not consistent and it propagates the check beyond where's necessary.

src/file_pipe_log/file_mgr.rs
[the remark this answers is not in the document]
tabokie: If you want to represent null, just use Option. It's the whole purpose of Option. Otherwise you return a meaningful result. (0,0) is not meaningful (because directly using it will break almost all calculations around first_seq, last_seq), it's just a manmade special marker.

src/file_pipe_log/file_mgr.rs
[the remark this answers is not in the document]
tabokie: There's no need to think beyond this. Memory overhead is negligible.

src/file_pipe_log/file_mgr.rs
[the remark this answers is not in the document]
LykxSassinator: Still not recommended. With this implementation, another corner case do not meet the requirement.
That is, when `file_seq` > `self.first_seq`, `split_by` should not return a splitted `FileList`, but `split_off` will return an unexpected `panic!`.
```Rust
// 11 12 13 14 15 => call split_by(17)
(1) current implementation => returns an empty FileList.
(2) this implementation => returns an (panic!) because of 17 - 11 > self.len()
```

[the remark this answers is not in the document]
tabokie: Then add a check.

src/engine.rs
```diff
@@ -2104,23 +2159,23 @@ mod tests {
             .must_rewrite_append_queue(Some(end - 1), None);
         assert!(start < engine.file_span(LogQueue::Append).0);
         assert_eq!(engine.file_count(Some(LogQueue::Append)), 1);
-        // no file have been physically deleted.
+        // Stale files have been reused.
```
tabokie, line 2162: Should check `stale_metadata` too.

src/engine.rs
```diff
@@ -2104,23 +2159,23 @@ mod tests {
...
-            &start
+            &(start + 20)
         );
         // reusing these files.
         for rid in 1..=5 {
             engine.append(rid, 1, 11, Some(&entry_data));
         }
         let start_1 = *fs.append_metadata.lock().unwrap().iter().next().unwrap();
-        assert!(start < start_1);
+        assert!(start <= start_1);
```
tabokie, line 2172: Test `stale_metadata.next()` has increased.

src/engine.rs
```diff
@@ -2104,23 +2159,23 @@ mod tests {
...
 
-        // reopen the engine and validate the stale files are removed
+        // reopen the engine and validate the stale files are reserved
         let engine = engine.reopen();
-        assert_eq!(fs.inner.file_count(), engine.file_count(None));
+        assert!(fs.inner.file_count() > engine.file_count(None));
```
tabokie, line 2176: Test the `fs.file_count()` before and after stays unchanged.

src/engine.rs
```diff
@@ -2104,23 +2159,23 @@ mod tests {
...
         let start_2 = *fs.append_metadata.lock().unwrap().iter().next().unwrap();
-        assert!(start_1 < start_2);
+        assert_eq!(start_1, start_2);
```
tabokie, line 2178: Test `stale_metadata.next()` is unchanged.

tests/failpoints/test_io_error.rs
```diff
@@ -439,3 +439,53 @@ fn test_file_allocate_error() {
...
+fn test_start_with_file_allocate_error() {
+    let dir = tempfile::Builder::new()
+        .prefix("test_start_with_file_allocate_error")
+        .tempdir()
+        .unwrap();
+    let cfg = Config {
+        dir: dir.path().to_str().unwrap().to_owned(),
+        target_file_size: ReadableSize::kb(1),
+        purge_threshold: ReadableSize::kb(10), // capacity is 12
+        ..Default::default()
```
tabokie, line 453: Ditto, explicitly set the config that is necessary for this test.

tests/failpoints/test_io_error.rs
```diff
@@ -439,3 +439,53 @@ fn test_file_allocate_error() {
     assert_eq!(engine.first_index(1).unwrap(), 1);
     assert_eq!(engine.last_index(1).unwrap(), 4);
 }
+
+#[test]
+fn test_start_with_file_allocate_error() {
```
tabokie, line 444: I don't know exactly what is tested. Because "whether the file is reused of not" is not checked at all. I suggest adding checks with `fs.file_count()`.

src/engine.rs
```diff
@@ -2230,4 +2285,128 @@ mod tests {
...
+        // Append data.
+        let entry_data = vec![b'x'; 512];
+        for rid in 1..=5 {
+            engine.append(rid, 11, 20, Some(&entry_data));
+        }
+        assert_eq!(engine.file_span(LogQueue::Append).0, start);
+        let (start, end) = engine.file_span(LogQueue::Append);
+        assert_eq!(file_system.inner.file_count(), engine.file_count(None));
+        drop(engine);
+        // Reopen the engine with a smaller capacity.
```
tabokie, line 2317: Try to use newlines more often to make the code clearer.
```suggestion
        drop(engine);

        // Reopen the engine with a smaller capacity.
```

src/engine.rs
```diff
@@ -2230,4 +2285,128 @@ mod tests {
...
+        for rid in 1..=20 {
+            engine.append(rid, 20, 31, Some(&entry_data));
+        }
+        engine.purge_expired_files().unwrap();
+        assert!(engine.file_span(LogQueue::Append).0 > start);
+        let (start, end) = engine.file_span(LogQueue::Append);
+        let engine = engine.reopen();
+        assert_eq!(engine.file_span(LogQueue::Append), (start, end));
+        // Stale files will be reused for new append files.
+        assert!(stale_count > (file_system.inner.file_count() - engine.file_count(None)));
```
tabokie, line 2343: Move this check right after the `append`, and check the `stale_count == capacity` here.

src/engine.rs
```diff
@@ -2230,4 +2285,128 @@ mod tests {
...
+        drop(engine);
+        // Reopen the engine without log recycling.
+        let cfg_v3 = Config {
+            target_file_size: ReadableSize::kb(2),
+            purge_threshold: ReadableSize::kb(100),
+            enable_log_recycle: false,
+            ..cfg_v2
+        };
+        let engine = RaftLogEngine::open_with_file_system(cfg_v3, file_system.clone()).unwrap();
+        // Restart the engine without log recycling, stale logs should be reserved.
```
tabokie, line 2406: I thought they should be removed? And if the capacity shrinks, redundant files should be removed too.

src/file_pipe_log/file_mgr.rs
```diff
@@ -0,0 +1,660 @@
...
+
+use super::format::{FileNameExt, LogFileFormat, StaleFileNameExt};
+
+/// Default buffer size for building stale file, unit: byte.
+const LOG_STALE_FILE_BUF_SIZE: usize = 16 * 1024 * 1024;
+/// Default path id for log files. 0 => Main dir
+const LOG_DEFAULT_PATH_ID: PathId = 0;
+
+pub type PathId = usize;
+pub type Paths = [String; 1];
```
tabokie, line 24:
```suggestion
pub type Paths = [PathBuf; 1];
```

src/file_pipe_log/file_mgr.rs
```diff
@@ -0,0 +1,660 @@
...
+
+    #[inline]
+    pub fn get(&self, file_seq: FileSeq) -> Result<Arc<F::Handle>> {
+        if !(self.first_seq..self.first_seq + self.fds.len() as u64).contains(&file_seq) {
+            return Err(Error::Corruption("file seqno out of range".to_owned()));
+        }
+        Ok(self.fds[(file_seq - self.first_seq) as usize]
+            .handle
+            .clone())
+    }
```
tabokie, line 85:
```suggestion
    pub fn get(&self, file_seq: FileSeq) -> Option<Arc<F::Handle>> {
        self.fds.get((file_seq - self.first_seq) as usize).map(|f| f.handle.clone())
    }
```

src/file_pipe_log/file_mgr.rs
```diff
@@ -0,0 +1,660 @@
...
+        self.stale_files.read().span()
+    }
+
+    #[inline]
+    pub fn push_back(&self, seq: FileSeq, handle: FileWithFormat<F>) -> FileState {
+        self.active_files.write().push_back(seq, handle)
+    }
+
+    #[inline]
+    pub fn get_target_file_size(&self) -> usize {
```
tabokie, line 238:
```suggestion
    pub fn target_file_size(&self) -> usize {
```

src/file_pipe_log/file_mgr.rs
```diff
@@ -0,0 +1,660 @@
...
+        self.target_file_size
+    }
+
+    /// Rotate a new file handle and return it to the caller.
+    ///
+    /// Returns the relevant `FileSeq`, and `FileWithFormat` of the file
+    /// handle to the caller.
+    /// Attention please, the returned fd should be manually appended to current
+    /// `FileCollection`.
```
tabokie, line 247: Maybe move the file writer creation inside, so that the whole step can be done atomically here.

src/file_pipe_log/file_mgr.rs
```diff
@@ -0,0 +1,660 @@
...
+                stale_files.push_back(
+                    seq,
+                    self.build_stale_file(
+                        file_id,
+                        LogFileFormat::new(Version::V2, 0 /* alignment */),
+                    )?,
+                );
+            }
+        }
+        info!(
```
tabokie, line 397: Move log inside the if block.

LykxSassinator:
Ping @tabokie ,thx

tests/failpoints/test_io_error.rs
LykxSassinator: Extra annotations have been supplemented and the case has been refined.

This case is meant to mock the scenario that Engine is restarted after it is failed to be opened firstly when generating stale files in fail.

tabokie (commented):
Not much time this week. Some minor comments and please fix the stable build by applying https://github.com/tikv/raft-engine/pull/285.

src/file_pipe_log/file_mgr.rs
```diff
@@ -0,0 +1,742 @@
...
+    ) -> Result<FileWithFormat<F>> {
+        let path_id = FileCollection::<F>::get_valid_path(&self.paths, self.target_file_size);
+        let file_path = file_id.build_stale_file_path(&self.paths[path_id]);
+        let fd = Arc::new(self.file_system.create(&file_path)?);
+        let mut writer = self.file_system.new_writer(fd.clone())?;
+        let mut written = 0_usize;
+        let buf = vec![0; std::cmp::min(LOG_STALE_FILE_BUF_SIZE, self.target_file_size)];
+        while written <= self.target_file_size {
+            writer.write_all(&buf).unwrap_or_else(|e| {
+                panic!("failed to prepare stale file: {}", e);
```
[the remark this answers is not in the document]
LykxSassinator: It's tested in `test_start_with_stale_file_allocate_error`.
IMO, if the `write` was abnormal when starting, the incoming bytes appending operations might also fail by `Engine.write()`. So, it's should throw the `panic` to users.

src/engine.rs
```diff
@@ -1929,26 +1950,46 @@ mod tests {
...
         fn update_metadata(&self, path: &Path, delete: bool) -> bool {
-            let id = FileId::parse_file_name(path.file_name().unwrap().to_str().unwrap()).unwrap();
-            if id.queue == LogQueue::Append {
-                if delete {
-                    self.append_metadata.lock().unwrap().remove(&id.seq)
-                } else {
-                    self.append_metadata.lock().unwrap().insert(id.seq)
+            let path = path.file_name().unwrap().to_str().unwrap();
+            let parse_append = FileId::parse_file_name(path);
+            let parse_stale = FileId::parse_stale_file_name(path);
```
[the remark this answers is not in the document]
LykxSassinator: Good idea. It's redundant to return a FileId with `LogQueue::Append`, as Reserved files only available for `LogQueue::Append`

LykxSassinator:
Wait for final review after #285 is merged.

LykxSassinator:
> Wait for final review after #285 is merged.

Ignore, related updates of the dependency on rust version have been updated in this pr.

LykxSassinator:
@tabokie ping.

src/file_pipe_log/file_mgr.rs
```diff
@@ -0,0 +1,751 @@
...
+    pub fn target_file_size(&self) -> usize {
+        self.target_file_size
+    }
+
+    #[inline]
+    /// Returns the last active file handle to the caller.
+    ///
+    /// Attention, this func can only be called when the caller firstly
+    /// tried to access the latest active file.
+    pub fn fetch_active_file(&self) -> Option<ActiveFile<F>> {
```
tabokie, line 266: Remove the `_init_from_empy_active_files`, and change this function into `fn get_or_create_active_file(&mut self) -> (FileWithFormat, bool)`

src/file_pipe_log/pipe_builder.rs
```diff
@@ -398,29 +435,55 @@ impl<F: FileSystem> DualPipesBuilder<F> {
...
+                            format: LogFileFormat::new(Version::V2, 0),
+                            path_id: f.path_id,
+                        })
+                        .collect(),
+                ),
+            ),
+            LogQueue::Rewrite => (
+                0,
+                &self.rewrite_files,
+                FileList::new(0, VecDeque::<FileWithFormat<F>>::default()),
```
tabokie, line 458:
```suggestion
                FileList::default(),
```

src/file_pipe_log/pipe_builder.rs
```diff
@@ -398,29 +435,55 @@ impl<F: FileSystem> DualPipesBuilder<F> {
...
+            LogQueue::Append => (
+                self.cfg.recycle_capacity(),
+                &self.append_files,
+                FileList::new(
+                    self.reserved_files.first().map(|f| f.seq).unwrap_or(0),
+                    self.reserved_files
+                        .iter()
+                        .map(|f| FileWithFormat {
+                            handle: f.handle.clone(),
+                            format: LogFileFormat::new(Version::V2, 0),
```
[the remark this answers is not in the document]
LykxSassinator: As we do not need to parse the header and format of reserved files, the `f.format` of each file in `self.reserved_files` is just `None`.
So, it's no need to give the `f.format` to it here. And I've add related comments.

tests/failpoints/test_io_error.rs
```diff
@@ -439,3 +439,62 @@ fn test_file_allocate_error() {
...
+        target_file_size: ReadableSize::kb(1),
+        purge_threshold: ReadableSize::kb(10), // capacity is 12
+        enable_log_recycle: true,
+        prefill_for_recycle: true,
+        ..Default::default()
+    };
+    let entry = vec![b'x'; 1024];
+    // Mock that the engine is failed because of the stale file
+    // with seqno[5] failed to be generated.
+    assert!(catch_unwind_silent(|| {
```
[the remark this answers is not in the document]
LykxSassinator: It's incorrect here, it have been revised.

LykxSassinator:
ping @tabokie  PTAL, thx.

src/file_pipe_log/pipe.rs
```diff
@@ -782,19 +483,20 @@ mod tests {
...
             // Bypass pipe log
             let mut reader = build_file_reader(
                 fs.as_ref(),
-                Arc::new(fs.open(handle.id.build_file_path(path)).unwrap()),
+                Arc::new(fs.open(handle.id.build_stale_file_path(path)).unwrap()),
             )
             .unwrap();
-            assert_eq!(reader.read(handle).unwrap(), content(i));
+            assert!(reader.read(handle).unwrap().is_empty());
             // Delete file so that it cannot be reused.
```
tabokie, line 644: The original intention of this test is broken. You should either remove the comment or try some other way to test it.

tabokie (commented):
I took the liberty of cleaning up the abstraction myself, it grows to be, somewhat superfluous IMO.

And by the way, https://github.com/tikv/raft-engine/pull/278#discussion_r1060299404 is not addressed. When you use `&mut self`, there's no need to lock again, because mutable reference guarantees exclusive access.

LykxSassinator:
> I took the liberty of cleaning up the abstraction myself, it grows to be, somewhat superfluous IMO.
> 
> And by the way, [#278 (comment)](https://github.com/tikv/raft-engine/pull/278#discussion_r1060299404) is not addressed. When you use `&mut self`, there's no need to lock again, because mutable reference guarantees exclusive access.

Thx for your suggestions. I'll take some time to comprehend the newly introduced refactoring works.

src/file_pipe_log/pipe_builder.rs
```diff
@@ -429,10 +429,13 @@ impl<F: FileSystem> DualPipesBuilder<F> {
     fn initialize_files(&mut self) -> Result<()> {
         let now = Instant::now();
         let target_file_size = self.cfg.target_file_size.0 as usize;
-        let mut target = self
```
tabokie: Return early if `!prefill_for_recycle`.

src/file_pipe_log/pipe.rs
```diff
@@ -232,15 +235,16 @@ impl<F: FileSystem> SinglePipe<F> {
...
-        if let Some(fd) = {
-            let files = self.active_files.read();
-            let off = handle.id.seq - files[0].seq;
-            files.get(off as usize).map(|f| f.handle.clone())
-        } {
-            let mut reader = build_file_reader(self.file_system.as_ref(), fd)?;
-            reader.read(handle)
-        } else {
-            Err(Error::Corruption("file seqno out of range".to_string()))
+        let files = self.active_files.read();
```
tabokie, line 238: This holds lock for IO.

src/file_pipe_log/pipe.rs
```diff
@@ -83,15 +83,16 @@ impl<F: FileSystem> SinglePipe<F> {
...
         };
         let default_format = LogFileFormat::new(cfg.format_version, alignment());
 
         // Open or create active file.
-        if active_files.is_empty() {
+        let no_active_files = active_files.is_empty();
+        if no_active_files {
             let path_id = DEFAULT_PATH_ID;
-            let file_id = FileId::new(queue, FileSeq::default());
+            let file_id = FileId::new(queue, 1 as FileSeq);
```
tabokie, line 95: What's the issue with 0?

src/config.rs
```diff
@@ -186,12 +187,12 @@ impl Config {
             return 0;
         }
         if self.enable_log_recycle && self.purge_threshold.0 >= self.target_file_size.0 {
-            // (1) At most u32::MAX so that the file number can be capped into an u32
-            // without colliding. (2) Add some more file as an additional buffer to
-            // avoid jitters.
+            // (1) At most MAX_RECYCLE_CAPACITY so that the file number can be capped into
```
tabokie, line 190: Why have a limit?

src/file_pipe_log/pipe.rs
LykxSassinator: At the previous work, we build the prerequisite that **All file_seqno starts from 1** and keep this regulation in later developments.
For example:
https://github.com/tikv/raft-engine/blob/8dd2a39f359ff16f5295f35343f626e0c10132fa/src/file_pipe_log/pipe.rs#L271-L272
IMO, I still recommend to keep it.

src/file_pipe_log/pipe_builder.rs
LykxSassinator: Not recommended. 
When `!prefill_for_recycle`, it should check whether there still exists remained `recycled_files` and clear them if exists. And by the way, the later processing when `!prefill_for_recycle` just do several checking jobs, not costing much on CPU or other resources.

src/config.rs
LykxSassinator: Concerns on exceeding the limitation of open files when testing. It has been revoked with refactoring the related test case.

src/file_pipe_log/pipe.rs
tabokie: So to put it plainly, the issue with 0 is it will break backward compatibility (for debug binary). You should replace 1 with a constant.

tests/failpoints/test_engine.rs
```diff
@@ -724,3 +724,32 @@ fn test_build_engine_with_datalayout_abnormal() {
...
+        prefill_for_recycle: true,
+        ..Default::default()
+    };
+    if let Some(max_fd_limit) = fdlimit::raise_fd_limit() {
+        let recycle_capacity = std::cmp::min(
+            (cfg.purge_threshold.0 / cfg.target_file_size.0) as usize + 2,
+            u32::MAX as usize,
+        );
+        if recycle_capacity >= max_fd_limit as usize {
+            // It should panic by **open too many files**.
```
tabokie, line 749: Where?

src/config.rs
```diff
@@ -172,9 +186,9 @@ impl Config {
             return 0;
         }
         if self.enable_log_recycle && self.purge_threshold.0 >= self.target_file_size.0 {
-            // (1) At most u32::MAX so that the file number can be capped into an u32
-            // without colliding. (2) Add some more file as an additional buffer to
-            // avoid jitters.
+            // (1) At most u32::MAX  so that the file number can be capped into
```
tabokie, line 189: remove the redundant space

src/file_pipe_log/pipe.rs
```diff
@@ -349,25 +242,39 @@ impl<F: FileSystem> SinglePipe<F> {
         reader.read(handle)
     }
 
+    /// Returns a shared [`LogFd`] for the specified file sequence number.
+    fn get_fd(&self, file_seq: FileSeq) -> Result<Arc<F::Handle>> {
```
tabokie, line 246: Use the old implementation and placement.

tests/failpoints/test_engine.rs
LykxSassinator: Incorrect case, fixed.

src/config.rs
```diff
@@ -172,9 +186,9 @@ impl Config {
             return 0;
         }
         if self.enable_log_recycle && self.purge_threshold.0 >= self.target_file_size.0 {
-            // (1) At most u32::MAX so that the file number can be capped into an u32
-            // without colliding. (2) Add some more file as an additional buffer to
-            // avoid jitters.
+            // (1) At most u32::MAX so that the file number can be capped into
+            // an u32 without colliding. (2) Add some more file as an additional
+            // buffer to avoid jitters.
```
tabokie, line 191: revert.

tests/failpoints/test_engine.rs
```diff
@@ -943,7 +943,7 @@ fn test_split_rewrite_batch_with_only_kvs() {
         engine.purge_manager().must_rewrite_rewrite_queue();
     }
     {
-        let _f = fail::FailGuard::new("force_use_atomic_group", "return");
+        let _f = FailGuard::new("force_use_atomic_group", "return");
```
tabokie, line 946: fail already has a FailGuard now, let's remove our own implementation.

src/engine.rs
```diff
@@ -700,14 +700,43 @@ mod tests {
...
+            let cfg = Config {
+                dir: dir.path().to_str().unwrap().to_owned(),
+                target_file_size: ReadableSize(1),
+                enable_log_recycle: true,
+                prefill_for_recycle: true,
+                ..Default::default()
+            };
+            if let Some(max_fd_limit) = fdlimit::raise_fd_limit() {
+                if cfg.recycle_capacity() >= max_fd_limit as usize {
+                    // It should return Error **open too many files**.
```
tabokie, line 725: Again, where is the error raised? I didn't find any using the key word "too many files".

tabokie: Anyway, it's a bad idea to enforce a limit using `recycle_capacity`. (1) number of opened files will not be affected by whether or not `recycle_capacity` is capped (2) even if append queue files are capped, rewrite queue files will grow out of limit too.

LykxSassinator: > Again, where is the error raised? I didn't find any using the key word "too many files".

When prefilling with recycled files by setting `prefill_recycle_logs`, it will reach this error by setting `target_file_size == 1`.

LykxSassinator: > Anyway, it's a bad idea to enforce a limit using `recycle_capacity`. (1) number of opened files will not be affected by whether or not `recycle_capacity` is capped (2) even if append queue files are capped, rewrite queue files will grow out of limit too.

Seems that this case is redundant and unnecessary. Originally, i just wanna add a case to mock the scenario where users set a huge `recycle_capacity` and `Engine` fails to be open as it encounters `too many open files` error.

LykxSassinator:
ping @tabokie PTAL, thx.

src/file_pipe_log/pipe.rs
tabokie: "placement" too, so the diff is shorter and history is clean.

src/file_pipe_log/pipe.rs
```diff
@@ -428,68 +330,105 @@ impl<F: FileSystem> SinglePipe<F> {
...
-        } else if prev == current {
-            return Ok(0);
-        }
-        for seq in prev.first_seq..current.first_seq {
-            let file_id = FileId {
-                queue: self.queue,
-                seq,
+        let (len, purged_files) = {
+            let mut files = self.active_files.write();
+            let off = (file_seq - files[0].seq) as usize;
```
tabokie, line 349: `file_seq < files[0].seq`?

src/file_pipe_log/pipe.rs
```diff
@@ -428,68 +330,105 @@ impl<F: FileSystem> SinglePipe<F> {
...
+            if off >= files.len() {
+                return Err(box_err!("Active file cannot be purged"));
+            }
+            let mut tail = files.split_off(off);
+            std::mem::swap(&mut tail, &mut files);
+            (files.len(), tail)
+        };
+        let purged_len = purged_files.len();
+        {
+            let remains_capacity = self.capacity.saturating_sub(len);
```
tabokie, line 359: Only enter this block if capacity is non-zero.

src/file_pipe_log/pipe.rs
```diff
@@ -428,68 +330,105 @@ impl<F: FileSystem> SinglePipe<F> {
...
+                        }
+                        Err(e) => error!("failed to reuse purged file {:?}", e),
+                    }
                 }
+                self.file_system.delete(&path)?;
+            }
+            // If there exists several recycled files out of space, contained in
+            // `self.recycled_files` and `purged_files`, we should check and remove them
+            // to avoid the `size` of whole files beyond `self.capacity`.
+            if recycled_len > remains_capacity {
```
tabokie, line 399: This block is not covered.

LykxSassinator: It's a stale and redundant judgement here, as we have ensured the remained count of recycled files when restarting.

tabokie: Add a debug assertion.

src/file_pipe_log/pipe.rs
tabokie: Is there a test?

src/file_pipe_log/pipe_builder.rs
```diff
@@ -403,45 +433,89 @@ impl<F: FileSystem> DualPipesBuilder<F> {
         )
     }
 
-    /// Builds a new storage for the specified log queue.
-    fn build_pipe(&self, queue: LogQueue) -> Result<SinglePipe<F>> {
-        let files = match queue {
-            LogQueue::Append => &self.append_files,
-            LogQueue::Rewrite => &self.rewrite_files,
+    fn initialize_files(&mut self) -> Result<()> {
+        let now = Instant::now();
```
tabokie, line 437: Move inside `if to_create > 0 {`.

src/file_pipe_log/pipe.rs
tabokie: Placement means put this function before `rotate_imp`, where it was before this PR.

tabokie (commented):
Needs changelog under "New Features", rest looks good.

src/file_pipe_log/pipe.rs
tabokie: Ping @LykxSassinator

src/file_pipe_log/pipe.rs
tabokie: Ping

LykxSassinator:
ping @tabokie, thx.

CHANGELOG.md
```diff
@@ -11,6 +11,10 @@
 
 * Fix data loss caused by aborted rewrite operation. Downgrading to an earlier version without the fix may produce phantom Raft Groups or keys, i.e. never written but appear in queries.
 
+### New Features
```
tabokie, line 14: test

src/file_pipe_log/pipe.rs
tabokie: Ping again.

src/file_pipe_log/pipe.rs
tabokie: Ping.

tabokie (commented):
//

src/file_pipe_log/pipe.rs
LykxSassinator: This branch has already been tested in `test_pipe_log`:
```Rust
        ...
        // cannot purge active file
        assert!(pipe_log.purge_to(FileId { queue, seq: 4 }).is_err());
        ...
```
And more boundary test cases have been supplemented into `test_pipe_log_with_recycle`.

src/file_pipe_log/pipe.rs
LykxSassinator: Supplemented.

src/file_pipe_log/pipe.rs
tabokie: Let's keep this in mind. Line coverage can only be useful when there's something not covered. What's covered is far from enough and cannot be used as a proof of software quality. (MC/DC coverage, on the other hand, is a much better criterion.) Whenever there's a bug, it's vital to keep the condition covered before fixing it.

tabokie (approved):

PR #286, opened 2022-12-05
read entries using AIO (version 1.0)

ustc-wxy:


src/engine.rs
```diff
@@ -308,11 +311,95 @@ where
...
+                    0 => true,
+                    _ => ents_idx[t-1].entries.unwrap() != ents_idx[t].entries.unwrap(),
+                }{
+                    block_sum += 1;
+                    new_block_flags.push(true);
+                }else{
+                    new_block_flags.push(false);
+                }
+            }
+            let mut a_list: Vec<aiocb> = Vec::with_capacity(block_sum);
```
tabokie, line 352: Hide aiocb inside `AioContext` (it holds `Vec<aiocb>` and create new ones when a new read request is issued via `fs::read_async`). No one other than the `DefaultFileSystem` should have access to aiocb.

tabokie (commented):
Check the contributing guide [here](https://github.com/tikv/raft-engine#contributing). Needs to sign the commit and run `make format & make clippy`.

src/engine.rs
ustc-wxy: Do you mean to use one AioContext to manage all read requests? Then I need to redesign the AioContext, like this
`pub struct AioContext{
    fd_vec: Vec<Arc<LogFd>>,
    aio_vec: Vec<aiocb>,
    buf_vec: Vec<Vec<u8>>,
}`

tabokie: I'm not sure if there's a performance difference between waiting on each aiocb separately, or waiting all of them in one single syscall (maybe you can benchmark it too).

If there's no difference, then it makes sense to create new AioContext for each request. But either way, all the details must be put inside the file system implementation. E.g.
```
impl AioContext {
  pub fn wait() -> Result<()>;
  // UB if `wait()` is not called and returns `Ok(())`.
  pub fn data() -> &[u8];
}
```

src/env/default.rs
```diff
@@ -97,6 +102,21 @@ impl LogFd {
         Ok(readed)
     }
 
+    pub fn read_aio(&self, seq: usize, ctx: &mut AioContext, offset: u64) {
+        let mut buf = ctx.buf_vec.last().unwrap().lock().unwrap();
```
tabokie, line 106: The lock is meaningless. The mutable reference (`&mut AioContext`) ensures there's only one thread accessing the ctx.

Of course, due to the unsafe block, the `buf_vec` is in fact "leaked" to aio code. That's why we must manually guarantee there's no one reading or writing to `buf_vec` after calling `aio_read`.

src/env/default.rs
```diff
@@ -257,13 +277,76 @@ impl WriteExt for LogFile {
...
+        Self {
+            aio_vec: vec,
+            buf_vec: vec![],
+        }
+    }
+}
+
+impl AsyncContext for AioContext {
+    fn single_wait(&mut self, seq: usize) -> IoResult<usize> {
+        let buf_len = self.buf_vec[seq].lock().unwrap().len();
```
tabokie, line 301: ditto, the `&mut self` has the same effect.

src/env/default.rs
```diff
@@ -257,13 +277,76 @@ impl WriteExt for LogFile {
...
 impl FileSystem for DefaultFileSystem {
     type Handle = LogFd;
     type Reader = LogFile;
     type Writer = LogFile;
 
+    fn read_aio(
+        &self,
+        handle: Arc<Self::Handle>,
+        seq: usize,
+        ctx: &mut AioContext,
```
tabokie, line 343: `AioContext` should be a generic type, different file system implementation can have different context.

The code will look like:
```rust
trait WaitData {
  fn wait(&mut self, index: usize) -> &[u8];
}

trait FileSystem {
  type AsyncContext: WaitData;
  fn new_async_context(&self) -> Self::AsyncContext;
}

impl FileSystem for DefaultFileSystem {
  type AsyncContext = AioContext;
  fn new_async_context(&self) {
    AioContext::new()
  }
}
```

src/env/default.rs
```diff
@@ -257,12 +275,100 @@ impl WriteExt for LogFile {
...
+    }
+
+    fn data(&self, seq: usize) -> Vec<u8> {
+        self.buf_vec[seq].to_vec()
+    }
+
+    fn single_wait(&mut self, seq: usize) -> IoResult<usize> {
+        let buf_len = self.buf_vec[seq].len();
+        unsafe {
+            loop {
```
tabokie, line 328: What if the read hits EOF? It will loop forever? This needs testing as well.

tabokie (commented):
1. Please run the checks locally before submitting a commit, as a contributor the basic level of respect is to pass CI before requesting reviews.
2. A good PR description should contain three parts: the problem, the solution, the verification method (test plan).

src/pipe_log.rs
```diff
@@ -172,6 +176,19 @@ pub trait PipeLog: Sized {
...
 
+    /// Read entries from pipe logs using 'Async IO'.
+    fn async_entry_read<M: Message + MessageExt<Entry = M>>(
+        &self,
+        ents_idx: &mut Vec<EntryIndex>,
+        vec: &mut Vec<M::Entry>,
+    ) -> Result<()>;
+
+    /// Reads bytes from multi blocks using 'Async IO'.
+    fn async_read_bytes(
```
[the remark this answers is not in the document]
ustc-wxy: @tabokie Now my implementation idea is to read all bytes through `async_entry_read`, and then parse them in turn. like
```
pub fn fetch_entries_to_aio(){
    ...
    let bytes = async_entry_read();
    for (idx,i) in ents_idx.iter().enumerate(){
          entry = parse_from_bytes(byte[idx]);
          vec.push(entry);
    }
    ...
}
```
Does it ok？

src/env/default.rs
```diff
@@ -257,12 +275,100 @@ impl WriteExt for LogFile {
...
+        }
+    }
+
+    pub fn new_reader(&mut self, fd: Arc<LogFd>) -> IoResult<()> {
+        self.inner = Some(fd);
+        Ok(())
+    }
+}
+
+impl AsyncContext for AioContext {
```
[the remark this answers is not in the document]
ustc-wxy: @tabokie Whether it is necessary to consider such situations? That is in an `fetch_entries_to_aio` call, some `handles` belong to the Append queue and others belong to the Rewrite queue. They correspond to different `files_ system`, then you need to call  2 times `async_finish()`.In more complex cases, they are interspersed, like _Append, Rewrite, Append, Rewrite, Rewrite..._，then the design of `async_finish()` will doesn't work.

src/pipe_log.rs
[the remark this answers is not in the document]
tabokie: Yes, that's what I meant. But the name should not be `async_entry_read`, the underlying pipe is not aware of the "entry" concept.

src/env/default.rs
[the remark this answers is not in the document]
tabokie: Like this:

```
impl DualPipes {
  fn read_async(&self, handls: Vec<FileBlockHandle>) ->Vec<Vec<u8>> {
    let mut ctx = fs.new_context();
    for handle in handles {
      fs.read_async(&mut ctx, handle);
    }
    fs.async_finish(ctx);
  }
}

```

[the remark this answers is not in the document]
ustc-wxy: @tabokie How can I determine which `fs` to use? Use `self.pipes[LogQueue::Append].file_system`, or `self.pipes[LogQueue::Rewrite].file_system`, or both?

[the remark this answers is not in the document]
tabokie: Either one is fine. They are always the same.

ustc-wxy:
I have completed the code modification according to your comments and run the unit test locally, all of them have passed. PTAL. @tabokie

tabokie:
@ustc-wxy format and clippy failed.

src/pipe_log.rs
```diff
@@ -172,6 +173,9 @@ pub trait PipeLog: Sized {
     /// Reads some bytes from the specified position.
     fn read_bytes(&self, handle: FileBlockHandle) -> Result<Vec<u8>>;
 
+    /// Reads bytes from multi blocks using 'Async IO'.
+    fn async_read_bytes(&self, ents_idx: &mut Vec<EntryIndex>) -> Result<Vec<Vec<u8>>>;
```
tabokie, line 177: As mentioned before, pipe is not aware of the "entry" concept. Here should use `Vec<FileBlockHandle>`.

src/file_pipe_log/pipe.rs
```diff
@@ -444,6 +451,32 @@ impl<F: FileSystem> PipeLog for DualPipes<F> {
...
+        let fs = &self.pipes[LogQueue::Append as usize].file_system;
+        let mut ctx = fs.new_async_io_context(blocks.len()).unwrap();
+
+        for block in blocks.iter_mut() {
+            match block.id.queue {
+                LogQueue::Append => {
+                    self.pipes[LogQueue::Append as usize].async_read(block, &mut ctx);
+                }
+                LogQueue::Rewrite => {
+                    self.pipes[LogQueue::Rewrite as usize].async_read(block, &mut ctx);
```
tabokie, line 472: Change the function to `async_read(ctx, buf)`. It's the common order to pass in a context.

src/file_pipe_log/pipe.rs
```diff
@@ -444,6 +451,32 @@ impl<F: FileSystem> PipeLog for DualPipes<F> {
...
+                }
+            }
+        }
+        let res = fs.async_finish(&mut ctx).unwrap();
```
tabokie, line 476: Change it to `async_finish(ctx: Context)` (instead of `async_finish(ctx: &mut Context)`). This makes sure the context isn't reused afterwards.

src/file_pipe_log/pipe.rs
```diff
@@ -444,6 +451,32 @@ impl<F: FileSystem> PipeLog for DualPipes<F> {
         self.pipes[handle.id.queue as usize].read_bytes(handle)
     }
 
+    #[inline]
+    fn async_read_bytes(&self, ents_idx: &mut Vec<EntryIndex>) -> Result<Vec<Vec<u8>>> {
+        let mut blocks: Vec<FileBlockHandle> = vec![];
```
tabokie, line 456: Won't this compile? `let mut blocks = Vec::new();`

src/file_pipe_log/pipe.rs
```diff
@@ -254,6 +254,13 @@ impl<F: FileSystem> SinglePipe<F> {
         reader.read(handle)
     }
 
+    fn async_read(&self, block: &mut FileBlockHandle, ctx: &mut F::AsyncIoContext) {
```
tabokie, line 257: There's no need for the `block` to be mutable.

src/file_pipe_log/pipe.rs
```diff
@@ -254,6 +254,13 @@ impl<F: FileSystem> SinglePipe<F> {
...
+        let fd = self.get_fd(block.id.seq).unwrap();
+        let buf = vec![0_u8; block.len];
+
+        self.file_system.async_read(ctx, fd, buf, block).unwrap();
```
tabokie, line 261: Why not creating the vector inside `async_read`? So there's one less argument to pass into (and validate).

src/env/obfuscated.rs
```diff
@@ -84,11 +86,54 @@ impl ObfuscatedFileSystem {
...
+    fn submit_read_req(&mut self, buf: Vec<u8>, offset: u64) -> IoResult<()> {
+        self.0.submit_read_req(buf, offset)
+    }
+}
 
 impl FileSystem for ObfuscatedFileSystem {
     type Handle = <DefaultFileSystem as FileSystem>::Handle;
     type Reader = ObfuscatedReader;
     type Writer = ObfuscatedWriter;
+    type AsyncIoContext = ObfuscatedContext;
```
tabokie, line 112: Why? Doesn't this work already?
`type AsyncIoContext = <DefaultFileSystem as FileSystem>::AsyncIoContext`

src/env/obfuscated.rs
```diff
@@ -84,11 +86,54 @@ impl ObfuscatedFileSystem {
         self.files.load(Ordering::Relaxed)
     }
 }
+pub struct ObfuscatedContext(<DefaultFileSystem as FileSystem>::AsyncIoContext);
+impl AsyncContext for ObfuscatedContext {
+    fn wait(&mut self) -> IoResult<usize> {
+        self.0.wait()
+    }
```
tabokie, line 93: These codes are not covered, meaning you didn't test `ObfuscatedFileSystem` in unit tests.

src/env/default.rs
```diff
@@ -257,12 +277,113 @@ impl WriteExt for LogFile {
...
+        }
+        Self {
+            inner: None,
+            index: 0,
+            aio_vec,
+            buf_vec,
+        }
+    }
+
+    pub fn set_fd(&mut self, fd: Arc<LogFd>) {
```
tabokie, line 303: I understand this is to hide the `LogFd`. But it isn't necessary if you treat `AsyncContext` as a data-only struct (inline all code into file system, and remove the trait `AsyncContext` entirely). i.e.
```
impl SomeFileSystem {
  fn async_read(ctx: &mut Self::Context, handle: Self::Handle, block: FileBlockHandle) {
    handle.submit_async_read(ctx.buf[i], ...)?;
  }
}
```

src/env/default.rs
```diff
@@ -257,12 +277,113 @@ impl WriteExt for LogFile {
...
+    pub(crate) buf_vec: Vec<Vec<u8>>,
+}
+impl AioContext {
+    pub fn new(block_sum: usize) -> Self {
+        let mut aio_vec = vec![];
+        let buf_vec = vec![];
+        unsafe {
+            for _ in 0..block_sum {
+                aio_vec.push(mem::zeroed::<libc::aiocb>());
+            }
```
tabokie, line 293: Why creating them before hand, instead of creating them the same time as `buf_vec`?

ustc-wxy: @tabokie If `libc:: aiocb` is created in multiple function calls, the compiler will assign the same address to the pointer every time, which will leads incorrect result. I'm not so familiar with the memory allocate API of trust, do you have any other good methods?

src/env/obfuscated.rs
ustc-wxy: @tabokie I introduced the `ObfuscatedContext` struct based on your comments:https://github.com/tikv/raft-engine/pull/286#discussion_r1058024065

src/file_pipe_log/pipe.rs
ustc-wxy: @tabokie Did you mean `async_read(ctx, blocks)`? Like
```
impl<F: FileSystem> PipeLog for DualPipes<F> {
  fn async_read_bytes(blocks:Vec<FileBlockHandle>) -> Result<Vec<Vec<u8>>>{
      ...
      self.pipes[LogQueue::Append].async_read(ctx,blocks);
      let bytes = fs.async_finish(ctx);
      ...
  }
}
```

tabokie: Yes

src/env/obfuscated.rs
tabokie: Adding a layer of wrapper is only useful if you have implemented something on top of the base context. Right now you didn't implement anything, so there is no need to wrap it.

src/env/default.rs
tabokie: I don't think it's possible to have multiple variables with the same address. It's possible that you mistakenly free some of them.

ustc-wxy:
@tabokie PTAL

src/file_pipe_log/pipe.rs
```diff
@@ -184,7 +185,6 @@ impl<F: FileSystem> SinglePipe<F> {
         }
         Ok(files[(file_seq - files[0].seq) as usize].handle.clone())
     }
-
```
tabokie, line 187: restore the newline.

src/file_pipe_log/pipe.rs
```diff
@@ -254,6 +254,13 @@ impl<F: FileSystem> SinglePipe<F> {
         reader.read(handle)
     }
 
+    fn async_read(&self, blocks: Vec<FileBlockHandle>, ctx: &mut F::AsyncIoContext) {
```
tabokie, line 257: Move ctx first, as explained in https://github.com/tikv/raft-engine/pull/286#discussion_r1118813985.

src/file_pipe_log/pipe.rs
```diff
@@ -254,6 +254,13 @@ impl<F: FileSystem> SinglePipe<F> {
...
+        for block in blocks.iter() {
+            let fd = self.get_fd(block.id.seq).unwrap();
+            self.file_system.async_read(ctx, fd, block).unwrap();
```
tabokie, line 260: shouldn't use unwrap.

src/env/obfuscated.rs
```diff
@@ -89,6 +90,30 @@ impl FileSystem for ObfuscatedFileSystem {
...
+    ) -> IoResult<()> {
+        self.inner.async_read(ctx, handle, block)
+    }
+
+    fn async_finish(&self, ctx: &mut Self::AsyncIoContext) -> IoResult<Vec<Vec<u8>>> {
+        let base = self.inner.async_finish(ctx).unwrap();
+        let mut res = vec![];
+        for v in base {
+            let mut temp = vec![];
+            //do obfuscation.
```
tabokie, line 109:
```suggestion
            // do obfuscation.
```

src/env/obfuscated.rs
```diff
@@ -89,6 +90,30 @@ impl FileSystem for ObfuscatedFileSystem {
...
+        block: &FileBlockHandle,
+    ) -> IoResult<()> {
+        self.inner.async_read(ctx, handle, block)
+    }
+
+    fn async_finish(&self, ctx: &mut Self::AsyncIoContext) -> IoResult<Vec<Vec<u8>>> {
+        let base = self.inner.async_finish(ctx).unwrap();
+        let mut res = vec![];
+        for v in base {
+            let mut temp = vec![];
```
tabokie, line 108: I don't know why you need to create another vector instead of modifying `base` directly.

src/env/mod.rs
```diff
@@ -7,14 +7,25 @@ use std::sync::Arc;
 mod default;
 mod obfuscated;
 
+pub use default::AioContext;
```
tabokie, line 10: remove this, as explained in https://github.com/tikv/raft-engine/pull/286/files#r1058018164

tabokie (commented):
https://github.com/tikv/raft-engine/pull/286#discussion_r1118822597 and https://github.com/tikv/raft-engine/pull/286#discussion_r1118815799 are not addressed.

ustc-wxy:
```
fn async_finish(&self, ctx: &mut Self::AsyncIoContext) -> IoResult<Vec<Vec<u8>>> {
      for seq in 0..ctx.aio_vec.len() {
          let buf_len = ctx.buf_vec[seq].len();
          aio_suspend(...);
          assert_eq!(ctx.aio_vec[seq].as_mut().aio_return().unwrap(), buf_len);
      }
      ...
}
```
@tabokie I really understand what you mean, but we must call `aio_return()` to return the final result of the asynchronous operation in `file_system::async_finish()`. It's defined in nix::aio：
`pub fn aio_read(aiocbp: *mut aiocb) -> ::c_int;` 
So I need to transfer `ctx` through mutable borrowing.

tabokie:
https://github.com/tikv/raft-engine/pull/286#discussion_r1118822597 is still not addressed. Is there any confusion?

> shouldn't use unwrap.

`expect` is the same as `unwrap` except with some message. What I meant is the code shouldn't panic on error. It should bubble the error. And there're obviously other unwrap-s (like the line above it) beside the line I was pointing at.

ustc-wxy:
@tabokie PTAL, thx!

src/engine.rs
```diff
@@ -308,11 +308,45 @@ where
                 vec.push(read_entry_from_file::<M, _>(self.pipe_log.as_ref(), i)?);
             }
             ENGINE_READ_ENTRY_COUNT_HISTOGRAM.observe(ents_idx.len() as f64);
+
             return Ok(ents_idx.len());
         }
         Ok(0)
     }
 
+    pub fn fetch_entries_to_aio<M: Message + MessageExt<Entry = M>>(
```
tabokie, line 317:
```suggestion
    pub fn fetch_entries_to_aio<M: MessageExt>(
```

src/env/mod.rs
```diff
@@ -10,11 +10,21 @@ mod obfuscated;
...
 
+use crate::pipe_log::FileBlockHandle;
 /// FileSystem
 pub trait FileSystem: Send + Sync {
     type Handle: Send + Sync + Handle;
     type Reader: Seek + Read + Send;
     type Writer: Seek + Write + Send + WriteExt;
+    type AsyncIoContext;
+
+    fn async_read(
```
tabokie, line 21: Since all the async happens inside the implementation, we can rename this to something like RocksDB's MultiGet, i.e. multi_read.

src/env/mod.rs
```diff
@@ -10,11 +10,21 @@ mod obfuscated;
 pub use default::DefaultFileSystem;
 pub use obfuscated::ObfuscatedFileSystem;
 
+use crate::pipe_log::FileBlockHandle;
 /// FileSystem
 pub trait FileSystem: Send + Sync {
     type Handle: Send + Sync + Handle;
     type Reader: Seek + Read + Send;
     type Writer: Seek + Write + Send + WriteExt;
+    type AsyncIoContext;
```
tabokie, line 19: This to MultiReadContext.

src/engine.rs
```diff
@@ -683,6 +742,41 @@ mod tests {
                 reader(e.index, entry_index.entries.unwrap().id.queue, &e.data);
             }
         }
+        fn scan_entries_aio<FR: Fn(u64, LogQueue, &[u8])>(
```
tabokie, line 745: Add a newline between functions.

src/engine.rs
```diff
@@ -308,11 +308,45 @@ where
...
     }
 
+    pub fn fetch_entries_to_aio<M: Message + MessageExt<Entry = M>>(
+        &self,
+        region_id: u64,
+        begin: u64,
+        end: u64,
+        max_size: Option<usize>,
+        vec: &mut Vec<M::Entry>,
+    ) -> Result<usize> {
```
tabokie, line 324: I think based on the tests, you can antomatically select single_read or multi_read and avoid creating two different engine methods, e.g. use aio when blocks.len() > 4 or something.

One issue though is that I'm not sure if aio syscall is portable enough. You might need to do some research on how to detect if aio is available (maybe take a look at how RocksDB did it).

ustc-wxy: OK. Let me try.

Issue #252, opened 2022-07-25
introduce file manifest


tabokie:
Memtable recovery logic is over-complicated because we have to handle intermediate states during GC. It would be much easier if we could maintain the file states in a persistent manifest.

tabokie:
Alternatively, we could add an "epoch" field to the log file name. During recovery, a file named "n.e.log" shadows all files named "k.epoch.log" where `k <= n && epoch < e`.

PR #289, opened 2023-01-09
fix data loss after partial rewrite-rewrite

### LINKED ISSUE #288: aborted rewrite-rewrite causes data loss
Say we have the data:
```
append: [11, 15]
rewrite: [1, 10]
```
When doing rewrite on the whole rewrite queue, it panicked. The old rewrite queue is not cleaned up and only partial output is written:
```
append: [11, 15]
rewrite: [1. 10] [1, 5]
```
After restart, only logs from `[11, 15]` will be recovered.

tabokie:
Fix #288 by adding a file marker during rewrite.

A more elegant fix would be https://github.com/tikv/raft-engine/issues/252, but that requires some radical architectural changes.

### DIFF
--- a/src/file_pipe_log/pipe.rs
+++ b/src/file_pipe_log/pipe.rs
@@ -15,8 +15,7 @@ use crate::env::FileSystem;
 use crate::event_listener::EventListener;
 use crate::metrics::*;
 use crate::pipe_log::{
-    FileBlockHandle, FileId, FileSeq, LogFileContext, LogQueue, PipeLog, PipeSnapshot,
-    ReactiveBytes,
+    FileBlockHandle, FileId, FileSeq, LogFileContext, LogQueue, PipeLog, ReactiveBytes,
 };
 use crate::{perf_context, Error, Result};
 
@@ -478,12 +477,7 @@ impl<F: FileSystem> SinglePipe<F> {
         Ok((current.first_seq_in_use - prev.first_seq_in_use) as usize)
     }
 
-    fn snapshot(&self) -> (FileSeq, usize) {
-        let active_file = self.active_file.lock();
-        (active_file.seq, active_file.writer.offset())
-    }
-
-    fn restore(&self, file_seq: FileSeq, offset: usize) -> Result<()> {
+    fn truncate(&self, file_seq: FileSeq) -> Result<()> {
         let mut active_file = self.active_file.lock();
         let mut files = self.files.write();
         if file_seq < files.first_seq_in_use {
@@ -510,11 +504,9 @@ impl<F: FileSystem> SinglePipe<F> {
                 .delete(file_id.build_file_path(&self.dir))?;
             self.sync_dir()?;
             files.fds.pop_back();
-            *active_file = new_active_file;
-
             self.flush_metrics(files.fds.len());
+            *active_file = new_active_file;
         }
-        active_file.writer.truncate(Some(offset))?;
         Ok(())
     }
 }
@@ -591,17 +583,8 @@ impl<F: FileSystem> PipeLog for DualPipes<F> {
     }
 
     #[inline]
-    fn snapshot(&self, queue: LogQueue) -> PipeSnapshot {
-        let (seq, offset) = self.pipes[queue as usize].snapshot();
-        PipeSnapshot {
-            file_id: FileId::new(queue, seq),
-            offset,
-        }
-    }
-
-    #[inline]
-    fn restore(&self, snapshot: PipeSnapshot) -> Result<()> {
-        self.pipes[snapshot.file_id.queue as usize].restore(snapshot.file_id.seq, snapshot.offset)
+    fn truncate(&self, file_id: FileId) -> Result<()> {
+        self.pipes[file_id.queue as usize].truncate(file_id.seq)
     }
 }
 
--- a/src/file_pipe_log/pipe.rs
+++ b/src/file_pipe_log/pipe.rs
@@ -15,7 +15,8 @@ use crate::env::FileSystem;
 use crate::event_listener::EventListener;
 use crate::metrics::*;
 use crate::pipe_log::{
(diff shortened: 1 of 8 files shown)


src/purge.rs
```diff
@@ -41,6 +51,68 @@ where
...
+    rewrite_marker_path.push(REWRITE_MARKER);
+    if let Some(seq) = read_marker(&rewrite_marker_path)? {
+        let mut f = FileId::new(LogQueue::Rewrite, seq + 1);
+        let mut paths = Vec::new();
+        loop {
+            let path = f.build_file_path(&cfg.dir);
+            if path.exists() {
+                paths.push(path);
+            } else {
+                break;
```
BusyJay, line 102: What if some files are deleted while some not?

src/purge.rs
```diff
@@ -71,6 +146,10 @@ where
             warn!("Unable to purge expired files: locked");
             return Ok(vec![]);
         }
+        if self.rewrite_marker_path.exists() {
+            // Otherwise all writes will be erased next reboot.
+            remove_marker(&self.rewrite_marker_path)?;
```
BusyJay, line 151: Why it's safe to remove it? Or why it exists in the first place?

src/purge.rs
```diff
@@ -41,6 +51,68 @@ where
...
+
+pub fn before_recover(cfg: &Config, fs: &impl crate::env::FileSystem) -> Result<()> {
+    use crate::file_pipe_log::FileNameExt;
+    let mut rewrite_marker_path = PathBuf::from(&cfg.dir);
+    rewrite_marker_path.push(REWRITE_MARKER);
+    if let Some(seq) = read_marker(&rewrite_marker_path)? {
+        let mut f = FileId::new(LogQueue::Rewrite, seq + 1);
+        let mut paths = Vec::new();
+        loop {
+            let path = f.build_file_path(&cfg.dir);
```
BusyJay, line 98: What if new appends are flushed into a new file before purge_expires_files finished?

src/purge.rs
tabokie: The engine will fail to open, next time it will redo the cleanup.

src/purge.rs
tabokie: The last `purge` fails to remove it, subsequent purge will never write to rewrite queue unless the marker is removed here. (it's guaranteed that only purge will write to rewrite queue).

src/purge.rs
tabokie: This is called before opening engine, no one can write.

src/purge.rs
BusyJay: How? It will break the loop becasue of those deleted files.

src/purge.rs
BusyJay: But the file is created during rewrite, right? And rewrite can happen anytime?

src/purge.rs
tabokie: Oh, here I only consider the case that purge (aka rewrite) succeeds but fails to remove the marker. I'll need to do log file clean up here as well to make it work.

As for your question, a failed `purge` doesn't necessarily mean the whole engine is shutdown. User can still call purge over it, he just has to wait until the cleanup is finished here.

src/purge.rs
tabokie: The files are deleted in reverse, it doesn't leave hole if any one file fails to be deleted. So we can always redo as long as the marker is still there.

src/purge.rs
BusyJay: How to know it's the case that "rewrite succeeds but fail to delete file" or "rewrite fails and fail to delete file"?

src/purge.rs
tabokie: I still don't understand your case.

src/purge.rs
tabokie: We will always treat it as the latter. The code is updated to protect the entire purge operation as an atomic part.

src/file_pipe_log/pipe.rs
```diff
@@ -476,6 +477,46 @@ impl<F: FileSystem> SinglePipe<F> {
...
+                )?,
+                format: new_active_fd.format,
+            };
+            let file_id = FileId {
+                queue: self.queue,
+                seq: active_file.seq,
+            };
+            self.file_system
+                .delete(file_id.build_file_path(&self.dir))?;
+            self.sync_dir()?;
```
BusyJay, line 511: Why not sync after all are deleted?

src/file_pipe_log/pipe.rs
```diff
@@ -476,6 +477,46 @@ impl<F: FileSystem> SinglePipe<F> {
...
+            files.fds.pop_back();
+            *active_file = new_active_file;
+
+            self.flush_metrics(files.fds.len());
```
BusyJay, line 515: Why not flush after all are deleted?

BusyJay (commented):
I'm afraid this patch is not backward compatible. For example, if TiKV is downgraded to the old version and then upgrade to the new version again, new writes in old version may be truncated by new version.

src/file_pipe_log/pipe.rs
tabokie: If there's a hole between files, recovery will discard older files. Here we must make sure each file is deleted before moving forward, to avoid leaving hole.

BusyJay: If I delete a, and then delete b in the same directory, is it possible that b is actually deleted while a still exists? On the other hand, maybe recovery should keep deleting files even some are missing instead of breaking the looop.

tabokie: I believe it is possible.
> should keep deleting files even some are missing instead of breaking the looop.

It will cause data loss. Assuming b(seq=10) is deleted but a(seq=11) is not, on next reboot all files with seq<11 will be discarded.

PR #290, opened 2023-01-12
fix data loss of partial rewrite with atomic group

### LINKED ISSUE #288: aborted rewrite-rewrite causes data loss

tabokie:
A second attempt to fix #288.

This PR introduces the concept of "atomic group". It is then used by rewrite-rewrite operation to make sure the rewrite of each region is perceived as an atomic operation.

A group of writes is made atomic by each carrying a special marker. During recovery, log batch with the marker will be stashed until all parts of the group are found. Caveats for this approach is commented near `AtomicGroupBuilder`.

Also fixed a bug that a partial rewrite-rewrite (due to batch being split) is not applied correctly.


BusyJay:
Is it still compatible with old version?

src/file_pipe_log/pipe_builder.rs
```diff
@@ -214,7 +214,10 @@ impl<F: FileSystem> DualPipesBuilder<F> {
         &mut self,
         machine_factory: &FA,
     ) -> Result<(M, M)> {
-        let threads = self.cfg.recovery_threads;
+        let threads = std::cmp::min(
```
BusyJay, line 217: What if it's 0?

src/lib.rs
```diff
@@ -160,6 +161,19 @@ impl GlobalStats {
     }
 }
 
+const INTERNAL_KEY_PREFIX: &[u8] = b"__";
+
+pub(crate) fn make_internal_key(k: &[u8]) -> Vec<u8> {
```
BusyJay, line 166: What if it's used outside the crate?

src/log_batch.rs
```diff
@@ -932,6 +941,114 @@ fn verify_checksum_with_signature(buf: &[u8], signature: Option<u32>) -> Result<
...
+impl AtomicGroupBuilder {
+    /// Each log batch can only carry one atomic group marker. If multiple are
+    /// present only the first is recognized.
+    pub fn begin(&mut self, lb: &mut LogBatch) {
+        fail::fail_point!("atomic_group::begin");
+        assert!(!self.begin);
+        let mut s = Vec::with_capacity(ATOMIC_GROUP_VALUE_LEN);
+        s.encode_u64(self.id).unwrap();
+        s.push(AtomicGroupStatus::Begin as u8);
+        lb.put(self.id, crate::make_internal_key(ATOMIC_GROUP_KEY), s);
```
BusyJay, line 1025: What if it's conflict with region ID?

src/log_batch.rs
```diff
@@ -932,6 +941,114 @@ fn verify_checksum_with_signature(buf: &[u8], signature: Option<u32>) -> Result<
...
+        self.begin = true;
+    }
+
+    pub fn add(&self, lb: &mut LogBatch) {
+        fail::fail_point!("atomic_group::add");
+        assert!(self.begin);
+        let mut s = Vec::with_capacity(ATOMIC_GROUP_VALUE_LEN);
+        s.encode_u64(self.id).unwrap();
+        s.push(AtomicGroupStatus::Middle as u8);
+        lb.put(self.id, crate::make_internal_key(ATOMIC_GROUP_KEY), s);
```
BusyJay, line 1035: Since the id is already a part of key, why encode it in value again?

src/log_batch.rs
```diff
@@ -932,6 +941,114 @@ fn verify_checksum_with_signature(buf: &[u8], signature: Option<u32>) -> Result<
...
+    }
+
+    pub fn end(&self, lb: &mut LogBatch) {
+        assert!(self.begin);
```
BusyJay, line 1039: `begin` can be reset to false for robust.

src/file_pipe_log/pipe_builder.rs
tabokie: It's checked in `Config::sanitize`.

src/lib.rs
tabokie: It will behave like a normal key, until:
(1) restart, the key will disappear from memory
(2) rewritten, the key will both memory and disk

src/log_batch.rs
tabokie: The key will be filtered out during apply memtable, and recover memtable.

src/log_batch.rs
tabokie: It is not supposed to be reused. Initially the signature is `fn end(self, lb)`, but it makes the code a bit messier.

src/log_batch.rs
tabokie: Good catch.

src/log_batch.rs
BusyJay: Then you can use an enum.

src/lib.rs
BusyJay: My point is we should forbid users use such keys.

src/file_pipe_log/pipe_builder.rs
tabokie: Then nothing needs to be recovered, our code handles zero gracefully, but I noticed rayon thread pool will treat 0 as default value though.

src/lib.rs
tabokie: The overhead should be opt-in, if user is unsure he can use `is_internal_key` as check.

BusyJay: It's about correctness not about performance. Checking if user writes an internal key is cheap as only prefix (2 bytes) needs to be checked.

CHANGELOG.md
```diff
@@ -5,6 +5,11 @@
 ### Behavior Changes
 
 * Disable log recycling by default.
+* `LogBatch::put` returns a `Result<()>` instead of `()`. It errs when the key is reserved for internal use.
```
BusyJay, line 8: How about panicking inside to make API clean? As it's a static format and user can do nothing except panic.

src/purge.rs
```diff
@@ -323,32 +332,56 @@ where
...
-                    cursor = 0;
-                } else {
-                    cursor += 1;
+            // Split the entries into smaller chunks, so that we don't OOM, and the
+            // compression overhead is not too high.
+            let mut chunks = vec![Vec::new()];
+            let mut current_size = 0;
+            debug_assert!(log_batch.approximate_size() <= max_batch_bytes());
+            for ei in entry_indexes {
+                current_size += ei.entry_len as usize;
```
BusyJay, line 341: What's the difference between `entry.len()` and `ei.entry_len`?

src/purge.rs
```diff
@@ -323,32 +332,56 @@ where
...
-                    std::mem::swap(&mut take_entries, &mut entries);
-                    let mut take_entry_indexes = entry_indexes.split_off(cursor + 1);
-                    std::mem::swap(&mut take_entry_indexes, &mut entry_indexes);
-                    log_batch.add_raw_entries(region_id, take_entry_indexes, take_entries)?;
-                    self.rewrite_impl(&mut log_batch, rewrite, false)?;
-                    total_size = 0;
-                    cursor = 0;
-                } else {
-                    cursor += 1;
+            // Split the entries into smaller chunks, so that we don't OOM, and the
```
BusyJay, line 335: Can following logic be implemented as two functions depending on whether `atomic_group` is used? I find it quite hard to understand.

src/log_batch.rs
```diff
@@ -932,6 +958,122 @@ fn verify_checksum_with_signature(buf: &[u8], signature: Option<u32>) -> Result<
...
+        ));
+        self.status = Some(AtomicGroupStatus::Middle);
+        self.flush(lb);
+    }
+
+    pub fn end(&mut self, lb: &mut LogBatch) {
+        assert!(matches!(
+            self.status.unwrap(),
+            AtomicGroupStatus::Begin | AtomicGroupStatus::Middle
+        ));
```
BusyJay, line 1059:
```suggestion
        assert!(matches!(
            self.status,
            Some(AtomicGroupStatus::Begin | AtomicGroupStatus::Middle)
        ));
```

CHANGELOG.md
tabokie: In this crate panic is only used for unrecoverable error. For your case user can choose to bubble the error to UI for example.

src/purge.rs
tabokie: They should be the same, the `read_entry_bytes_from_file` makes sure of it.

src/purge.rs
tabokie: Because use_atomic_group=true doesn't necessarily mean atomic_group.is_some(), most of the logic cannot be simplified for the use_atomic_group=true case.

And for the atomic_group.is_some() case, I think the logic is relatively simple (only involves one branch).

CHANGELOG.md
BusyJay: I'm afraid not. It's used for raft log, it's an undefined error if continue processing while last logs failed to be written.

src/log_batch.rs
```diff
@@ -932,6 +958,122 @@ fn verify_checksum_with_signature(buf: &[u8], signature: Option<u32>) -> Result<
...
+        if let LogItemContent::Kv(KeyValue {
+            op_type,
+            key,
+            value,
+            ..
+        }) = &item.content
+        {
+            if *op_type == OpType::Put
+                && value.as_ref().unwrap().len() == ATOMIC_GROUP_VALUE_LEN
+                && *key == crate::make_internal_key(ATOMIC_GROUP_KEY)
```
BusyJay, line 988: How about avoiding allocation?

CHANGELOG.md
tabokie: It's a log batch method, not an engine method. Even if it's an engine method, returning an error means this operation can be retried by user without causing damage to existing data.

src/memtable.rs
```diff
@@ -1176,9 +1213,64 @@ impl<A: AllocatorTrait> MemTableRecoverContext<A> {
 
     pub fn merge_append_context(&self, append: MemTableRecoverContext<A>) {
         self.memtables
-            .apply_append_writes(append.log_batch.clone().drain());
+            .apply_append_writes(append.tombstone_items.into_iter());
         self.memtables.merge_append_table(append.memtables);
     }
+
+    #[inline]
+    fn is_tombstone(item: &LogItem) -> bool {
```
BusyJay, line 1221: What does `is_tombstone` mean?

src/memtable.rs
```diff
@@ -1176,9 +1213,64 @@ impl<A: AllocatorTrait> MemTableRecoverContext<A> {
...
+        assert_eq!(queue, LogQueue::Rewrite);
+        if let Some(groups) = self.pending_atomic_groups.get_mut(&id) {
+            let group = groups.last_mut().unwrap();
+            match (group.status, new_group.status) {
+                (AtomicGroupStatus::End, AtomicGroupStatus::Begin) => {
+                    groups.push(new_group);
+                }
+                // (begin, begin), (middle, begin)
+                (_, AtomicGroupStatus::Begin) => {
+                    warn!("discard atomic group: {group:?}");
```
BusyJay, line 1240: Why just discard?

src/memtable.rs
```diff
@@ -1176,9 +1213,64 @@ impl<A: AllocatorTrait> MemTableRecoverContext<A> {
...
+                    *group = new_group;
+                }
+                // (end, middle), (end, end)
+                (AtomicGroupStatus::End, _) => {
+                    warn!("discard atomic group: {new_group:?}");
+                }
+                (AtomicGroupStatus::Begin, AtomicGroupStatus::Middle)
+                | (AtomicGroupStatus::Middle, AtomicGroupStatus::Middle) => {
```
BusyJay, line 1248: So if the groups is `Begin, Middle, Middle, End`, the processing order will be `merge (Begin, Middle) -> merge(Begin, Middle) -> merge(Begin, End)`?

tabokie: Yes. Begin and End are like left and right brackets.

src/memtable.rs
tabokie: The left hand side group is incomplete, we will assume it is caused by a power off.

src/memtable.rs
tabokie: Deletion marks like Compact or Clean. They need to be passed down to older files to fully delete history data.

src/memtable.rs
BusyJay: Then how can `merge(Middle, End)` happen?

tabokie: The replay is done in parallel. It's possible an atomic group is replayed be two different threads, then merged together in the end.

BusyJay: This behavior should be commented.

src/memtable.rs
```diff
@@ -1189,30 +1281,69 @@ impl Default for MemTableRecoverContext<VacantAllocator> {
...
+            let mut new_tombstones = Vec::new();
+            let mut is_group = None;
+            let items = item_batch
+                .drain()
+                .filter(|item| {
+                    if let Some(g) = AtomicGroupStatus::parse(item) {
+                        if is_group.is_none() {
+                            is_group = Some(g);
+                        } else {
+                            debug_assert!(false, "skipped an atomic group: {:?}", g);
```
BusyJay, line 1304: Should log.

src/memtable.rs
```diff
@@ -1189,30 +1281,69 @@ impl Default for MemTableRecoverContext<VacantAllocator> {
...
-            LogQueue::Rewrite => self.memtables.replay_rewrite_writes(item_batch.drain()),
-        }
         Ok(())
     }
 
     fn merge(&mut self, mut rhs: Self, queue: LogQueue) -> Result<()> {
-        self.log_batch.merge(&mut rhs.log_batch.clone());
+        self.tombstone_items
+            .append(&mut rhs.tombstone_items.clone());
+        for (id, groups) in rhs.pending_atomic_groups.drain() {
```
BusyJay, line 1335: Is it possible the merge is not done continuously? For example, (Begin, Middle), (Middle, Middle), (Middle, End) are merged as (Middle, Middle), (Middle, End), (Begin, Middle).

src/purge.rs
```diff
@@ -323,31 +333,61 @@ where
...
-                    std::mem::swap(&mut take_entries, &mut entries);
-                    let mut take_entry_indexes = entry_indexes.split_off(cursor + 1);
-                    std::mem::swap(&mut take_entry_indexes, &mut entry_indexes);
-                    log_batch.add_raw_entries(region_id, take_entry_indexes, take_entries)?;
-                    self.rewrite_impl(&mut log_batch, rewrite, false)?;
-                    total_size = 0;
-                    cursor = 0;
-                } else {
-                    cursor += 1;
+            let mut alien_size = log_batch.approximate_size();
```
BusyJay, line 336: What is "alien_size"?

src/purge.rs
BusyJay: I tried the second time, still found the code hard to read and lost in context switch.

tabokie: 🤷 I have refactored it several times too, I guess it's just that complex.

src/purge.rs
tabokie: Doesn't belong to current raft group.

src/memtable.rs
tabokie: No, it's guaranteed to merge neighbors.

src/memtable.rs
tabokie: The parallel replay part is commented near `trait ReplayMachine` in file_pipe_log/pipe_builder.rs.

BusyJay (approved):

PR #294, opened 2023-02-15
Auxiliary directory for supporting the multi-directory configuration.

### LINKED ISSUE #257: Support secondary log directory

LykxSassinator:
## Brief Introduction
Close #257 .

This pr is used to support the `auxiliary directory` feature, referring to the previous pr: #261.
It mainly contains:
* Add an extra configuration `auxiliary-dir` to enable this feature. Default is `None`, and if set with `Some(...)`, this directory will be used if the main dir is full, specified by `dir` option.

IMO, `auxiliary-dir` is more conspicuous to understand and use for users, compared to the previous name `secondary-dir`.


LykxSassinator:
ping @tabokie PTAL, thx. This pr is a successor to the obsolete #261.

src/log_batch.rs
```diff
@@ -497,6 +503,7 @@ impl LogItemBatch {
         }
         let checksum = crc32(&buf[offset..]);
         buf.encode_u32_le(checksum)?;
```
tabokie, line 505: Don't encode, leave a placeholder here, and comment it's delayed to prepare_write.

src/log_batch.rs
```diff
@@ -419,17 +422,20 @@ impl LogItemBatch {
     /// The `signature` is both generated by the given `LogFileContext`.
     /// That is, the final checksum of each `LogBatch` consists of this
     /// `signature` and the original `checksum` of the contents.
-    pub(crate) fn prepare_write(buf: &mut Vec<u8>, file_context: &LogFileContext) -> Result<()> {
+    pub(crate) fn prepare_write(
```
tabokie, line 425: `fn prepare_write(&self, buf: &[u8], ctx)`

src/log_batch.rs
```diff
@@ -378,6 +380,7 @@ impl LogItemBatch {
     }
 
     pub fn drain(&mut self) -> LogItemDrain {
+        self.checksum = 0;
```
tabokie, line 383: Move this to the last, the same order as variable definition.

src/file_pipe_log/pipe.rs
```diff
@@ -23,14 +23,17 @@ use super::format::{build_recycled_file_name, FileNameExt, LogFileFormat};
 use super::log_file::build_file_reader;
 use super::log_file::{build_file_writer, LogFileWriter};
 
+pub type PathId = usize;
+pub type Paths = Vec<PathBuf>;
+
+/// Main directory path id.
 pub const DEFAULT_PATH_ID: PathId = 0;
+/// Auxiliary directory path id.
+pub const AUXILIARY_PATH_ID: PathId = 1;
```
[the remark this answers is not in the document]
LykxSassinator: I hold the different perspective on this point.
+ Firstly, the original design to this is to support an **EXTRA** dir to store logs  if and only if the main dir is full, ideally placed on another **DEVICE**. It serves the target that the directory of storing raft logs could be placed on an individual device to make IOs individual to other dirs' IO operations. 
+ Secondly, in normal cases, two directories, specified by `dir` and `auxiliary_dir`, are enough for the practical production envs. And we do not expect to support 3,4, or more dir lists for users.

So, I still recommend to use this design.

src/file_pipe_log/pipe.rs
```diff
@@ -479,6 +513,34 @@ impl<F: FileSystem> PipeLog for DualPipes<F> {
...
+    let force_no_spare_space = || {
+        fail_point!("file_pipe_log::force_no_spare_space", |_| true);
+        false
+    };
+    for t in [DEFAULT_PATH_ID, AUXILIARY_PATH_ID] {
+        let idx = t as usize;
+        if idx >= paths.len() {
+            break;
+        }
+        let disk_stats = match fs2::statvfs(&paths[idx]) {
```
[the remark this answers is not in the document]
LykxSassinator: IMO, we need `Err`, which means that there exists no free space for generating new files.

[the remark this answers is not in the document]
tabokie: Two reasons: (1) `statvfs` is only a hint. There could be another user deleting files or creating files. The result of `fetch_dir` doesn't promise anything at all. And it is false to rely on it. (2) All the error handling logic is already in place, adding another code path that handles the error of `fetch_dir` is redundant. Even if they (`fetch_dir().is_err()` and `write().is_err()`) share the same code path, it is always superior to have fewer exit points.

src/file_pipe_log/pipe.rs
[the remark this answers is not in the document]
tabokie: It is contradicting the whole point of using a vector as paths. If you need to hardcode the path ID, it is clearly better to just use two variables `dir1` and `dir2`, because it checks the code statically instead of runtime (via vector indexing). And my originally idea of using a vector is to not restricting the amount of directories. For example, if in the future we would support #258, user can configure multiple disks to increase the write parallelism.

src/file_pipe_log/pipe_builder.rs
```diff
@@ -100,141 +102,136 @@ impl<F: FileSystem> DualPipesBuilder<F> {
...
+                } else if f.seq < current_seq {
+                    return Err(Error::InvalidArgument("Duplicate file".to_string()));
+                }
+            }
+            files.drain(..invalid_files);
+            // Try to cleanup stale metadata left by the previous version.
+            if files.is_empty() {
+                continue;
+            }
+            let mut cleared = 0_u64;
```
tabokie, line 145: I don't think you fully understand the old code. The new implementation is really bad.

src/file_pipe_log/pipe_builder.rs
```diff
@@ -100,141 +102,136 @@ impl<F: FileSystem> DualPipesBuilder<F> {
...
+            (LogQueue::Append, &mut self.recycled_files, true),
+        ] {
+            if files.is_empty() {
+                continue;
+            }
+            // Check the file_list and remove the hole of files.
+            let mut current_seq = files[0].seq;
+            let mut invalid_files = 0_usize;
+            debug_assert!(current_seq > 0);
+            for (i, f) in files.iter().enumerate() {
```
tabokie, line 128:
```
for (i, (f, next_f)) in files.windows(2).enumerate() {
  if f.seq == next_f.seq {
  } else if f.seq + 1 < next_f.seq {
    to_be_removed = i;
  }
}
```

src/file_pipe_log/pipe_builder.rs
```diff
@@ -100,141 +102,136 @@ impl<F: FileSystem> DualPipesBuilder<F> {
...
+                if f.seq > current_seq + (i - invalid_files) as u64 {
+                    warn!(
+                        "Detected a hole when scanning directory, discarding files before file_seq {}.",
+                        f.seq,
+                    );
+                    current_seq = f.seq + 1;
+                    invalid_files = i;
+                } else if f.seq < current_seq {
+                    return Err(Error::InvalidArgument("Duplicate file".to_string()));
```
tabokie, line 137: Not covered.

src/file_pipe_log/pipe_builder.rs
```diff
@@ -461,14 +460,28 @@ impl<F: FileSystem> DualPipesBuilder<F> {
                 while written < target_file_size {
                     writer.write_all(&buf).unwrap_or_else(|e| {
                         warn!("failed to build recycled file, err: {}", e);
+                        no_enough_space = true;
```
tabokie, line 463: Use a function `is_out_of_space(e: IoError)`.

src/file_pipe_log/pipe_builder.rs
```diff
@@ -503,13 +518,14 @@ impl<F: FileSystem> DualPipesBuilder<F> {
         )?;
         let rewriter = SinglePipe::open(
             &self.cfg,
+            self.dirs.clone(),
             self.file_system.clone(),
             self.listeners.clone(),
```
tabokie, line 523: These clone seem unnecessary. I wonder why clippy didn't warn about them.

src/file_pipe_log/pipe_builder.rs
```diff
@@ -461,14 +460,28 @@ impl<F: FileSystem> DualPipesBuilder<F> {
                 while written < target_file_size {
                     writer.write_all(&buf).unwrap_or_else(|e| {
                         warn!("failed to build recycled file, err: {}", e);
+                        no_enough_space = true;
                     });
                     written += buf.len();
                 }
+                if no_enough_space {
```
tabokie, line 467: Let's play it safe, there's little benefit to have partially prefilled files anyway:
```
if no_space {
  target = 0;
}
```

src/log_batch.rs
```diff
@@ -419,18 +422,22 @@ impl LogItemBatch {
...
             if let Some(signature) = file_context.get_signature() {
-                // Insert the signature into the encoded bytes. Rewrite checksum of
-                // `LogItemBatch` in `LogBatch`.
-                let footer_checksum_offset = buf.len() - LOG_BATCH_CHECKSUM_LEN;
-                let original_checksum = codec::decode_u32_le(&mut &buf[footer_checksum_offset..])?;
                 // The final checksum is generated by `signature` ***XOR***
                 // `original checksum of buf`.
-                (&mut buf[footer_checksum_offset..])
-                    .write_u32::<LittleEndian>(original_checksum ^ signature)?;
+                footer_checksum ^= signature;
```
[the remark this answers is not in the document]
LykxSassinator: No modification on `self.checksum`.
`self.checksum` just caches the local footer checksum of `LogItemBatch`. The final signature just use the `self.checksum ^ signature`. And on the 2nd time, when `engine.write(...)` calls `prepare_write(...)` again, it will be replaced by `self.checksum ^ signature2`.

src/engine.rs
```diff
@@ -142,7 +144,14 @@ where
         let start = Instant::now();
         let len = log_batch.finish_populate(self.cfg.batch_compression_threshold.0 as usize)?;
         debug_assert!(len > 0);
-        let block_handle = {
+
+        let mut attempt_count = 0_u64;
+        let block_handle = loop {
```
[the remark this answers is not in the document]
LykxSassinator: I've tried this format before, but it didn't work, as the syntax `break block_handle` was invalid under the `for` loop.
FYI, from rust syntax:
```
`break` with value from a `for` loop can only break with a value inside `loop` or breakable block.
...
you can't `break` with a value in a `for` loop,
use `break` on its own without a value inside this `for` loop: `break`
```

src/log_batch.rs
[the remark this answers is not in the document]
tabokie: Okay, you need to change the signature from `&mut self` to `&self`.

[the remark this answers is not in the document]
LykxSassinator: > Okay, you need to change the function signature from `&mut self` to `&self`.

Yep, I found that I've missed it.

src/config.rs
```diff
@@ -26,11 +26,20 @@ pub enum RecoveryMode {
...
     pub dir: String,
 
+    /// Auxiliary directory to store log files. Will create on startup if
+    /// set and not exists.
+    ///
+    /// Newly logs will be put into this dir when the main `dir` is full
+    /// or not accessible.
+    ///
+    /// Default: None
+    pub auxiliary_dir: Option<String>,
```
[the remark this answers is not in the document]
LykxSassinator: Let me think about it.

src/engine.rs
```diff
@@ -142,7 +144,14 @@ where
...
+            // Max retry count is limited to `WRITE_MAX_RETRY_TIMES`, that is, 2.
+            // If the first `append` retry because of NOSPC error, the next `append`
+            // should success, unless there exists several abnormal cases in the IO device.
+            // In that case, `Engine::write` must return `Err`.
+            attempt_count += 1;
             let mut writer = Writer::new(log_batch, sync);
```
tabokie, line 155: The `ENGINE_WRITE_PREPROCESS_DURATION_HISTOGRAM.observe` needs to be adjusted to only observe at the first attempt.

src/file_pipe_log/pipe.rs
```diff
@@ -302,6 +304,20 @@ impl<F: FileSystem> SinglePipe<F> {
                     seq, e, te
                 );
             }
+            if is_no_space_err(&e) {
```
tabokie, line 307: Let's just make all methods of `LogFileWriter` returns io::Result instead.

src/errors.rs
```diff
@@ -30,3 +32,19 @@ pub enum Error {
 }
 
 pub type Result<T> = ::std::result::Result<T, Error>;
+
+/// Check whether the given error is a nospace error.
+pub(crate) fn is_no_space_err(e: &IoError) -> bool {
+    // TODO: make the following judgement more elegant when the error type
+    // `ErrorKind::StorageFull` is stable.
+    if_chain::if_chain! {
+        if let Some(err) = e.get_ref();
```
tabokie, line 41: I think this step is unnecessary?

src/file_pipe_log/log_file.rs
```diff
@@ -66,21 +66,23 @@ impl<F: FileSystem> LogFileWriter<F> {
...
 
-    fn write_header(&mut self, format: LogFileFormat) -> Result<()> {
+    fn write_header(&mut self, format: LogFileFormat) -> IoResult<()> {
         self.writer.seek(SeekFrom::Start(0))?;
         self.written = 0;
         let mut buf = Vec::with_capacity(LogFileFormat::encoded_len(format.version));
-        format.encode(&mut buf)?;
+        format
+            .encode(&mut buf)
+            .map_err(|e| std::io::Error::new(std::io::ErrorKind::Other, e))?;
```
tabokie, line 75: Nah, this isn't right. Just unwrap it for now.

src/file_pipe_log/pipe.rs
```diff
@@ -147,14 +149,14 @@ impl<F: FileSystem> SinglePipe<F> {
 
     /// Synchronizes all metadatas associated with the working directory to the
     /// filesystem.
-    fn sync_dir(&self) -> Result<()> {
+    fn sync_dir(&self, path_id: PathId) -> Result<()> {
         debug_assert!(!self.paths.is_empty());
-        let path = PathBuf::from(&self.paths[0]);
+        let path = PathBuf::from(&self.paths[path_id]);
         std::fs::File::open(path).and_then(|d| d.sync_all())?;
```
tabokie, line 155:
```suggestion
        std::fs::File::open(&self.paths[path_id]).and_then(|d| d.sync_all())?;
```

src/file_pipe_log/pipe.rs
```diff
@@ -295,14 +301,36 @@ impl<F: FileSystem> SinglePipe<F> {
...
+        if let Err(e) = writer.write(buf_bytes, self.target_file_size) {
             if let Err(te) = writer.truncate() {
                 panic!(
                     "error when truncate {} after error: {}, get: {}",
                     seq, e, te
                 );
             }
-            return Err(e);
+            if is_no_space_err(&e) {
+                // To avoid incorrect `rotate_imp` on the same dir when the size of
```
tabokie, line 313: I don't understand. If the available space is less than `buf_bytes.len()`, it should panic when actually writing that `buf_bytes`, instead of panicking here too early.

LykxSassinator: In this case, it won't panic but fail on `engine.write(...)` unexpectedly.
Given that we have a two directories setting with `dir` and `spill-dir`. If `dir` has free space for rotating several new log files, but no enough space for dumping the given `buf`. And `spill-dir` has free space for dumping this `buf`.
That is: 
+ `buf_bytes.len()` > `target_file_size` == true;
+ `n * target_file_size` < `dir.spare_space()` && `buf_bytes.len() > dir.spare_space()` && `buf_bytes.len() < spill-dir.spare_space()`. 

Then, with the previous `rotate_imp(self.target_file_size)`,
+ 1. First attempt for writing, failed on `self.write()` and try to rotate.
+ 2. As `target_file_size` < `dir.spare_space()`, `rotate_imp(target_file_size)` successfully, and returned `TryAgain(...)` to trigger the second attempt for write. (Naming this newly rotated file as `rotate-1`.)
+ 3. Secondly, it tried to dump this buf into `rotate-1`. But `self.write()` failed again as the `dir.spare_space()` < `buf_bytes.len()`, and it returned the `NOSPAC` error. Buffer still could not be dumped into current `rotate-1` file. Then, it made the next `rotate_imp()`. And unfortunately, it succeed again because the `dir` still had enough space for rotating a new file. So, it returned `TryAgain()` error again. It would cause the final failure on `engine.write(...)`.
+ 4. The whole write of this LogBatch failed because its trying times on write exceed the trying limitation. However, the `spill-dir` still had enough space for it.

For example, given that, `target_file_size == 1MB`, `dir.spare_space() == 1.5MB`, `spill-dir.spare_space() == 100MB` and `buf_bytes.len() == 3MB`. The above steps would be executed.
```Rust
   Compiling raft-engine v0.3.0 (/nvme1n1/lucasliang/workspace/raft-engine)
    Finished test [unoptimized + debuginfo] target(s) in 13.33s
     Running tests/failpoints/mod.rs (target/debug/deps/failpoints-23610de2336687d7)
thread 'test_engine::test_build_engine_with_spill_dir' panicked at 'called `Result::unwrap()` on an `Err` value: 
TryAgain("Failed to write logbatch, exceed max_retry_count: (2),
err: failed to write 41 file, get IO Error: Custom { kind: StorageFull, error: \"nospace\" } try to flush it to another dir")', 
tests/failpoints/test_engine.rs:37:40
```

src/errors.rs
LykxSassinator: Yep.

src/file_pipe_log/pipe_builder.rs
```diff
@@ -100,143 +103,149 @@ impl<F: FileSystem> DualPipesBuilder<F> {
...
+            self.dirs.push(dir.to_path_buf());
+            return Ok(());
+        }
+        if !dir.is_dir() {
+            return Err(box_err!("Not directory: {}", dir.display()));
+        }
+        self.dir_locks.push(lock_dir(dir)?);
+        self.dirs.push(dir.to_path_buf());
+        assert_eq!(self.dirs.len(), self.dir_locks.len());
+        assert!(!self.dirs.is_empty());
```
tabokie, line 204: Useless assertion.

src/file_pipe_log/pipe_builder.rs
```diff
@@ -100,143 +103,149 @@ impl<F: FileSystem> DualPipesBuilder<F> {
...
+        let path_id = self.dirs.len() - 1; // path_id to current dir.
+
+        fs::read_dir(dir)?.try_for_each(|e| -> Result<()> {
+            if let Ok(e) = e {
```
tabokie, line 208: This error shouldn't be ignored.

src/file_pipe_log/pipe_builder.rs
```diff
@@ -461,15 +472,23 @@ impl<F: FileSystem> DualPipesBuilder<F> {
                 while written < target_file_size {
                     writer.write_all(&buf).unwrap_or_else(|e| {
```
tabokie, line 473:
```
if let Err(e) = write_all() {
  warn!(..);
  if is_no_space_err(&e) {
    target = 0;
  }
  break;
}
```

src/file_pipe_log/pipe.rs
```diff
@@ -302,6 +303,34 @@ impl<F: FileSystem> SinglePipe<F> {
...
+                    if err_msg.contains("nospace");
+                    then {
+                        true
+                    } else {
+                        false
+                    }
+                }
+            };
+            if no_space_err {
+                if let Err(e) = self.rotate_imp(&mut writable_file) {
```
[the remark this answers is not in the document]
LykxSassinator: > Manual is fine. I don't think unit tests have the permission to create mount points.

Agree.

src/file_pipe_log/pipe.rs
```diff
@@ -258,7 +263,7 @@ impl<F: FileSystem> SinglePipe<F> {
         fail_point!("file_pipe_log::append");
         let mut writable_file = self.writable_file.lock();
         if writable_file.writer.offset() >= self.target_file_size {
-            if let Err(e) = self.rotate_imp(&mut writable_file) {
+            if let Err(e) = self.rotate_imp(&mut writable_file, self.target_file_size) {
```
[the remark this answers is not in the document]
LykxSassinator: That's right. I've considered it and will patch it just in this PR. But I'm struggled with how to add a unit case for this.

I'll update it in next commit, thx for any supplementary comments.

src/file_pipe_log/pipe.rs
```diff
@@ -147,19 +149,34 @@ impl<F: FileSystem> SinglePipe<F> {
...
             seq,
             queue: self.queue,
         };
-        if let Some(f) = self.recycled_files.write().pop_front() {
+        if expected_file_size > self.target_file_size {
+            // On this state, it means that there already exists `nospace` errors and
+            // current buffer is too huge to be appended directly. So, we should free
+            // enough space by clearing several recycled log files until this buffer
+            // could be dumped.
+            let mut expected_size = expected_file_size;
```
tabokie, line 168: (1) recycled file could be on a different disk than the full one
(2) should not modify the `expected_file_size` passed into `fetch_dir`. There could be another writer occupying the space you just released.
(3) if deleting recycled file is enough, there's no need to rotate, which makes the rotated file smaller than usual and hurt future performance.

This branch needs test.

src/file_pipe_log/pipe.rs
```diff
@@ -479,8 +522,27 @@ impl<F: FileSystem> PipeLog for DualPipes<F> {
     }
 }
 
+/// Fetch and return a valid `PathId` of the specific directories.
+pub(crate) fn fetch_dir(paths: &Paths, target_size: usize) -> PathId {
```
tabokie, line 526:
```suggestion
pub(crate) fn find_available_dir(paths: &Paths, target_size: usize) -> PathId {
```

src/log_batch.rs
```diff
@@ -809,8 +819,10 @@ impl LogBatch {
     #[inline]
     pub(crate) fn prepare_write(&mut self, file_context: &LogFileContext) -> Result<()> {
         match self.buf_state {
-            BufState::Encoded(header_offset, entries_len) => {
-                LogItemBatch::prepare_write(&mut self.buf, file_context)?;
+            // `BufState::Sealed` means that `LogBatch` is under a repeated state of dumping.
+            BufState::Encoded(header_offset, entries_len)
+            | BufState::Sealed(header_offset, entries_len) => {
+                self.item_batch.prepare_write(&mut self.buf, file_context)?;
```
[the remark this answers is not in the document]
LykxSassinator: > What if item batch is empty?

In this case, it will return `Ok(0)` directly when calling `engine.write(...)`.
https://github.com/tikv/raft-engine/blob/404e3fefaeeb4da6b7650268d500cfd3fbd29cae/src/engine.rs#L138-L141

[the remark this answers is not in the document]
tabokie: Better make the function more robust. Skip writing when empty should only be an optimization instead of being depended on.

src/file_pipe_log/pipe.rs
LykxSassinator: Yep, these cases have already considered in this branch.

Firstly, about:
>(1) recycled file could be on a different disk than the full one

That's right, this case has three corner cases(Suppose `main-dir` is full, and if and only if the first attempt by `engine.write(...)` is failed, will the whole writing progress try to enter this branch.):
+ 1. `main-dir` has no recycled logs, and `spill-dir` have several recycled logs. 
As the `main-dir` is full, the following `find_available_dir` must skip the `main-dir`, and `spill-dir` should prepare enough space for this write. Otherwise, the whole write progress(`engine.write(...)`) will fail finally with `Err:TryAgain`.
So, in this case, the deletion is necessary for preparing enough space for write. And if there did not exists enough space for writing this `LogBatch`, it will fail as expected because there truly exists no enough space on both disks.
+ 2. `main-dir` has several recycled logs, and `sum(recycled_logs.size())` < `expected_file_size`, but no recycled logs exist in `spill-dir`.
After deletion, the `main-dir` still have no enough space for this second attempt. So, the following `find_available_dir` must skip the `main-dir` as expected. And whether this second attempt on write will success or not depends on whether there exists enough space on `spill-dir`. It's expected. So, int this case, this deletion is also necessary.
+ 3. Both `main-dir` and `spill-dir` have several recycled logs.
The deletion progress will try to release free space on one dir until it's enough to dump this LogBatch into this dir. If and only if free space is still not enough, that is, `deleted_recycled_logs.size() + free_space.size()` < `expected_file_size`, does fail this second write. So, the deletion should be kept.

Secondly, about:
> (2) should not modify the expected_file_size passed into fetch_dir. There could be another writer occupying the space you just released.

Here, the `expected_file_size` do not be modified, just copy it to `expected_size` in this pr.

And for:
> (3) if deleting recycled file is enough, there's no need to rotate, which makes the rotated file smaller than usual and hurt future performance.

Yep, unless `expected_file_size` <= `self.target_file_size`, it will do reuse recycled logs by `rotate`. Otherwise, it will try to free enough space and find an available dir to place this LogBatch by the following `find_available_dir` directly.

As for testing on this branch, I've do it by the following manual case on `centos7`:
```Rust
#[test]
fn test_build_engine_with_spill_dir() {
    fn mount_tmpfs(path: &str) {
        debug_assert!(!path.is_empty());
        #[cfg(target_os = "linux")]
        {
            use std::process::Command;

            let mut mk_device = Command::new("sudo");
            mk_device.args([
                "mount",
                "-t",
                "tmpfs",
                "-o",
                "size=128M",
                "tmpfs",
                path.into(),
            ]);
            mk_device.status().expect("process failed to execute");
        }
    }

    fn umount_tmpfs(path: &str) {
        debug_assert!(!path.is_empty());
        #[cfg(target_os = "linux")]
        {
            use std::process::Command;

            let mut mk_device = Command::new("sudo");
            mk_device.args(["umount", path.into()]);
            mk_device.status().expect("process failed to execute");
        }
    }

    let main_dir = tempfile::Builder::new()
        .prefix("test_build_engine_with_spill_dir_main")
        .tempdir()
        .unwrap();
    let spill_dir = tempfile::Builder::new()
        .prefix("test_build_engine_with_spill_dir_spill")
        .tempdir()
        .unwrap();
    {
        // Mount a tmpfs with limited size.
        mount_tmpfs(main_dir.path().to_str().unwrap());
    }
    let cfg = Config {
        dir: main_dir.path().to_str().unwrap().to_owned(),
        spill_dir: Some(spill_dir.path().to_str().unwrap().to_owned()),
        target_file_size: ReadableSize::mb(1),
        purge_threshold: ReadableSize::mb(128),
        enable_log_recycle: true,
        prefill_for_recycle: true,
        ..Default::default()
    };
    let data = vec![b'x'; 17371];
    let engine = Engine::open(cfg.clone()).unwrap();
    let _f1 = FailGuard::new("file_pipe_log::force_choose_dir", "return");
    for rid in 1..12000 {
        assert!(catch_unwind_silent(|| append(&engine, rid, 1, 100, Some(&data))).is_ok());
    }
    {
        // TODO: failed because current pid still occupy this dir.
        umount_tmpfs(main_dir.path().to_str().unwrap());
    }
}
```
But I'm wondering how to release the temporarily mounted device and make it approved by CI/CD.

src/log_batch.rs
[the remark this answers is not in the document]
LykxSassinator: > Better make the function more robust. Skip writing when empty should only be an optimization instead of being depended on.

Got.

src/file_pipe_log/pipe.rs
tabokie: Honestly, I don't even know how to respond to such a long comment. If the situation is indeed so complicated, you should comment it in the code so that we can discuss line by line there.

> (1.i) So, in this case, the deletion is necessary for preparing enough space for write.

You didn't check the available space of spill_dir. It might not need deletion at all.

> (1.iii) The deletion progress will try to release free space on one dir until it's enough to dump this LogBatch into this dir.

Same thing as before, you don't know if the other directory has enough available space so that it can avoid deletion. And in addition to that, your code won't work if the sum of deleted files from dir1 and dir2 reaches target_size, but the the two individual directories still haven't deleted enough.

> (2) Here, the expected_file_size do not be modified, just copy it to expected_size in this pr.

Well, it's a misread, but you shouldn't copy something into another with nearly the same name.

> Yep, unless expected_file_size <= self.target_file_size, it will do reuse recycled logs by rotate. Otherwise, it will try to free enough space and find an available dir to place this LogBatch by the following find_available_dir directly.

I think there's some misunderstanding. I meant you don't need to rotate the active file, i.e. no need to close the file writer and open a new one.

tabokie: > As for testing on this branch, I've do it by the following manual case on centos7:

Just some failpoint unit tests are enough. Hardware doesn't really matter to this particular function. I only reason I asked before to test on hardware, is because I don't know if truncating a file on an already full disk can succeed.

LykxSassinator: > You didn't check the available space of spill_dir. It might not need deletion at all.

IMO, the too early and unified `deletion` is focused on reducing the extra checking of `available_space` on each dir, and unify the code into one branch `expected_file_size > target_file_size`. For example, the `available_space` consists of recycled logs and free space on each dir respectively, and the checking and deletion of recycled files should be separated individually.

As this pr has paid too long comments on this point, making it too ugly to be reviewed, I accept the [previous advice](https://github.com/tikv/raft-engine/pull/294#discussion_r1121174848) that we should open another issue to focus on optimizing this issue.

src/file_pipe_log/pipe_builder.rs
```diff
@@ -100,141 +103,142 @@ impl<F: FileSystem> DualPipesBuilder<F> {
...
+            (LogQueue::Append, &mut self.append_files, false),
+            (LogQueue::Rewrite, &mut self.rewrite_files, false),
+            (LogQueue::Append, &mut self.recycled_files, true),
+        ] {
+            // Check the file_list and remove the hole of files.
+            let mut invalid_idx = 0_usize;
+            for (i, file_pair) in files.windows(2).enumerate() {
+                // If there exists a black hole or duplicate scenario on FileSeq, these
+                // files should be skipped and cleared.
+                if file_pair[1].seq - file_pair[0].seq != 1 {
```
tabokie, line 126: is duplicate file tested?

tabokie (commented):
is https://github.com/tikv/raft-engine/pull/294#discussion_r1122727110 tested?

tabokie:
update changelog.

src/file_pipe_log/pipe_builder.rs
LykxSassinator: Yep, this case has already been supplemented in `src/engine.rs +2735` as following shows:
```Rust
// Case 4: abnormal case - duplicate FileSeq among different dirs.
{
    // Prerequisite: choose several files and duplicate them to main dir.
    let mut file_count = 0;
    std::fs::read_dir(paths[1]).unwrap().for_each(|e| {
        let p = e.unwrap().path();
        let file_name = p.file_name().unwrap().to_str().unwrap();
        if let Some(FileId {
            queue: LogQueue::Append,
            seq: _,
        }) = FileId::parse_file_name(file_name)
        {
            let mut dst_path = PathBuf::from(&paths[0]);
            dst_path.push(file_name);
            if file_count % 2 == 0 {
                std::fs::copy(p, dst_path).unwrap();
            }
            file_count += 1;
        }
    });
}
let engine = engine.reopen();
// Duplicate log files will be skipped and cleared.
assert!(engine.file_span(LogQueue::Append).0 > start_5);
```

LykxSassinator:
> is [#294 (comment)](https://github.com/tikv/raft-engine/pull/294#discussion_r1122727110) tested?

Done for it.

tabokie:
> update changelog.

ping

LykxSassinator:
> > update changelog.
> 
> ping

Also done in the last commit:
```Markdown
* Add a new configuration `spill-dir` to allow automatic placement of logs into an auxiliary directory when `dir` is full.
```

tabokie (approved):

PR #295, opened 2023-03-01
fix two minor issues

tabokie:
- Sync file after truncating tail corruption.
- Fix a header decoding check.

### DIFF
--- a/src/log_batch.rs
+++ b/src/log_batch.rs
@@ -891,9 +891,10 @@ impl LogBatch {
             )));
         }
 
-        let len = codec::decode_u64(buf)? as usize;
+        let len_and_type = codec::decode_u64(buf)? as usize;
+        let compression_type = CompressionType::from_u8(len_and_type as u8)?;
+        let len = len_and_type >> 8;
         let offset = codec::decode_u64(buf)? as usize;
-        let compression_type = CompressionType::from_u8(len as u8)?;
         if offset > len {
             return Err(Error::Corruption(
                 "Log item offset exceeds log batch length".to_owned(),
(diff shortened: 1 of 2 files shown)


src/log_batch.rs
```diff
@@ -891,9 +891,10 @@ impl LogBatch {
             )));
         }
 
-        let len = codec::decode_u64(buf)? as usize;
+        let len_and_type = codec::decode_u64(buf)? as usize;
+        let compression_type = CompressionType::from_u8(len_and_type as u8)?;
+        let len = len_and_type >> 8;
```
BusyJay, line 896: Any tests?

tabokie: added

src/log_batch.rs
```diff
@@ -1531,6 +1531,49 @@ mod tests {
...
+            .add_entries::<Entry>(region_id, &generate_entries(1, 11, Some(&data)))
+            .unwrap();
+        batch
+            .put(region_id, b"key".to_vec(), b"value".to_vec())
+            .unwrap();
+        // enable compression so that len_and_type > len.
+        batch.finish_populate(1).unwrap();
+        let file_context = LogFileContext::new(FileId::dummy(LogQueue::Append), Version::default());
+        batch.prepare_write(&file_context).unwrap();
+        let encoded = batch.encoded_bytes();
```
BusyJay, line 1549: Still missing a valid decode so we can verify it won't panic anymore.

BusyJay (approved):

src/log_batch.rs
tabokie: There're plenty of tests above testing the valid cases.

PR #296, opened 2023-03-03
open files with less permissions if possible

hicqu:
open files with less permissions if possible.

### DIFF
--- a/src/copy.rs
+++ b/src/copy.rs
@@ -1,0 +1,120 @@
+// Copyright (c) 2023-present, PingCAP, Inc. Licensed under Apache-2.0.
+
+use std::fs::{copy, create_dir_all};
+use std::os::unix::fs::symlink;
+use std::path::Path;
+use std::sync::Arc;
+
+use crate::config::Config;
+use crate::env::{FileSystem, OFlag};
+use crate::file_pipe_log::{FileNameExt, FilePipeLogBuilder};
+use crate::pipe_log::{FileId, LogQueue};
+
+/// Make a copy from `source` to `target`. `source` should exists but `target`
+/// shouldn't.
+///
+/// *minimum* means *symlink* will be used if possbile, otherwise *copy* will be
+/// used instead. Generally all inactive log files will be symlinked, but the
+/// last active one will be copied.
+///
+/// After the copy is made both of 2 engines can be started and run at the same
+/// time.
+pub fn minimum_copy<F, P>(cfg: &Config, fs: Arc<F>, target: P) -> Result<(), String>
+where
+    F: FileSystem,
+    P: AsRef<Path>,
+{
+    if cfg.enable_log_recycle {
+        return Err("enable_log_recycle should be false".to_owned());
+    }
+
+    let mut cfg = cfg.clone();
+    cfg.sanitize()
+        .map_err(|e| format!("sanitize config: {}", e))?;
+
+    create_dir_all(&target)
+        .map_err(|e| format!("create_dir_all({}): {}", target.as_ref().display(), e))?;
+
+    let mut builder = FilePipeLogBuilder::new(cfg.clone(), fs, vec![]);
+    builder
+        .scan_impl(|_, _| -> OFlag { OFlag::O_RDONLY })
+        .map_err(|e| format!("scan files: {}", e))?;
+
+    // Iterate all log files and rewrite files.
+    for (queue, files) in [
+        (LogQueue::Append, builder.get_append_queue_files()),
+        (LogQueue::Rewrite, builder.get_rewrite_queue_files()),
+    ] {
+        let count = files.len();
+        for (i, f) in files.iter().enumerate() {
+            let file_id = FileId::new(queue, f.seq);
+            let src = file_id.build_file_path(&cfg.dir);
+            let tgt = file_id.build_file_path(&target);
+            let (tag, res) = if i < count - 1 {
+                ("symlink", symlink(&src, &tgt))
+            } else {
+                ("copy", copy(&src, &tgt).map(|_| ()))
+            };
+            res.map_err(|e| format!("{}({}, {}): {}", tag, src.display(), tgt.display(), e))?;
+        }
+    }
+
+    Ok(())
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::engine::tests::RaftLogEngine;
+    use crate::env::DefaultFileSystem;
+    use crate::{LogBatch, ReadableSize};
+    use std::path::PathBuf;
+
+    #[test]
+    fn test_minimum_copy() {
+        let dir = tempfile::Builder::new()
+            .prefix("test_minimum_copy")
+            .tempdir()
+            .unwrap();
+
+        let mut source = PathBuf::from(dir.as_ref());
+        source.push("source");
+        let mut cfg = Config {
+            dir: source.to_str().unwrap().to_owned(),
+            target_file_size: ReadableSize::kb(1),
+            enable_log_recycle: false,
+            ..Default::default()
+        };
+        let engine = RaftLogEngine::open(cfg.clone()).unwrap();
+
+        let mut log_batch = LogBatch::default();
+        log_batch.put(1, vec![b'1'; 16], vec![b'v'; 1024]).unwrap();
+        engine.write(&mut log_batch, false).unwrap();
+        engine.purge_manager().must_rewrite_append_queue(None, None);
+
+        let mut log_batch = LogBatch::default();
+        log_batch.put(1, vec![b'2'; 16], vec![b'v'; 1024]).unwrap();
+        engine.write(&mut log_batch, false).unwrap();
+        engine.purge_manager().must_rewrite_append_queue(None, None);
+
+        let mut log_batch = LogBatch::default();
+        log_batch.put(1, vec![b'3'; 16], vec![b'v'; 1024]).unwrap();
+        engine.write(&mut log_batch, false).unwrap();
+
+        let mut log_batch = LogBatch::default();
+        log_batch.put(1, vec![b'4'; 16], vec![b'v'; 1024]).unwrap();
+        engine.write(&mut log_batch, false).unwrap();
+        drop(engine);
+
+        let mut target = PathBuf::from(dir.as_ref());
+        target.push("target");
+        minimum_copy(&cfg, Arc::new(DefaultFileSystem), &target).unwrap();
+        cfg.dir = target.to_str().unwrap().to_owned();
+        let engine = RaftLogEngine::open(cfg).unwrap();
+
+        assert!(engine.get(1, vec![b'1'; 16].as_ref()).is_some());
+        assert!(engine.get(1, vec![b'2'; 16].as_ref()).is_some());
+        assert!(engine.get(1, vec![b'3'; 16].as_ref()).is_some());
+        assert!(engine.get(1, vec![b'4'; 16].as_ref()).is_some());
+    }
+}
--- a/src/env/mod.rs
+++ b/src/env/mod.rs
@@ -7,8 +7,30 @@ use std::sync::Arc;
 mod default;
 mod obfuscated;
 
+use bitflags::bitflags;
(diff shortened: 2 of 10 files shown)


src/file_pipe_log/pipe_builder.rs
```diff
@@ -223,7 +236,8 @@ impl<F: FileSystem> DualPipesBuilder<F> {
                         );
                         files.clear();
                     } else {
-                        let handle = Arc::new(self.file_system.open(&path)?);
+                        let flag = get_open_flag(self.cfg.recovery_mode, seq == max_id);
```
hicqu, line 239: I guess `seq == max_id` means this file is the active one. Is it correct? @tabokie

tabokie: yes

src/env/mod.rs
```diff
@@ -18,7 +40,7 @@ pub trait FileSystem: Send + Sync {
 
     fn create<P: AsRef<Path>>(&self, path: P) -> Result<Self::Handle>;
 
-    fn open<P: AsRef<Path>>(&self, path: P) -> Result<Self::Handle>;
+    fn open<P: AsRef<Path>>(&self, path: P, flags: OFlag) -> Result<Self::Handle>;
```
tabokie, line 43: Let's not use bitmaps. You can use a simple read_only flag or ReadOptions.

src/file_pipe_log/pipe_builder.rs
```diff
@@ -100,6 +99,20 @@ impl<F: FileSystem> DualPipesBuilder<F> {
     /// Scans for all log files under the working directory. The directory will
     /// be created if not exists.
     pub fn scan(&mut self) -> Result<()> {
```
tabokie, line 101: No need for a closure to accommodate one single corner case. Maybe use a read_only flag too.

src/copy.rs
```diff
@@ -0,0 +1,120 @@
...
+/// last active one will be copied.
+///
+/// After the copy is made both of 2 engines can be started and run at the same
+/// time.
+pub fn minimum_copy<F, P>(cfg: &Config, fs: Arc<F>, target: P) -> Result<(), String>
+where
+    F: FileSystem,
+    P: AsRef<Path>,
+{
+    if cfg.enable_log_recycle {
```
tabokie, line 27: why? the error from recycle should be ignored.

tabokie (commented):
We use delete to purge log files, an error here will cause purge to abort.

src/copy.rs
hicqu: I thin minimun_copy is only allowed when `recycle` is disabled. Otherwise the target instance can't be run concurrently with the original one.

tabokie: But you can't disallow recycle when opening this db again. e.g. user change the config after creating a checkpoint.

hicqu: Users can also disable recycle in the config before creating a checkpoint. I think it lefts less chances to users to make mistakes.

src/file_pipe_log/pipe_builder.rs
hicqu: `read_only` can't describe the behavior accurately. The last log file in the queue is still opened in read-write mode.

tabokie: The default behavior is to open active in rw, others in r. The corner case behavior is to open all files in r. You only needs a flag to describe the corner case behavior, namely "read_only" or "force_read_only".

src/copy.rs
```diff
@@ -0,0 +1,120 @@
...
+/// Make a copy from `source` to `target`. `source` should exists but `target`
+/// shouldn't.
+///
+/// *minimum* means *symlink* will be used if possbile, otherwise *copy* will be
+/// used instead. Generally all inactive log files will be symlinked, but the
+/// last active one will be copied.
+///
+/// After the copy is made both of 2 engines can be started and run at the same
+/// time.
+pub fn minimum_copy<F, P>(cfg: &Config, fs: Arc<F>, target: P) -> Result<(), String>
```
tabokie, line 22: I prefer it be an engine method, and hooked to engine-ctl, like `consistency_check`.

tabokie: A better name would be "fork".

src/copy.rs
```diff
@@ -0,0 +1,120 @@
...
+
+/// Make a copy from `source` to `target`. `source` should exists but `target`
+/// shouldn't.
+///
+/// *minimum* means *symlink* will be used if possbile, otherwise *copy* will be
+/// used instead. Generally all inactive log files will be symlinked, but the
+/// last active one will be copied.
+///
+/// After the copy is made both of 2 engines can be started and run at the same
+/// time.
```
tabokie, line 21: Comment the caveat about the risk of opening the forked engine with recycled enabled.

src/copy.rs
```diff
@@ -0,0 +1,120 @@
...
+pub fn minimum_copy<F, P>(cfg: &Config, fs: Arc<F>, target: P) -> Result<(), String>
+where
+    F: FileSystem,
+    P: AsRef<Path>,
+{
+    if cfg.enable_log_recycle {
+        return Err("enable_log_recycle should be false".to_owned());
```
tabokie, line 28: Needs tests.

src/copy.rs
```diff
@@ -0,0 +1,120 @@
...
+        engine.write(&mut log_batch, false).unwrap();
+        drop(engine);
+
+        let mut target = PathBuf::from(dir.as_ref());
+        target.push("target");
+        minimum_copy(&cfg, Arc::new(DefaultFileSystem), &target).unwrap();
+        cfg.dir = target.to_str().unwrap().to_owned();
+        let engine = RaftLogEngine::open(cfg).unwrap();
+
+        assert!(engine.get(1, vec![b'1'; 16].as_ref()).is_some());
```
tabokie, line 115: Needs to test both engines can function independently.

src/env/mod.rs
```diff
@@ -8,8 +8,24 @@ mod default;
...
+use nix::fcntl::OFlag as NixOFlag;
 pub use obfuscated::ObfuscatedFileSystem;
 
+#[derive(Clone, Copy)]
+pub enum Permission {
+    ReadOnly,
+    ReadWrite,
+}
+
+impl From<Permission> for NixOFlag {
```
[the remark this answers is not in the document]
hicqu: Seems it's not introduced as a public name.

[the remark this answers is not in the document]
tabokie: Doesn't matter. This file is for trait definition only. A util function in default.rs works the same.

src/fork.rs
```diff
@@ -0,0 +1,175 @@
...
+    let mut cfg = cfg.clone();
+    cfg.sanitize()
+        .map_err(|e| format!("sanitize config: {}", e))?;
+
+    create_dir_all(&target)
+        .map_err(|e| format!("create_dir_all({}): {}", target.as_ref().display(), e))?;
+
+    let mut builder = FilePipeLogBuilder::new(cfg.clone(), fs, vec![]);
+    builder
+        .scan_and_sort(false)
```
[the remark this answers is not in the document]
tabokie: Needs comment that forking an opened engine will cause corruption.

src/env/mod.rs
```diff
@@ -10,6 +10,12 @@ mod obfuscated;
 pub use default::DefaultFileSystem;
 pub use obfuscated::ObfuscatedFileSystem;
 
+#[derive(Clone, Copy)]
```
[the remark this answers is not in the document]
hicqu: It's required by some users. So I add a case to cover this line.

src/file_pipe_log/pipe_builder.rs
```diff
@@ -541,3 +610,10 @@ pub(super) fn lock_dir<P: AsRef<Path>>(dir: P) -> Result<StdFile> {
     })?;
     Ok(lock_file)
 }
+
+pub(crate) struct FileName {
+    pub seq: FileSeq,
+    name: String,
```
tabokie, line 616: Why not store path directly?

src/file_pipe_log/pipe_builder.rs
```diff
@@ -184,22 +203,44 @@ impl<F: FileSystem> DualPipesBuilder<F> {
         Ok(())
     }
 
-    fn scan_dir(&mut self, dir: String) -> Result<()> {
-        let dir = Path::new(&dir);
+    pub(crate) fn scan_and_sort(&mut self, exclude_others: bool) -> Result<()> {
```
tabokie, line 206:
```suggestion
    pub(crate) fn scan_and_sort(&mut self, lock: bool) -> Result<()> {
```

src/file_pipe_log/pipe_builder.rs
```diff
@@ -184,22 +203,44 @@ impl<F: FileSystem> DualPipesBuilder<F> {
...
+            self.dir_locks.push(lock_dir(dir)?);
+        }
         self.dirs.push(dir.to_path_buf());
 
-        let path_id = self.dirs.len() - 1; // path_id to current dir.
+        let path_id = self.dirs.len() - 1;
+        self.do_scan(dir, path_id)
+    }
+
+    pub fn do_scan(&mut self, dir: &Path, path_id: usize) -> Result<()> {
```
tabokie, line 243: No need for a new function.

src/file_pipe_log/pipe_builder.rs
```diff
@@ -80,9 +81,14 @@ pub struct DualPipesBuilder<F: FileSystem> {
     /// Only filled after a successful call of `DualPipesBuilder::scan`.
     dirs: Paths,
     dir_locks: Vec<StdFile>,
+
     append_files: Vec<File<F>>,
     rewrite_files: Vec<File<F>>,
     recycled_files: Vec<File<F>>,
+
+    pub(crate) append_file_names: Vec<FileName>,
```
tabokie, line 89: Move these between dir_locks and append_files, the same order as they are created.

tabokie (commented):
Needs update changelog, rest looks good.

src/file_pipe_log/pipe_builder.rs
```diff
@@ -184,22 +203,40 @@ impl<F: FileSystem> DualPipesBuilder<F> {
...
+            self.scan_dir(&dir, lock)?;
+        }
+
+        self.append_file_names.sort_by(|a, b| a.seq.cmp(&b.seq));
+        self.rewrite_file_names.sort_by(|a, b| a.seq.cmp(&b.seq));
+        self.recycled_file_names.sort_by(|a, b| a.seq.cmp(&b.seq));
+        Ok(())
+    }
+
+    fn scan_dir(&mut self, dir: &str, exclude_others: bool) -> Result<()> {
```
tabokie, line 220:
```suggestion
    fn scan_dir(&mut self, dir: &str, lock: bool) -> Result<()> {
```

src/file_pipe_log/pipe_builder.rs
```diff
@@ -528,6 +567,25 @@ impl<F: FileSystem> DualPipesBuilder<F> {
         )?;
         DualPipes::open(self.dir_locks, appender, rewriter)
     }
+
+    fn open(&self, file_name: &FileName, is_last_one: bool) -> Result<File<F>> {
+        let perm = if file_name.recycled
```
tabokie, line 572: No need to write recycled? recycled file will be reopened with the right permission before write.

tabokie (approved):

src/file_pipe_log/pipe_builder.rs
```diff
@@ -528,6 +564,25 @@ impl<F: FileSystem> DualPipesBuilder<F> {
         )?;
         DualPipes::open(self.dir_locks, appender, rewriter)
     }
+
+    fn open(&self, file_name: &FileName, is_last_one: bool) -> Result<File<F>> {
+        let perm = if false ||
```
tabokie, line 569: why false?

hicqu: Following 2 conditions have their own comments. Use `false` as a placeholder can make 2 comments more clear.

PR #297, opened 2023-03-09
add metrics of recycled file count

LykxSassinator:
## Description
This pr supplements the necessary metrics for monitoring the amount of recycled logs.

### DIFF
--- a/src/file_pipe_log/pipe.rs
+++ b/src/file_pipe_log/pipe.rs
@@ -120,7 +120,7 @@ impl<F: FileSystem> SinglePipe<F> {
             format: f.format,
         };
 
-        let len = active_files.len();
+        let (len, recycled_len) = (active_files.len(), recycled_files.len());
         for f in active_files.iter() {
             for listener in &listeners {
                 listener.post_new_log_file(FileId { queue, seq: f.seq });
@@ -143,7 +143,7 @@ impl<F: FileSystem> SinglePipe<F> {
             recycled_files: RwLock::new(recycled_files.into()).into(),
             writable_file: Mutex::new(writable_file).into(),
         };
-        pipe.flush_metrics(len);
+        pipe.flush_metrics(len, recycled_len);
         Ok(pipe)
     }
 
@@ -227,7 +227,8 @@ impl<F: FileSystem> SinglePipe<F> {
             files.push_back(f);
             files.len()
         };
-        self.flush_metrics(len);
+        let recycled_len = self.recycled_files.read().len();
+        self.flush_metrics(len, recycled_len);
         for listener in &self.listeners {
             listener.post_new_log_file(FileId {
                 queue: self.queue,
@@ -238,10 +239,16 @@ impl<F: FileSystem> SinglePipe<F> {
     }
 
     /// Synchronizes current states to related metrics.
-    fn flush_metrics(&self, len: usize) {
+    fn flush_metrics(&self, len: usize, recycled_len: usize) {
         match self.queue {
-            LogQueue::Append => LOG_FILE_COUNT.append.set(len as i64),
-            LogQueue::Rewrite => LOG_FILE_COUNT.rewrite.set(len as i64),
+            LogQueue::Append => {
+                LOG_FILE_COUNT.append.set(len as i64);
+                RECYCLED_FILE_COUNT.append.set(recycled_len as i64);
+            }
+            LogQueue::Rewrite => {
+                LOG_FILE_COUNT.rewrite.set(len as i64);
+                RECYCLED_FILE_COUNT.rewrite.set(recycled_len as i64);
+            }
         }
     }
 }
@@ -382,7 +389,10 @@ impl<F: FileSystem> SinglePipe<F> {
             (files.len(), tail)
         };
         let purged_len = purged_files.len();
-        if purged_len > 0 {
+        let recycled_len = if purged_len == 0 {
+            // No need to purge files, return the count of recycled files directly.
+            self.recycled_files.read().len()
+        } else {
             let remains_capacity = self.capacity.saturating_sub(len);
             let (recycled_start, mut recycled_len) = {
                 let files = self.recycled_files.read();
@@ -425,8 +435,9 @@ impl<F: FileSystem> SinglePipe<F> {
             }
             debug_assert!(recycled_len <= remains_capacity);
             self.recycled_files.write().append(&mut new_recycled);
-        }
-        self.flush_metrics(len);
+            recycled_len
+        };
+        self.flush_metrics(len, recycled_len);
         Ok(purged_len)
     }
 }
--- a/src/file_pipe_log/pipe.rs
+++ b/src/file_pipe_log/pipe.rs
@@ -158,7 +158,7 @@ impl<F: FileSystem> SinglePipe<F> {
 
     /// Recycles one obsolete file from the recycled file list and return its
     /// [`PathId`] and [`F::Handle`] if success.
-    fn recycle_file(&self, seq: FileSeq) -> Option<(PathId, F::Handle)> {
+    fn recycle_file(&self, seq: FileSeq) -> Option<Result<(PathId, F::Handle)>> {
         let new_file_id = FileId {
             seq,
             queue: self.queue,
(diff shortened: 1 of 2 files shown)


LykxSassinator:
@tabokie PTAL, thx.

src/file_pipe_log/pipe.rs
```diff
@@ -382,7 +391,10 @@ impl<F: FileSystem> SinglePipe<F> {
             (files.len(), tail)
         };
         let purged_len = purged_files.len();
-        if purged_len > 0 {
+        let recycled_len = if purged_len == 0 {
+            // No need to purge files, return the count of recycled files directly.
+            self.recycled_files.read().len()
```
tabokie, line 396: No need to update too?

LykxSassinator: IMO, the `rotate` will also cost recycled files if there exists some. And if we do not update the metrics on `recycled_files.len()` by the period `purge` progress, the count of `recycled_files.len()` won't be fresh as expected.

So, I delay the update of `recycled_files.len()` in `rotate` to `purge` at here.

tabokie: How about update it in rotate? You can break new_file into recycle_file -> Option<File> and new_file -> Result<File>.

src/file_pipe_log/pipe.rs
```diff
@@ -155,12 +156,22 @@ impl<F: FileSystem> SinglePipe<F> {
...
+
+    /// Creates a new log file according to the given [`FileSeq`].
+    ///
+    /// If `recycle_file` is valid, it will reuse it as priority.
+    fn new_file(&self, seq: FileSeq, recycle_file: Option<File<F>>) -> Result<(PathId, F::Handle)> {
         let new_file_id = FileId {
             seq,
             queue: self.queue,
         };
-        if let Some(f) = self.recycled_files.write().pop_front() {
```
tabokie, line 163: I just realized there's a bug at L183? If we fail to reuse file, the dst_path should be empty, fs.open will fail.

src/file_pipe_log/pipe.rs
```diff
@@ -155,12 +156,22 @@ impl<F: FileSystem> SinglePipe<F> {
...
+        if let Some(f) = recycle_file {
             let src_path = self.paths[f.path_id].join(build_recycled_file_name(f.seq));
```
tabokie, line 175: Better move these parts into `recycle_file` too, and update the counters in `recycle_file`. The reason is we try to put the metrics update the same level as the file collection update.

src/file_pipe_log/pipe.rs
LykxSassinator: > I just realized there's a bug at L183? If we fail to reuse file, the dst_path should be empty, fs.open will fail.

By the definition of `reuse, `reuse` just use a reference of `dst_path`, it should not be empty. Right ?

LykxSassinator: > I just realized there's a bug at L183? If we fail to reuse file, the dst_path should be empty, fs.open will fail.

Precisely, if we fail to reuse it, the code path will go through `L186` to create a new file according to the `dst_path`. Only if `reuse` success, the code path will go to L183 (open it). So, this case should not exist, right?

tabokie: You are right, I was being silly.

src/file_pipe_log/pipe.rs
```diff
@@ -199,7 +216,14 @@ impl<F: FileSystem> SinglePipe<F> {
 
         writable_file.writer.close()?;
 
-        let (path_id, handle) = self.new_file(new_seq)?;
+        let (path_id, handle) = {
+            if let Some((id, path)) = self.recycle_file(new_seq) {
```
tabokie, line 220: Why not open it inside too? So the code can be more concise like `self.recycle_file(seq).unwrap_or(|| self.new_file(seq))?`

LykxSassinator: Several things for the reason why I coded as this format:
- If we do just like u recommend, just like:
```Rust
let (path_id, handle) = self
            .recycle_file(new_seq)
            .unwrap_or_else(|_| self.new_file(new_seq).unwrap());
```
It will `panic` if `new_file` returns a `Err`.

- If we want to return the `Err` to above, just like:
```Rust
let (path_id, handle) = self
            .recycle_file(new_seq)
            .unwrap_or_else(|_| self.new_file(new_seq)?);
```
It's invalid in rust syntax as:
```Rust
error[E0277]: the `?` operator can only be used in a closure that returns `Result` or `Option` (or another type that implements `FromResidual`)
   --> src/file_pipe_log/pipe.rs:226:55
    |
226 |             .unwrap_or_else(|_| self.new_file(new_seq)?);
    |                             --------------------------^
    |                             |                         |
    |                             |                         cannot use the `?` operator in a closure that returns `(usize, <F as FileSystem>::Handle)`
    |                             this function should return `Result` or `Option` to accept `?`
    |
    = help: the trait `FromResidual<std::result::Result<Infallible, errors::Error>>` is not implemented for `(usize, <F as FileSystem>::Handle)`
```

- Also, we can code it just like:
```Rust
let (path_id, handle) = self
            .recycle_file(new_seq)
            .unwrap_or(self.new_file(new_seq)?);
```
But this is not what we expect, as the `unwrap_or(function(...))` format will try to get the value of `function(...)` and pass it to `unwrap_or` directly:
```
Arguments passed to unwrap_or are eagerly evaluated;
```

tabokie: By opening inside, the recycle_file will return a Option<Result<File>>, and new_file returns Result<File>, you can fuse them with recycle_file().unwrap_or_else(|| self.new_file())?

src/file_pipe_log/pipe.rs
```diff
@@ -169,12 +176,24 @@ impl<F: FileSystem> SinglePipe<F> {
                     error!("error while trying to delete recycled file, err: {}", e);
                 }
             } else {
-                return Ok((f.path_id, self.file_system.open(&dst_path)?));
+                self.flush_recycle_metrics(recycle_len);
+                if let Ok(handle) = self.file_system.open(&dst_path) {
```
tabokie, line 180: This changes old behavior, the error of open should be bubbled out.

src/file_pipe_log/pipe.rs
```diff
@@ -177,8 +177,9 @@ impl<F: FileSystem> SinglePipe<F> {
                 }
             } else {
                 self.flush_recycle_metrics(recycle_len);
-                if let Ok(handle) = self.file_system.open(&dst_path) {
-                    return Some(Ok((f.path_id, handle)));
+                match self.file_system.open(&dst_path) {
+                    Ok(handle) => return Some(Ok((f.path_id, handle))),
+                    Err(e) => return Some(Err(Error::Io(e))),
```
tabokie, line 182:
```suggestion
                Some(self.file_system.open(&dst_path).into())
```
or
```suggestion
                Some(Ok(self.file_system.open(&dst_path)?))
```

LykxSassinator: Nope, the return type of this function `recycle_file` is defined as `Option<Result<(PathID, F::Handle)>>`, so it cannot be adapted to the syntax `?`.

Let me guess how to make it more elegant.

LykxSassinator: Oops...failed. I recommend to use current style of coding.

src/file_pipe_log/pipe.rs
```diff
@@ -169,13 +176,25 @@ impl<F: FileSystem> SinglePipe<F> {
                     error!("error while trying to delete recycled file, err: {}", e);
                 }
             } else {
-                let handle = self.file_system.open(&dst_path, Permission::ReadWrite)?;
-                return Ok((f.path_id, handle));
+                self.flush_recycle_metrics(recycle_len);
+                match self.file_system.open(&dst_path, Permission::ReadWrite) {
+                    Ok(handle) => return Some(Ok((f.path_id, handle))),
+                    Err(e) => return Some(Err(Error::Io(e))),
+                }
```
tabokie, line 183:
```suggestion
                return match self.file_system.open(&dst_path, Permission::ReadWrite) {
                    Ok(handle) => Some(Ok((f.path_id, handle))),
                    Err(e) => Some(Err(e.into())),
                };
```

PR #299, opened 2023-03-29
use option<> for purge_threshold and recovery_threads

tonyxuqqi:
This is to make it possible for the caller to dynamically choose the default value under different situations like single rocksdb or multi-rocksdb.

### DIFF
--- a/src/purge.rs
+++ b/src/purge.rs
@@ -175,7 +175,7 @@ where
 
         let total_size = self.pipe_log.total_size(queue);
         match queue {
-            LogQueue::Append => total_size > self.cfg.purge_threshold.0 as usize,
+            LogQueue::Append => total_size > self.cfg.purge_threshold().0 as usize,
             LogQueue::Rewrite => {
                 let compacted_rewrites_ratio = self.global_stats.deleted_rewrite_entries() as f64
                     / self.global_stats.rewrite_entries() as f64;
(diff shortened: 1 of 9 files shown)


src/purge.rs
```diff
@@ -175,7 +175,7 @@ where
 
         let total_size = self.pipe_log.total_size(queue);
         match queue {
-            LogQueue::Append => total_size > self.cfg.purge_threshold.0 as usize,
+            LogQueue::Append => total_size > self.cfg.purge_threshold().0 as usize,
```
LykxSassinator, line 178: Maybe this is not what u expect. It's incorrect here. 

It's recommend to `exit` this function if `self.cfg.purge_threshold.is_none()` to avoid `purge`. Also, u can add this processing in `purge_expired_files()`, just like this:
```Rust
pub fn purge_expired_files(&self) -> Result<Vec<u64>> {
        let _t = StopWatch::new(&*ENGINE_PURGE_DURATION_HISTOGRAM);
        let guard = self.force_rewrite_candidates.try_lock();
        if guard.is_none() {
            warn!("Unable to purge expired files: locked");
            return Ok(vec![]);
        }
        // If no settings for `purge_threshold`, just skip the whole purge progress.
        if self.cfg.purge_threshold.is_none() {
            return Ok(vec![]);
        }
        let mut rewrite_candidate_regions = guard.unwrap();
        // ...
```

tabokie:
Seems unneeded.

PR #301, opened 2023-03-30
Expose the setting of the maximum size of `prefill-for-recycle` to users

LykxSassinator:
## Description
This pr is used to unleash the limitation of `prefill-for-recycle`, used to adaptive to the requirement of large threshold.

Related issue: https://github.com/tikv/tikv/issues/14468.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -108,6 +108,13 @@ pub struct Config {
     ///
     /// Default: false
     pub prefill_for_recycle: bool,
+
+    /// Maximum capacity for preparing log files for recycling when start.
+    /// If not `None`, its size is equal to `purge-threshold`.
+    /// Only available for `prefill-for-recycle` is true.
+    ///
+    /// Default: None
+    pub prefill_limit: Option<ReadableSize>,
 }
 
 impl Default for Config {
@@ -129,6 +136,7 @@ impl Default for Config {
             memory_limit: None,
             enable_log_recycle: false,
             prefill_for_recycle: false,
+            prefill_limit: None,
         };
         // Test-specific configurations.
         #[cfg(test)]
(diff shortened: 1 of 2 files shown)


src/file_pipe_log/pipe_builder.rs
```diff
@@ -484,7 +483,6 @@ impl<F: FileSystem> DualPipesBuilder<F> {
         } else {
             0
         };
-        target = cmp::min(target, MAX_PREFILL_SIZE / target_file_size);
```
tonyxuqqi, line 487: target没了能编译吗？

tonyxuqqi (approved):

BusyJay:
Why prefill writes data instead of using falloc?

LykxSassinator:
> Why prefill writes data instead of using falloc?

Directly using `falloc` just creates sparse files. And it will introduce extra initializing on blocks when reusing them by recycling logs, making the first rewrite progress costs more time on `fsync`.

LykxSassinator:
@BusyJay PTAL, thx.

src/config.rs
```diff
@@ -180,17 +188,18 @@ impl Config {
                 "prefill is not allowed when log recycle is disabled"
             ));
         }
+        if !self.prefill_for_recycle && self.prefill_limit.is_some() {
+            warn!("prefill-limit will be ignored when prefill is disabled");
+            self.prefill_limit = None;
+        }
+        if self.prefill_for_recycle && self.prefill_limit.is_none() {
+            warn!("prefill-limit will be calibrated to purge-threshold");
```
BusyJay, line 196: Should be `info`.

BusyJay (approved):
Rest LGTM

PR #310, opened 2023-06-06
avoid rename during recycling

tabokie:


tabokie:
@LykxSassinator Let's use this version instead. I cleaned up some tests.

LykxSassinator (approved):
Pls update and merge latest master branch. Rest LGTM

src/file_pipe_log/pipe.rs
```diff
@@ -73,6 +76,17 @@ impl<F: FileSystem> Drop for SinglePipe<F> {
         if let Err(e) = writable_file.writer.close() {
             error!("error while closing the active writer: {e}");
         }
+        // Delete the newest first to create a hole.
```
tonyxuqqi, line 79: Do you want to delete one file or all the recycled_files?

src/file_pipe_log/pipe.rs
```diff
@@ -411,16 +432,11 @@ impl<F: FileSystem> SinglePipe<F> {
...
-                    .front()
-                    .map_or_else(|| (DEFAULT_FIRST_FILE_SEQ, 0), |f| (f.seq, files.len()))
-            };
+            let mut recycled_len = self.recycled_files.read().len();
             let mut new_recycled = VecDeque::new();
-            // The newly purged files from `self.active_files` should be renamed
-            // to recycled files with LOG_APPEND_RESERVED_SUFFIX suffix, to reduce the
-            // unnecessary recovery timecost when restarting.
+            // We don't rename the append file because on some platform it could cause I/O
+            // jitters. Instead we best-effort delete them during shutdown to reduce
```
tonyxuqqi, line 438: Once best-effort flush memtable is implemented, not sure if this will slow down the flush memtable. 
Also in tikv restart, the recreate the file takes extra IOs.  To me delete files seems not the efficient way. Can we do the rename in the start? It will be more efficient than recreate anyway.

tonyxuqqi (approved):

PR #314, opened 2023-06-12
fix data loss caused by atomic group being partially purged

### LINKED ISSUE #315: Atomic group causes data loss
Using dump tool, we locate the missing entry is in file X. File X is the oldest rewrite log available. Use dump tool to analyze file X, it only contains `AtomicGroup::Middle` but no `AtomicGroup::Begin`.

What happened is, memtable is not aware of the atomic group that each entry/kv belongs to. After user deletes some of the entries inside an atomic group, memtable will permit the purge thread to delete the physical files that contain these entries. As a result, the `Begin` marker of this atomic group is deleted. Next restart this group will be broken and all entries including those not deleted by user will disappear.

tabokie:
Close #315

Record the start and end position of an atomic group in memtable.

### DIFF
--- a/src/memtable.rs
+++ b/src/memtable.rs
@@ -155,6 +155,16 @@ pub struct MemTable<A: AllocatorTrait> {
     /// A map of active key value pairs.
     kvs: BTreeMap<Vec<u8>, (Vec<u8>, FileId)>,
 
+    /// (start_seq, end_seq).
+    /// If there's an active entry stored before end_seq, it possibly belongs to
+    /// an atomic group. In order to not lose this entry, We cannot delete any
+    /// other entries in that group.
+    /// Only applies to Rewrite queue. Each Raft Group has at most one atomic
+    /// group at a time, because we only use atomic group for rewrite-rewrite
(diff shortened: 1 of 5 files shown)


src/memtable.rs
```diff
@@ -155,6 +155,16 @@ pub struct MemTable<A: AllocatorTrait> {
...
 
+    /// (start_seq, end_seq).
+    /// If there's an active entry stored before end_seq, it possibly belongs to
+    /// an atomic group. In order to not lose this entry, We cannot delete any
+    /// other entries in that group.
+    /// Only applies to Rewrite queue. Each Raft Group has at most one atomic
+    /// group at a time, because we only use atomic group for rewrite-rewrite
+    /// operation, a group always contains all the Rewrite entries in a Raft
+    /// Group.
+    atomic_group: Option<(FileSeq, FileSeq)>,
```
overvenus, line 166: Is it possible to have two groups?

tabokie: The comment above says there can only be one group at a time.

overvenus (approved):

tonyxuqqi (approved):

PR #69, opened 2021-02-23
Initial win32 support

skyline75489:
Supports https://github.com/tikv/tikv/issues/9103

PR #322, opened 2023-07-04
build on windows system.

MichaelScofield:
Let cargo build pass under windows system:
1. use windows os specific symlink API in `fork.rs`
2. use different `LogFd` implementations in each os. Guarded (or choosen) by feature flags
    - for non-linux or non-unix system, use the plain `std::fs` APIs instead
    - for linux and unix system, still use the `nix` raw bindings

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -10,7 +10,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ ubuntu-latest, macos-latest ]
+        os: [ ubuntu-latest, macos-latest, windows-latest ]
     steps:
       - uses: actions/checkout@v2
         with:
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -68,7 +68,7 @@ kvproto = { git = "https://github.com/pingcap/kvproto.git", default-features = f
 raft = { git = "https://github.com/tikv/raft-rs", branch = "master", default-features = false, features = ["protobuf-codec"] }
 rand = "0.8"
 rand_distr = "0.4"
-tempfile = "3.1"
+tempfile = "3.6"
 toml = "0.7"
 
 [features]
@@ -87,6 +87,7 @@ swap = [
   "nightly",
   "memmap2",
 ]
+std_fs = []
 
 nightly_group = ["nightly", "swap"]
 
@@ -94,6 +95,8 @@ nightly_group = ["nightly", "swap"]
 raft-proto = { git = "https://github.com/tikv/raft-rs", branch = "master" }
 protobuf = { git = "https://github.com/pingcap/rust-protobuf", branch = "v2.8" }
 protobuf-codegen = { git = "https://github.com/pingcap/rust-protobuf", branch = "v2.8" }
+# TODO: Use official grpc-rs once https://github.com/tikv/grpc-rs/pull/622 is merged.
+grpcio = { git = "https://github.com/tabokie/grpc-rs", branch = "v0.10.x-win" }
 
 [workspace]
 members = ["stress", "ctl"]
--- a/Makefile
+++ b/Makefile
@@ -67,6 +67,8 @@ else
 test_matrix: test
 	cargo ${TOOLCHAIN_ARGS} test --all ${EXTRA_CARGO_ARGS} -- --nocapture
 	cargo ${TOOLCHAIN_ARGS} test --test failpoints --features failpoints ${EXTRA_CARGO_ARGS} -- --test-threads 1 --nocapture
+	cargo ${TOOLCHAIN_ARGS} test --all --features nightly_group,std_fs ${EXTRA_CARGO_ARGS} -- --nocapture
+	cargo ${TOOLCHAIN_ARGS} test --test failpoints --features nightly_group,std_fs,failpoints ${EXTRA_CARGO_ARGS} -- --test-threads 1 --nocapture
 endif
 
 ## Build raft-engine-ctl.
--- a/src/env/log_fd.rs
+++ b/src/env/log_fd.rs
@@ -1,0 +1,11 @@
+// Copyright (c) 2017-present, PingCAP, Inc. Licensed under Apache-2.0.
+
+#[cfg(not(any(windows, feature = "std_fs")))]
+mod unix;
+#[cfg(not(any(windows, feature = "std_fs")))]
+pub use unix::LogFd;
+
+#[cfg(any(windows, feature = "std_fs"))]
+mod plain;
+#[cfg(any(windows, feature = "std_fs"))]
+pub use plain::LogFd;
--- a/src/env/log_fd/plain.rs
+++ b/src/env/log_fd/plain.rs
@@ -1,0 +1,81 @@
+// Copyright (c) 2017-present, PingCAP, Inc. Licensed under Apache-2.0.
+
+use crate::env::{Handle, Permission};
+
+use fail::fail_point;
+use parking_lot::RwLock;
+
+use std::fs::{File, OpenOptions};
+use std::io::{Error, ErrorKind, Read, Result, Seek, SeekFrom, Write};
+use std::path::Path;
+use std::sync::Arc;
+
+pub struct LogFd(Arc<RwLock<File>>);
+
+impl LogFd {
+    pub fn open<P: AsRef<Path>>(path: P, _: Permission) -> Result<Self> {
+        OpenOptions::new()
+            .read(true)
+            .write(true)
+            .open(path)
+            .map(|x| Self(Arc::new(RwLock::new(x))))
+    }
+
+    pub fn create<P: AsRef<Path>>(path: P) -> Result<Self> {
+        OpenOptions::new()
+            .create(true)
+            .read(true)
+            .write(true)
+            .open(path)
+            .map(|x| Self(Arc::new(RwLock::new(x))))
+    }
+
+    pub fn read(&self, offset: usize, buf: &mut [u8]) -> Result<usize> {
+        let mut file = self.0.write();
+        let _ = file.seek(SeekFrom::Start(offset as u64))?;
+        file.read(buf)
+    }
+
+    pub fn write(&self, offset: usize, content: &[u8]) -> Result<usize> {
+        fail_point!("log_fd::write::no_space_err", |_| {
+            Err(Error::new(ErrorKind::Other, "nospace"))
+        });
+
+        let mut file = self.0.write();
+        let _ = file.seek(SeekFrom::Start(offset as u64))?;
+        file.write(content)
+    }
+
+    pub fn truncate(&self, offset: usize) -> Result<()> {
+        let file = self.0.write();
+        file.set_len(offset as u64)
+    }
+
+    pub fn allocate(&self, _offset: usize, _size: usize) -> Result<()> {
+        Ok(())
+    }
+}
+
+impl Handle for LogFd {
+    fn truncate(&self, offset: usize) -> Result<()> {
+        self.truncate(offset)
+    }
+
+    fn file_size(&self) -> Result<usize> {
+        fail_point!("log_fd::file_size::err", |_| {
+            Err(Error::new(ErrorKind::InvalidInput, "fp"))
+        });
+
+        let file = self.0.read();
+        file.metadata().map(|x| x.len() as usize)
+    }
+
+    fn sync(&self) -> Result<()> {
+        fail_point!("log_fd::sync::err", |_| {
+            Err(Error::new(ErrorKind::InvalidInput, "fp"))
+        });
+
+        let file = self.0.write();
+        file.sync_all()
+    }
+}
--- a/src/file_pipe_log/pipe.rs
+++ b/src/file_pipe_log/pipe.rs
@@ -175,6 +175,10 @@ impl<F: FileSystem> SinglePipe<F> {
     /// filesystem.
     fn sync_dir(&self, path_id: PathId) -> Result<()> {
         debug_assert!(!self.paths.is_empty());
+
+        // Skip syncing directory in Windows. Refer to badger's discussion for more
+        // detail: https://github.com/dgraph-io/badger/issues/699
+        #[cfg(not(windows))]
         std::fs::File::open(PathBuf::from(&self.paths[path_id])).and_then(|d| d.sync_all())?;
         Ok(())
     }
--- a/src/fork.rs
+++ b/src/fork.rs
@@ -1,9 +1,13 @@
 // Copyright (c) 2023-present, PingCAP, Inc. Licensed under Apache-2.0.
 
 use std::fs::{copy, create_dir_all};
-use std::os::unix::fs::symlink;
 use std::path::Path;
 use std::sync::Arc;
+
+#[cfg(not(windows))]
+use std::os::unix::fs::symlink;
+#[cfg(windows)]
+use std::os::windows::fs::symlink_file as symlink;
 
 use crate::config::{Config, RecoveryMode};
 use crate::env::FileSystem;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -16,6 +16,8 @@
 #![cfg_attr(feature = "nightly", feature(test))]
 #![cfg_attr(feature = "swap", feature(allocator_api))]
 #![cfg_attr(feature = "swap", feature(slice_ptr_get))]
+// Though the new nightly rust stablized this feature, keep it anyway
+// because some other project (like TiKV) is still using the old.
 #![cfg_attr(feature = "swap", feature(nonnull_slice_from_raw_parts))]
 #![cfg_attr(feature = "swap", feature(slice_ptr_len))]
 #![cfg_attr(feature = "swap", feature(alloc_layout_extra))]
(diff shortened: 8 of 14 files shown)


MichaelScofield:
@tabokie PTAL

src/fork.rs
```diff
@@ -1,7 +1,6 @@
 // Copyright (c) 2023-present, PingCAP, Inc. Licensed under Apache-2.0.
 
 use std::fs::{copy, create_dir_all};
-use std::os::unix::fs::symlink;
 use std::path::Path;
 use std::sync::Arc;
```
tabokie, line 6: Maybe import two symlink here? I see the logging code is duplicated.
```rust
#[cfg(target_os = "windows")]
use std::os::windows::fs::symlink_file as symlink;
#[cfg(not(target_os = "windows"))]
use std::os::unix::fs::symlink;
```

src/env/log_fd.rs
```diff
@@ -0,0 +1,11 @@
+// Copyright (c) 2017-present, PingCAP, Inc. Licensed under Apache-2.0.
+
+#[cfg(any(target_os = "unix", target_os = "linux"))]
```
tabokie, line 3: I believe `target_os = "unix"` is a noop. The correct one should be `cfg(target_family = "unix")` or simply `cfg(unix)`.

tabokie: Also, we need a user feature flag to force enable std. Because our CI only run coverage check on Linux, so we must enable this flag to cover the newly added code.

I wish we can hide this flag from user docs though. The next best option is to just name it like "_use_std_fs" or something.

src/env/log_fd.rs
```diff
@@ -0,0 +1,11 @@
...
+mod nix;
+#[cfg(any(target_os = "unix", target_os = "linux"))]
+pub use nix::LogFd;
+
+#[cfg(not(any(target_os = "unix", target_os = "linux")))]
+mod plain;
```
tabokie, line 9: I think s/plain/std is okay, consistent with nix.

src/env/log_fd/nix.rs
```diff
@@ -0,0 +1,206 @@
...
+/// A [`LogFd`] is essentially a thin wrapper around [`RawFd`]. It's only
+/// supported on *Unix*, and primarily optimized for *Linux*.
+///
+/// All [`LogFd`] instances are opened with read and write permission.
+pub struct LogFd(RawFd);
+
+impl LogFd {
+    /// Opens a file with the given `path`.
+    pub fn open<P: ?Sized + NixPath>(path: &P, perm: Permission) -> IoResult<Self> {
+        fail_point!("log_fd::open::err", |_| {
```
tabokie, line 43: This is tricky. All the failpoints need to be ported as well otherwise some unit tests can fail.

tabokie (commented):
Some CI changes that're needed:

- Add windows to https://github.com/tikv/raft-engine/blob/1c7a820b251e7d012f463e1aee722eac5b5a0440/.github/workflows/rust.yml#L13
- Makefile/Cargo.toml needs some heavy updates so that the newly introduced "_use_std_fs" is only enabled when running `make test_with_std_fs`, and disabled otherwise
- Add `make test_with_std_fs` to the coverage pipeline: https://github.com/tikv/raft-engine/blob/1c7a820b251e7d012f463e1aee722eac5b5a0440/.github/workflows/rust.yml#L100

src/env/log_fd.rs
MichaelScofield: Was named `mod std` until rust complained about naming clash with the its own "std".

src/env/log_fd/nix.rs
MichaelScofield: You mean the fail points should be enabled for both LogFd impl? WDYT if I made a new LogFdImpl trait, and use different impl in different env? Like this:

```rust
// in log_fd.rs:

trait LogFdImpl {
  fn foo(&self);
}

struct LogFd(Arc<dyn LogFdImpl>)

impl LogFd {
  fn foo(&self) {
    fail_points!(...)
    self.0.foo();
  }
}

// in log_fd/nix.rs:

struct NixLogFd

impl LogFdImpl for NixLogFd 

// in log_fd/plain.rs:

struct StdLogFd

impl LogFdImpl for StdLogFd

// then in log_fd.rs:

impl LogFd {
  fn open() -> Self {
    #[cfg(unix)]
    LogFd(Arc::new(logfd::nix::NixLogFd))
    #[cfg(not(unix))]
    LogFd(Arc::new(logfd::plain::StdLogFd))
  }
}
#[cfg(unix)]

```

tabokie: That seems too cumbersome. We can either port failpoints, or disable failpoint tests when std is used (using required-features):

https://github.com/tikv/raft-engine/blob/1c7a820b251e7d012f463e1aee722eac5b5a0440/Cargo.toml#L23-L26

tabokie: Oh, we can still move some of them to `LogFile` implementation. Only failpoints associated with methods from trait `Handle` is required to be implemented for `LogFd`.

MichaelScofield: Not sure I've fully understood it, sorry, but do you mean: 1. the current `LogFd` implementations (called by same function signatures, enabled under different `cfg`s) are fine, don't need to be further changed? 2. I have to port the fail points in log_fd::plain, too?

tabokie: No,
1. (optionally) You can move some failpoints under `LogFd` to `LogFile`. Because `LogFd` only implements one public interface `Handle`, other methods are only called by `LogFile`. So moving those not associated with `Handle` will be not change the test behavior.
2. Duplicate the remaining failpoints under `LogFd` to `plain::LogFd`.

src/env/log_fd.rs
tabokie: Yes

MichaelScofield:
@tabokie PTAL. Can you approve the workflow to run so I can debug the windows build in CI?

MichaelScofield:
@tabokie can you make this PR always run workflow? Need another approval again. Debugging github workflow is a pain, especially for os = windows. I can see more commits to debug it in the future.

tabokie:
I don't have the permission to do that. It's weird though because my fork repo runs the workflow just fine: https://github.com/tabokie/raft-engine/actions, I wonder why your fork repo ignores the workflow file.

tabokie:
@MichaelScofield You can work around this by merging another small PR first. E.g. Open another PR to update clap manually (https://github.com/tikv/raft-engine/pull/306)

tabokie:
It can build now, you just need to fix the tests.

MichaelScofield:
Running into infamous "access denied (os error 5)" error in windows. Convert this PR to draft first, to reduce github notification noise.

MichaelScofield:
@tabokie PTAL. The tests are all passed (at least in my repo's action, see [here](https://github.com/MichaelScofield/raft-engine/pull/1)). Let's approve the workflow in this PR to see how it goes!

Cargo.toml
```diff
@@ -86,16 +90,26 @@ swap = [
   "nightly",
   "memmap2",
 ]
+std_fs = []
 
 # Shortcuts
-all_except_failpoints = ["internals", "scripting", "nightly", "swap"]
-all_stable = ["internals", "scripting", "failpoints"]
-all_stable_except_failpoints = ["internals", "scripting"]
+all_unix_stable_without_failpoints = ["internals", "scripting"]
```
tabokie, line 96: These look a bit messy. I'm cleaning them up in https://github.com/tikv/raft-engine/pull/325. Once it is merged you can add std_fs to Makefile `test_matrix` without polluting other build options.

Cargo.toml
```diff
@@ -86,16 +90,26 @@ swap = [
...
+all_stable = ["all_stable_without_failpoints", "failpoints"]
+
+all_nightly_without_failpoints = ["internals", "scripting", "nightly", "swap", "std_fs"]
+all_nightly = ["all_nightly_without_failpoints", "failpoints"]
 
 [patch.crates-io]
 raft-proto = { git = "https://github.com/tikv/raft-rs", branch = "master" }
 protobuf = { git = "https://github.com/pingcap/rust-protobuf", branch = "v2.8" }
 protobuf-codegen = { git = "https://github.com/pingcap/rust-protobuf", branch = "v2.8" }
+grpcio = { git = "https://github.com/tabokie/grpc-rs", branch = "v0.10.x-win" }
```
tabokie, line 112: Add a comment to explain the build error. Ref https://github.com/tikv/grpc-rs/pull/622.

src/file_pipe_log/pipe.rs
```diff
@@ -175,6 +175,23 @@ impl<F: FileSystem> SinglePipe<F> {
     /// filesystem.
     fn sync_dir(&self, path_id: PathId) -> Result<()> {
         debug_assert!(!self.paths.is_empty());
+
+        // Cannot open directory in Windows, will end up with
+        // "Access Denied(OS Error 5)". We have to walk the
+        // directory to sync each file in it.
```
tabokie, line 181: This seems terrible in performance. Need a FIXME.

src/env/log_fd/plain.rs
```diff
@@ -0,0 +1,81 @@
...
+
+use fail::fail_point;
+use parking_lot::RwLock;
+
+use std::fs::{File, OpenOptions};
+use std::io::{Error, ErrorKind, Read, Result, Seek, SeekFrom, Write};
+use std::path::Path;
+use std::sync::Arc;
+
+pub struct LogFd(Arc<RwLock<File>>);
```
tabokie, line 13: This implementation kind of defeats the point of `Handle` abstraction, which is to be efficiently shared between multiple readers/writers. I like seeing a fallback solution that covers all platforms other than Windows, but this implementation is too inefficient for production use.

I can think of two ways to go around this:

(1) use raw windows HANDLE here like in https://github.com/tikv/raft-engine/pull/69, it has the downside of windows-only.
(2) further separate `LogFile` into unix and windows version. The windows version looks something like below. It has the downside of needing to port lots of failpoints again.
```rust
struct LogFd(File);
struct LogFile(File);
impl LogFd {
  fn new_writer(&self) -> LogFile {
    LogFile(self.0.try_clone().unwrap())
  }
}
```

tabokie (dismissed):

tabokie:
@MichaelScofield #325 is merged. Feel free to ask questions. https://github.com/tikv/raft-engine/pull/322#discussion_r1265127742 might be too much work for this PR, you can follow up in another one or let me handle it later.

MichaelScofield:
@tabokie Sorry for the late reply, busy week. 

I actually considered using the raw Windows API earlier. However, skimmed through the "[windows](https://microsoft.github.io/windows-docs-rs/doc/windows/index.html)" crate and the "[official win32 API docs](https://learn.microsoft.com/en-us/windows/win32/)", I find it's not a quick and easy task. It requires detailed understanding of Windows filesystem to do it well, which I'm currently lacking of, nor do I have the time to study it. Besides, the windows binding APIs are all "unsafe" to call in rust. So I decide to switch back to the good old std fs, make it right first.

I'll make the features worked in the new test matrix this weekend.

src/file_pipe_log/pipe.rs
MichaelScofield: Question: do we really need to sync dir, and need to do it under windows?

MichaelScofield:
@tabokie tests are ok now, PTAL

src/file_pipe_log/pipe.rs
tabokie: Let's remove it. [Apparently](https://learn.microsoft.com/en-us/previous-versions/windows/it-pro/windows-server-2003/cc781134(v=ws.10)?redirectedfrom=MSDN#within-a-directory-entry-for-a-file) windows has the concept of directory metadata too, but they choose not to say anything about its persistence. I guess we can assume they are flushed to disk the moment file is created, like [badger](https://github.com/dgraph-io/badger/issues/699) did.

.github/workflows/rust.yml
```diff
@@ -35,10 +35,10 @@ jobs:
       - name: Clippy
         run: make clippy
       - name: Run tests
-        run: make test
+        run: make test_matrix
```
tabokie, line 38: No need, run it for coverage is enough.

src/lib.rs
```diff
@@ -16,7 +16,6 @@
 #![cfg_attr(feature = "nightly", feature(test))]
 #![cfg_attr(feature = "swap", feature(allocator_api))]
 #![cfg_attr(feature = "swap", feature(slice_ptr_get))]
-#![cfg_attr(feature = "swap", feature(nonnull_slice_from_raw_parts))]
```
tabokie, line 19: Keep it for now, older nightly compiler (like the one TiKV is using) hasn't opt-in this feature yet.

Makefile
```diff
@@ -44,9 +44,9 @@ format:
...
 clippy:
 ifdef WITH_NIGHTLY_FEATURES
-	cargo ${TOOLCHAIN_ARGS} clippy --all --features nightly_group,failpoints --all-targets -- -D clippy::all
+	cargo ${TOOLCHAIN_ARGS} clippy --all --features nightly_group,failpoints,std_fs --all-targets -- -D clippy::all
 else
-	cargo ${TOOLCHAIN_ARGS} clippy --all --features failpoints --all-targets -- -D clippy::all
+	cargo ${TOOLCHAIN_ARGS} clippy --all --features failpoints,std_fs --all-targets -- -D clippy::all
 endif
 
 ## Run tests.
```
tabokie, line 52: Need to update "test_matrix" build entry to include the new feature.

MichaelScofield:
@tabokie PTAL

Now the "test_matrix" is only used in coverage job, and only runs under ubuntu.

.github/workflows/rust.yml
```diff
@@ -101,7 +101,7 @@ jobs:
         env:
           RUSTFLAGS: '-Zinstrument-coverage'
           LLVM_PROFILE_FILE: '%p-%m.profraw'
-          EXTRA_CARGO_ARGS: '--verbose'
+          EXTRA_CARGO_ARGS: '--features std_fs'
```
tabokie, line 104: No no, I meant modifying Makefile. `test_matrix` is an entry dedicated for testing different feature combinations.

tabokie (commented):
Rest LG.

.github/workflows/rust.yml
MichaelScofield: Are you suggesting define a matrix like this: https://github.com/MichaelScofield/raft-engine/commit/61ccf8d14666cdbfb340a73ec67d6fbe4503525b ?

tabokie: No, add "cargo test --features nightly_group,failpoints,std_fs xxx" to https://github.com/tikv/raft-engine/blob/58041ad3e69ae76941309733bba936743bf8b5a5/Makefile#L67

No need to update rust.yml.

MichaelScofield: You mean like this https://github.com/MichaelScofield/raft-engine/commit/bdc6d68860563827df26711dd269b36726e810b4 ?

Makefile
```diff
@@ -65,8 +65,16 @@ test_matrix:
...
+	cargo ${TOOLCHAIN_ARGS} test --all --features nightly_group ${EXTRA_CARGO_ARGS} -- --nocapture
+	cargo ${TOOLCHAIN_ARGS} test --all --features failpoints ${EXTRA_CARGO_ARGS} -- --nocapture
+	cargo ${TOOLCHAIN_ARGS} test --all --features std_fs ${EXTRA_CARGO_ARGS} -- --nocapture
+	cargo ${TOOLCHAIN_ARGS} test --all --features nightly_group,failpoints ${EXTRA_CARGO_ARGS} -- --nocapture
+	cargo ${TOOLCHAIN_ARGS} test --all --features nightly_group,std_fs ${EXTRA_CARGO_ARGS} -- --nocapture
+	cargo ${TOOLCHAIN_ARGS} test --all --features failpoints,std_fs ${EXTRA_CARGO_ARGS} -- --nocapture
+	cargo ${TOOLCHAIN_ARGS} test --all --features nightly_group,failpoints,std_fs ${EXTRA_CARGO_ARGS} -- --nocapture
+	cargo ${TOOLCHAIN_ARGS} test --test failpoints --features failpoints,nightly_group ${EXTRA_CARGO_ARGS} -- --test-threads 1 --nocapture
+	cargo ${TOOLCHAIN_ARGS} test --test failpoints --features failpoints,std_fs ${EXTRA_CARGO_ARGS} -- --test-threads 1 --nocapture
+	cargo ${TOOLCHAIN_ARGS} test --test failpoints --features failpoints,nightly_group,std_fs ${EXTRA_CARGO_ARGS} -- --test-threads 1 --nocapture
```
tabokie, line 77:
```suggestion
	cargo ${TOOLCHAIN_ARGS} test --all ${EXTRA_CARGO_ARGS} -- --nocapture
	cargo ${TOOLCHAIN_ARGS} test --test failpoints --features failpoints ${EXTRA_CARGO_ARGS} -- --test-threads 1 --nocapture
	cargo ${TOOLCHAIN_ARGS} test --all --features nightly_group,std_fs ${EXTRA_CARGO_ARGS} -- --nocapture
	cargo ${TOOLCHAIN_ARGS} test --test nightly_group,std_fs,failpoints --features failpoints ${EXTRA_CARGO_ARGS} -- --test-threads 1 --nocapture
```

tabokie (commented):
DCO needs fixing.

MichaelScofield:
@tabokie PTAL

tabokie:
GitHub action space is not enough.. I'll take a look tomorrow.

tabokie (approved):
LGTM, Thanks

PR #323, opened 2023-07-07
Introduce double write file system

Connor1996:


src/env/double_write.rs
```diff
@@ -0,0 +1,722 @@
...
+    Sync,
+    Write(usize),
+    Allocate,
+    NewWriter(F::Writer),
+}
+
+pub struct HedgedHandle<F: FileSystem> {
+    disk1: Sender<(FileTask, Callback<usize>)>,
+    disk2: Sender<(FileTask, Callback<usize>)>,
+    counter1: Arc<AtomicU64>,
```
coderplay, line 454: What do counters indicate?

src/env/double_write.rs
```diff
@@ -0,0 +1,722 @@
...
+            res2 = f2 => res2.unwrap().map(|w| HedgedHandle::new(
+                Either::Left(f1), Either::Right(w.unwrap()) )),
+        }
+        select! {
+            res1 = f1 => res1.unwrap().map(|_| ()),
+            res2 = f2 => res2.unwrap().map(|_| ()),
+        }
+    }
+
+    fn read(&self, offset: usize, buf: &mut [u8]) -> IoResult<usize> {
```
coderplay, line 563: Are there concurrent readers/writers when operating raft logs?

src/env/double_write.rs
```diff
@@ -0,0 +1,722 @@
...
+            std::cmp::Ordering::Greater => {
+                self.fd1.read().unwrap().as_ref().unwrap().read(offset, buf)
+            }
+            std::cmp::Ordering::Less => {
+                self.fd2.read().unwrap().as_ref().unwrap().read(offset, buf)
+            }
+        }
+    }
+
+    fn write(&self, offset: usize, content: &[u8]) -> IoResult<usize> {
```
coderplay, line 587: Is that possible that when we do a write, there is an another thread that's doing read?

Connor1996: yes

coderplay: Since based on the design, we achieve an eventual consistency on the two disks.  Meaning, at a time,  the content of the two disks can be different.   If the writer is writing disk 1,  while the reader read from disk 2,  looks like the raft engine would make a wrong action based on the wrong point of view.

src/env/double_write.rs
```diff
@@ -0,0 +1,722 @@
...
+    fn allocate(&mut self, offset: usize, size: usize) -> IoResult<()> {
+    }
+}
+
+impl<F: FileSystem> Seek for HedgedWriter<F> {
+    fn seek(&mut self, pos: SeekFrom) -> IoResult<u64> {
+        match pos {
+            SeekFrom::Start(offset) => self.offset = offset as usize,
+            SeekFrom::Current(i) => self.offset = (self.offset as i64 + i) as usize,
+            SeekFrom::End(i) => self.offset = (self.inner.file_size()? as i64 + i) as usize,
```
coderplay, line 673: The file size on the two disks can be different, right?  Looks like it will cause problems

src/env/double_write.rs
```diff
@@ -0,0 +1,722 @@
...
+        }
+        Ok(self.offset as u64)
+    }
+}
+
+pub struct HedgedReader<F: FileSystem> {
+    inner: Arc<HedgedHandle<F>>,
+    offset: usize,
```
coderplay, line 681: We don't add any concurrency protection here on the `offset`,  does that mean the offset for the files on two disks are equivalent ?

coderplay: I remember that from the design doc, you mentioned the content of the two disks can be different.  IIUC,  in the case of raft logs being purged on one disk while not yet be purged on the on the other disk,  the `inner: Arc<HedgedHandle<F>>` can points to one file, and points to nothing in another run.

src/env/double_write.rs
```diff
@@ -0,0 +1,722 @@
...
+            SeekFrom::End(i) => self.offset = (self.inner.file_size()? as i64 + i) as usize,
+        }
+        Ok(self.offset as u64)
+    }
+}
+
+impl<F: FileSystem> Read for HedgedReader<F> {
+    fn read(&mut self, buf: &mut [u8]) -> IoResult<usize> {
+        let len = self.inner.read(self.offset, buf)?;
+        self.offset += len;
```
coderplay, line 707: you mentioned that there might be concurrent readers,  is it safe doing `self.offset += len` from multiple threads?

Connor1996: Note that there would be multiple readers, but for one reader, it's not shared between threads. It doesn't implement `Sync` trait, so the rust compiler would make sure it is not shared between threads.

src/env/double_write.rs
Connor1996: No need for concurrency protection, cause it's not `Sync`. Rust compiler would take care of it

src/env/double_write.rs
Connor1996: Note the offset is specified. The reader always reads from the newer disk at the point of time. If the reader can read from disk2, it indicates that disk2 must have the data of that offset. It doesn't need to care about the latest write because it's not reading the latest writes' offset

src/env/double_write.rs
Connor1996: It would always return the file size of the newer disk. The result on each call for a being appended file must be monotonic increasing. lt has nothing different compared to the non-hedging file system.

src/env/double_write.rs
```diff
@@ -0,0 +1,722 @@
...
+                    panic!("Both fd1 and fd2 are None");
+                }
+            }
+            std::cmp::Ordering::Greater => {
+                self.fd1.read().unwrap().as_ref().unwrap().read(offset, buf)
+            }
+            std::cmp::Ordering::Less => {
+                self.fd2.read().unwrap().as_ref().unwrap().read(offset, buf)
+            }
+        }
```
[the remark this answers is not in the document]
Connor1996: Yes, as long as the disk has the data, then you can read it. Any concerns?

src/env/double_write.rs
```diff
@@ -0,0 +1,722 @@
...
+                cb(res);
+            }
+        });
+        let fs2 = base.clone();
+        let handle2 = thread::spawn(|| {
+            for (task, cb) in rx2 {
+                if task == Task::Stop {
+                    break;
+                }
+                let res = Self::process(&fs2, task);
```
coderplay, line 101: why doesn't set a fail_point of this handle?

src/env/double_write.rs
```diff
@@ -0,0 +1,722 @@
...
+            disk1: tx1,
+            disk2: tx2,
+            handle1: Some(handle1),
+            handle2: Some(handle2),
+        }
+    }
+
+    pub fn bootstrap(&self) -> Result<()> {
+        // catch up diff
+        let files1 = self.get_files(&self.path1)?;
```
coderplay, line 118: Is it possible that files1 is a set of {f1, f2, f3}, while file2 is {f2, f3, f4} ?

src/env/double_write.rs
```diff
@@ -0,0 +1,722 @@
...
+        let files2 = self.get_files(&self.path2)?;
```
coderplay, line 119: In the case of `enable-log-recycle` is enabled, IIUC,  file names are reused. Meaning the same file name might hold different raft log content over time. Is it possible that for the same file name `f`,  the file on disk1 and the one on disk2 hold different contents?

src/env/double_write.rs
coderplay: @Connor1996  could you please reply my 2nd comment?

src/env/double_write.rs
coderplay: Understand,  looks like it's a sole append-only writer + multiple readers design

src/env/double_write.rs
```diff
@@ -0,0 +1,722 @@
...
+            res2 = f2 => res2.unwrap().map(|_| ()),
+        }
+    }
+
+    fn read(&self, offset: usize, buf: &mut [u8]) -> IoResult<usize> {
+        // TODO: read simultaneously from both disks
+        // choose latest to perform read
+        let count1 = self.counter1.load(Ordering::Relaxed);
+        let count2 = self.counter2.load(Ordering::Relaxed);
+        match count1.cmp(&count2) {
```
coderplay, line 568: There are 3 steps here before reading a file: 
1. get counter1
2. get counter2
3. compare counter1 and counter2

The 3 steps together are not atomic. If we get an wrong result that counter1 < counter2 due to the in-atomicity, and the offset > file size for counter1, what will happen?

src/env/double_write.rs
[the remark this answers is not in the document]
coderplay: two cases: 

1) offset > file size
2) fd is null for one of the disks

src/env/double_write.rs
Connor1996: Impossible, log-recycle doesn't reuse file name.

src/env/double_write.rs
Connor1996: It's used to slow one disk not both

src/env/double_write.rs
```diff
@@ -0,0 +1,722 @@
...
+
+// TODO: read both dir at recovery, maybe no need? cause operations are to both
+// disks TODO: consider encryption
+
+impl<F: FileSystem> HedgedFileSystem<F> {
+    pub fn new(base: Arc<F>, path1: PathBuf, path2: PathBuf) -> Self {
+        let (tx1, rx1) = unbounded::<(Task, Callback<F::Handle>)>();
+        let (tx2, rx2) = unbounded::<(Task, Callback<F::Handle>)>();
+        let fs1 = base.clone();
+        let handle1 = thread::spawn(|| {
```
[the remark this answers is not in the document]
Connor1996: Crash would fire to the whole process not only the thread

src/env/double_write.rs
Connor1996: yes

src/env/double_write.rs
Connor1996: It doesn't matter.
Let's say disk2 is the slower one, and its file size < offset
- first get s1 = counter1, we can know that s1 >= offset >= file size,
- then get s2 = counter2, 
  - if s1 <= s2, we read on disk2, because s2 >= s1 >= offset >= file size, disk2 must have the data now
  - if s1 > s2, we read on disk1

src/env/double_write.rs
Connor1996: Yes, it's possible to point to one file, and the other points to nothing. Still doesn't matter, if it's purged which means nothing would visit it anymore.

src/env/double_write.rs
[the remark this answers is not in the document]
Connor1996: 1. Impossible for raft engine logic, How can you read data that you haven't ever written? 
2. Fd is null means the file is purged, which further means no one would visit the file anymore.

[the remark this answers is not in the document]
coderplay: 1. Why are you so sure the `offset` is always <= `file size` ?  Is there any contract on the layer above the FileSystem that guarantee it?  The upper layer `LogFileReader ` expose below public API, from which we can pass arbitrary value to `offset`  that might be larger than the filesize.

```
    /// Polls bytes from the file. Stops only when the buffer is filled or
    /// reaching the "end of file".
    pub fn read_to(&mut self, offset: u64, mut buf: &mut [u8]) -> Result<usize> {
```

2. see the below comment regarding fd for one disk is non-null, while for another disk is null

src/env/double_write.rs
coderplay: My overall impression of this PR is that we are leveraging certain undocumented invariants, e.g.  sole append-only writer plus multi-readers,  in the upper layers to ensure the safety of concurrency in the lower-level HedgedFileSystem. Two drawbacks I can think of
* There are many variants in the FileSystem layer,  for example the read-write conflicts,  in-atomicity of operations, etc. It's complicated to cover all the cases, and it make the code hard to be understood.
* In the future, if any engineer is unaware of these hidden relationships, it could easily lead to significant disasters.

coderplay:
My overall impression of this PR is that we are leveraging certain undocumented invariants, e.g. sole append-only writer plus multi-readers, in the upper layers to ensure the safety of concurrency in the lower-level HedgedFileSystem. Two drawbacks I can think of

There are many variants in the FileSystem layer, for example the read-write conflicts, in-atomicity of operations, etc. It's complicated to cover all the cases, and it make the code hard to be understood.
In the future, if any engineer is unaware of these hidden relationships, it could easily lead to significant disasters.

src/env/double_write.rs
coderplay: oh my bad, my understanding was wrong. How does log recycling work?

src/env/double_write.rs
coderplay: The fd may not be meaningful for the disk which is purged, but it's meaningful for the other disk which is not yet purged.

coderplay:
In a scenario where a fast disk wrote transactions 1, 2, and 3, while the slow disk only recorded transaction 1.  If the fast disk is lost somehow,  and only transaction 1 remaining on the slow disk, data consistency problems may arise. Right?

Connor1996:
> In a scenario where a fast disk wrote transactions 1, 2, and 3, while the slow disk only recorded transaction 1. If the fast disk is lost somehow, and only transaction 1 remaining on the slow disk, data consistency problems may arise. Right?

It doesn't improve the data durability, when either one of disks is lost or corrupted, raft engine should refuse to start. So, it never has the chance to expose data inconsistency.

src/env/hedged/mod.rs
```diff
@@ -0,0 +1,671 @@
...
+mod task;
+mod util;
+
+use runner::TaskRunner;
+use sender::HedgedSender;
+use task::{
+    empty_callback, paired_future_callback, Callback, FutureHandle, SeqTask, Task, TaskRes,
+};
+use util::replace_path;
+
```
LykxSassinator, line 34: Refer to https://github.com/tikv/raft-engine/pull/339, this import format should be reformatted.

LykxSassinator (commented):
More uts should be added to HedgeFileSystem.

src/config.rs
```diff
@@ -40,6 +40,8 @@ pub struct Config {
     /// Default: None
     pub spill_dir: Option<String>,
 
+    pub second_dir: Option<String>,
```
tabokie, line 43: maybe backup_dir or mirror_dir

src/env/hedged/recover.rs
```diff
@@ -0,0 +1,255 @@
...
+    pub fn into_handle(mut self, file_system: &DefaultFileSystem) -> Self {
+        if let SeqFile::Path(f) = self {
+            let fd = Arc::new(file_system.open(&f.path, Permission::ReadOnly).unwrap());
+            self = SeqFile::Handle((f, fd));
+        }
+        self
+    }
+}
+
+pub(crate) fn catch_up_diff(
```
tabokie, line 78: "if the file in to is not in from, delete it"
```suggestion
pub(crate) fn synchronize_files(
```

src/env/hedged/mod.rs
```diff
@@ -0,0 +1,671 @@
...
+            info!("Create raft log directory: {}", self.path2.display());
+            fs::create_dir(&self.path2).unwrap();
+        }
+        let files1 = recover::get_files(&self.path1)?;
+        let files2 = recover::get_files(&self.path2)?;
+
+        let count1 = recover::get_latest_valid_seq(&self.base, &files1)?;
+        let count2 = recover::get_latest_valid_seq(&self.base, &files2)?;
+
+        match count1.cmp(&count2) {
```
tabokie, line 270: `get_latest_valid_seq` returns the number of log items from latest file? Why is it used to determine the synchronize direction?

src/env/hedged/mod.rs
```diff
@@ -0,0 +1,671 @@
...
+    seqno1: Arc<AtomicU64>,
+    seqno2: Arc<AtomicU64>,
+
+    thread1: Option<JoinHandle<()>>,
+    thread2: Option<JoinHandle<()>>,
+}
+
+impl HedgedFileSystem {
+    pub fn new(base: Arc<DefaultFileSystem>, path1: PathBuf, path2: PathBuf) -> Self {
+        let (tx1, rx1) = unbounded::<(SeqTask, Callback)>();
```
tonyxuqqi, line 124: Should we limit the length of the channel to prevent the OOM, which is the hard limit  HedgedFileSystem can use

src/env/hedged/mod.rs
```diff
@@ -0,0 +1,671 @@
...
+        select! {
+            res1 = f1 => res1.unwrap().map(|_| ()),
+            res2 = f2 => res2.unwrap().map(|_| ()),
+        }
+    }
+}
+
+impl Drop for HedgedFileSystem {
+    fn drop(&mut self) {
+        block_on(self.wait(Task::Stop, Task::Stop)).unwrap();
```
tonyxuqqi, line 218: Should it wait for both Stop finished?

src/env/hedged/mod.rs
```diff
@@ -0,0 +1,671 @@
...
+                t1.join().unwrap();
+                t2.join().unwrap();
+                break;
+            }
+            times += 1;
+            if times > 100 {
+                // wait 1s
+                // one disk may be blocked for a long time,
+                // to avoid block shutdown process for a long time, do not join the threads
+                // here, only need at least to ensure one thread is exited
```
tonyxuqqi, line 234: HedgedFileSystem is dropped but its underlying thread may still live, which is not ideal. 

Maybe we should abort the pending tasks on slow disk and wait both threads exist.

src/env/hedged/util.rs
```diff
@@ -0,0 +1,9 @@
+use std::path::{Path, PathBuf};
+
+pub fn replace_path(path: &Path, from: &Path, to: &Path) -> PathBuf {
```
tonyxuqqi, line 3: The fn can be renamed as replace_prefix? Otherwise it's not obvious that parameter from must be prefix of path.

tonyxuqqi, line 3: The fn can be renamed as replace_prefix? Otherwise it's not obvious that parameter from must be prefix of path.

src/env/hedged/recover.rs
```diff
@@ -0,0 +1,255 @@
...
+        }
+        Ok(())
+    };
+
+    check_files(&from_files.append_files, &to_files.append_files)?;
+    if !skip_rewrite {
+        check_files(&from_files.rewrite_files, &to_files.rewrite_files)?;
+    }
+    check_files(&from_files.reserved_files, &to_files.reserved_files)?;
+    Ok(())
```
tonyxuqqi, line 164: We should double check if the files copy are correct by checksum.

src/env/hedged/sender.rs
```diff
@@ -0,0 +1,140 @@
...
+        let task2 = SeqTask {
+            inner: task2,
+            seq: inner.seq,
+        };
+        if matches!(inner.state, State::Normal) {
+            let check1 = inner.disk1.len() > get_pause_threshold();
+            let check2 = inner.disk2.len() > get_pause_threshold();
+            match (check1, check2) {
+                (true, true) => {
+                    panic!("Both channels of disk1 and disk2 are full")
```
tonyxuqqi, line 78: When both disks are slow, maybe should not panic but instead fallback to single disk approach.

src/engine.rs
```diff
@@ -74,12 +74,13 @@ where
...
     }
 
-    pub fn open_with(
+    fn open_with(
         mut cfg: Config,
         file_system: Arc<F>,
         mut listeners: Vec<Arc<dyn EventListener>>,
     ) -> Result<Engine<F, FilePipeLog<F>>> {
         cfg.sanitize()?;
+        file_system.bootstrap()?;
```
tonyxuqqi, line 83: I'm wondering if it can fallback to single disk solution dynamically.

tonyxuqqi (commented):
Overall LGTM

src/engine.rs
tonyxuqqi: Maybe in another PR

src/engine.rs
```diff
@@ -2629,6 +2630,284 @@ pub(crate) mod tests {
...
+    fn number_of_files(p: &Path) -> usize {
+        let mut r = 0;
+        std::fs::read_dir(p).unwrap().for_each(|e| {
+            if e.unwrap()
+                .path()
+                .file_name()
+                .unwrap()
+                .to_str()
+                .unwrap()
+                .starts_with("000")
```
tonyxuqqi, line 2642: is there a more specific file name pattern? "000" seems a bit too wild.

src/env/hedged/mod.rs
```diff
@@ -0,0 +1,671 @@
...
+            fs::create_dir(&self.path1).unwrap();
+        }
+        if !self.path2.exists() {
+            info!("Create raft log directory: {}", self.path2.display());
+            fs::create_dir(&self.path2).unwrap();
+        }
+        let files1 = recover::get_files(&self.path1)?;
+        let files2 = recover::get_files(&self.path2)?;
+
+        let count1 = recover::get_latest_valid_seq(&self.base, &files1)?;
```
tonyxuqqi, line 267: So here we only compare the last file's log count to decide which disk is newer?
What if their file counts are different and the older has more?

PR #326, opened 2023-08-09
metrics: supply the metric on storage usage.

LykxSassinator:
## Description
This pr is used to supply the necessary metric on storage usage of engine.

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -315,7 +315,10 @@ where
     /// Purges expired logs files and returns a set of Raft group ids that need
     /// to be compacted.
     pub fn purge_expired_files(&self) -> Result<Vec<u64>> {
-        self.purge_manager.purge_expired_files()
+        let purged = self.purge_manager.purge_expired_files();
+        // Update storage size after purge.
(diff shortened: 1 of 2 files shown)


tabokie (commented):
Why is this needed? `raft_engine_log_file_count * file_size` should be equivalent. If needed you can publish file_size via a separate counter.

LykxSassinator:
> Why is this needed? `raft_engine_log_file_count * file_size` should be equivalent. If needed you can publish file_size via a separate counter.

From users, they think the usage of raft-engine is not clear as expected. They expect the storage size can be easily checked.

tabokie:
> > Why is this needed? `raft_engine_log_file_count * file_size` should be equivalent. If needed you can publish file_size via a separate counter.
> 
> From users, they think the usage of raft-engine is not clear as expected. They expect the storage size can be easily checked.

You can either add a counter for raft-engine file_size config in TiKV, then use `file_count * file_size` to display the size. Or manually flush the metrics in TiKV's `EngineMetricsManager`. The information is already there, I don't think it's worth adding another counter to raft-engine.

PR #331, opened 2023-09-01
enlarge append recycle log threshold

glorv:
- This PR change the recycle log threshold from `(purge_threshold / target_file_size) + 2` to `(purge_threshold / target_file_size)*1.5` to avoid repeatedly creating and purging log files when tikv's log compaction has non-trivial latency.  
- Also changes the `prefill_limit`'s semantic meaning, after this change, it will means the total log files threshold instead of recycle files threshold.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -116,7 +116,7 @@ pub struct Config {
     pub prefill_for_recycle: bool,
 
     /// Maximum capacity for preparing log files for recycling when start.
-    /// If `None`, its size is equal to `purge-threshold`.
+    /// If `None`, its size is equal to `purge-threshold`*1.5.
     /// Only available for `prefill-for-recycle` is true.
     ///
     /// Default: None
--- a/src/config.rs
+++ b/src/config.rs
@@ -110,7 +110,7 @@ pub struct Config {
     pub prefill_for_recycle: bool,
 
     /// Maximum capacity for preparing log files for recycling when start.
-    /// If not `None`, its size is equal to `purge-threshold`.
+    /// If `None`, its size is equal to `purge-threshold`.
     /// Only available for `prefill-for-recycle` is true.
     ///
     /// Default: None
@@ -212,8 +212,8 @@ impl Config {
         }
         if self.enable_log_recycle && self.purge_threshold.0 >= self.target_file_size.0 {
             // (1) At most u32::MAX so that the file number can be capped into an u32
-            // without colliding. (2) Add some more file as an additional buffer to
-            // avoid jitters.
+            // without colliding. (2) Increase the threshold by 50% to add some more file
+            // as an additional buffer to avoid jitters.
             std::cmp::min(
                 (self.purge_threshold.0 / self.target_file_size.0) as usize * 3 / 2,
                 u32::MAX as usize,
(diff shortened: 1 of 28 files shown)


glorv:
@LykxSassinator PTAL thanks~

LykxSassinator (approved):
LGTM

src/config.rs
```diff
@@ -241,7 +241,7 @@ impl Config {
         if self.prefill_for_recycle && prefill_limit >= self.target_file_size.0 {
```
tabokie, line 241: When prefill_limit is None, use `recycle_capacity`.

src/file_pipe_log/pipe_builder.rs
```diff
@@ -528,13 +528,14 @@ impl<F: FileSystem> DualPipesBuilder<F> {
...
         let target_file_size = self.cfg.target_file_size.0 as usize;
-        let mut target = std::cmp::min(
-            self.cfg.prefill_capacity(),
-            self.cfg
-                .recycle_capacity()
-                .saturating_sub(self.append_files.len()),
-        );
-        let to_create = target.saturating_sub(self.recycled_files.len());
+        let prefill_target =
+            std::cmp::min(self.cfg.prefill_capacity(), self.cfg.recycle_capacity())
```
tabokie, line 532: This is not the same as old code.

glorv: Because the old code is not fit anymore after the recycle threshold is enlarged. From @LykxSassinator, the old logic is that `prefill_capacity` is the number of recycle files that should be pre-created even if there are already some append files; but the new logic changes the semantic to let `prefill_capacity` the total files number (recycle + append), I think this is more intuitive.  
In the general case, we expect the total files size is around the size of purge threshold but not the recycle threshold ( 1.5 * purge_threshold), so it's better not to pre-create too many redundant files.

tabokie: > we expect the total files size is around the size of purge threshold but not the recycle threshold ( 1.5 * purge_threshold)

I don't think so. We need to prepare the same amount of files as recycle-threshold. The condition where a large recycle-threshold is needed can also occur when raft-engine has just started.

glorv: I think there is a bug in the old code that it only count the recycle file to create prefill files count. While with the default config there is no problem because it is equal to the recycle threshold. But if the prefill threshold is smaller than recycle threshold, I can cause raft-engine to create more files after restarting which is unexpacted. One of the log recycle test case failed after I exlarged the recycle threshold.

tabokie, line 532: > the old logic is that prefill_capacity is the number of recycle files that should be pre-created even if there are already some append files; but the new logic changes the semantic to let prefill_capacity the total files number (recycle + append), I think this is more intuitive.

This doesn't seem more intuitive to me. The `prefill_limit` stands for the amount of data that user can tolerate to write before startup. It should not be subtracted with append files.

glorv: ok. I reverted this change as the default `purge_threshold` and `recycle_threshold` is still the same.

tabokie (approved):

PR #335, opened 2023-09-10
refactor: reduce scope of internal keys

### LINKED ISSUE #334: make the INTERNAL_KEY_PREFIX configurable
We run into a problem. We are trying to use `raft-engine` as our Kv store. However, Our existing business logic uses the keys with the prefix `__`.  How about let's make the INTERNAL_KEY_PREFIX configurable?

WenyXu:
Reduce the scope of internal keys 

close #334

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## [Unreleased]
 
 ### Behavior Changes
-* `LogBatch::put` and `LogBatch::put_message` allow to put key start with `__` but not equals the `__{ATOMIC_GROUP_KEY}`(0x95 0x95 0x01).
+* `LogBatch::put` and `LogBatch::put_message` allow to put key start with `__` but not equals the `__{ATOMIC_GROUP_KEY}`(0x5f 0x5f 0x01).
 
 ## [0.4.0] - 2023-09-01
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## [Unreleased]
 
 ### Behavior Changes
-* `LogBatch::put` and `LogBatch::put_message` allow to put key start with `__` but not equals the `__{ATOMIC_GROUP_KEY}`(0x5f 0x5f 0x01).
+* Reduce the scope of keys reserved for internal use.
 
 ## [0.4.0] - 2023-09-01
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## [Unreleased]
 
+### Behavior Changes
+* `LogBatch::put` and `LogBatch::put_message` allow to put key start with `__` but not equals the `__{ATOMIC_GROUP_KEY}`(0x95 0x95 0x01).
+
 ## [0.4.0] - 2023-09-01
 
 ### Behavior Changes
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -628,7 +628,7 @@ pub(crate) mod tests {
     use super::*;
     use crate::env::{ObfuscatedFileSystem, Permission};
     use crate::file_pipe_log::{parse_reserved_file_name, FileNameExt};
-    use crate::log_batch::{AtomicGroupBuilder, ATOMIC_GROUP_KEY};
+    use crate::log_batch::AtomicGroupBuilder;
     use crate::pipe_log::Version;
     use crate::test_util::{generate_entries, PanicGuard};
     use crate::util::ReadableSize;
@@ -1520,7 +1520,7 @@ pub(crate) mod tests {
             let mut total = 0;
             engine
                 .scan_raw_messages(rid, None, None, false, |k, _| {
-                    assert!(!crate::is_internal_key(k, Some(ATOMIC_GROUP_KEY)));
+                    assert!(!crate::is_internal_key(k, None));
                     total += 1;
                     true
                 })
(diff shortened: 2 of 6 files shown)


WenyXu:
Hi @tabokie PTAL

src/log_batch.rs
```diff
@@ -1562,12 +1562,9 @@ mod tests {
...
-                .put(0, crate::make_internal_key(&[0]), b"v".to_vec())
-                .unwrap_err(),
-            Error::InvalidArgument(_)
-        ));
+        assert!(batch
+            .put(0, crate::make_internal_key(&[0]), b"v".to_vec())
+            .is_ok());
         assert!(matches!(
             batch
                 .put_message(0, crate::make_internal_key(&[1]), &Entry::new())
```
tabokie, line 1570:
```suggestion
                .put_message(0, crate::make_internal_key(ATOMIC_GROUP_KEY), &Entry::new())
```

src/log_batch.rs
```diff
@@ -994,7 +995,6 @@ fn verify_checksum_with_signature(buf: &[u8], signature: Option<u32>) -> Result<
 lazy_static! {
     static ref ATOMIC_GROUP_ID: Arc<AtomicU64> = Arc::new(AtomicU64::new(0));
 }
-const ATOMIC_GROUP_KEY: &[u8] = &[0x01];
```
tabokie, line 997: We should keep it here. Except make it `pub(crate)`

src/log_batch.rs
```diff
@@ -728,7 +729,7 @@ impl LogBatch {
 
     /// Adds a protobuf key value pair into the log batch.
     pub fn put_message<S: Message>(&mut self, region_id: u64, key: Vec<u8>, s: &S) -> Result<()> {
-        if crate::is_internal_key(&key, None) {
+        if crate::is_internal_key(&key, Some(ATOMIC_GROUP_KEY)) {
```
tabokie, line 732: Don't change the callsite. Implement two versions of `is_internal_key` instead:

For `cfg(test)`, keep the behavior the same as before.

For `not(cfg(test))`, we only check `ATOMIC_GROUP_KEY` even for `is_internal_key(key, None)`. And we assert all key prefix passed to `make_internal_key` is `ATOMIC_GROUP_KEY`.

src/lib.rs
```diff
@@ -215,5 +216,6 @@ mod tests {
         assert!(crate::is_internal_key(&key, None));
         assert!(crate::is_internal_key(&key, Some(&[0])));
         assert!(!crate::is_internal_key(&key, Some(&[1])));
+        assert!(!crate::is_internal_key(&key, Some(ATOMIC_GROUP_KEY)));
```
tabokie, line 219: Remove this line. This line succeeds with or without this PR.

tabokie (commented):
Don't forget to update CHANGELOG.md, this should be filed under "Behavior Changes"

CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## [Unreleased]
 
+### Behavior Changes
+* `LogBatch::put` and `LogBatch::put_message` allow to put key start with `__` but not equals the `__{ATOMIC_GROUP_KEY}`(0x95 0x95 0x01).
```
WenyXu, line 6:
```suggestion
* `LogBatch::put` and `LogBatch::put_message` allow to put key start with `__` but not equals the `__{ATOMIC_GROUP_KEY}`(0x5f 0x5f 0x01).
```

src/log_batch.rs
```diff
@@ -728,6 +728,14 @@ impl LogBatch {
 
     /// Adds a protobuf key value pair into the log batch.
     pub fn put_message<S: Message>(&mut self, region_id: u64, key: Vec<u8>, s: &S) -> Result<()> {
+        #[cfg(not(test))]
```
tabokie, line 731: No, this is not what I meant. I mean this:
```
#[not(cfg(test))]
fn is_internal_key(key, prefix) {
  if let Some(p) = prefix { ... }
  else {
    return is_internal_key(key, Some(ATOMIC_GROUP_KEY));
  }
}
#[not(cfg(test))]
fn make_internal_key(key, prefix) {
  assert(prefix == ATOMIC_GROUP_KEY);
  ...
}
```

src/engine.rs
```diff
@@ -1520,7 +1520,7 @@ pub(crate) mod tests {
             let mut total = 0;
             engine
                 .scan_raw_messages(rid, None, None, false, |k, _| {
-                    assert!(!crate::is_internal_key(k, None));
+                    assert!(!crate::is_internal_key(k, Some(ATOMIC_GROUP_KEY)));
```
tabokie, line 1523: Should not change any callsite of `is_internal_key`.

CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## [Unreleased]
 
+### Behavior Changes
+* `LogBatch::put` and `LogBatch::put_message` allow to put key start with `__` but not equals the `__{ATOMIC_GROUP_KEY}`(0x5f 0x5f 0x01).
```
tabokie, line 6: I think this is too verbose, we have never described the exact internal prefix before.

How about "Reduce the scope of keys reserved for internal use."

src/log_batch.rs
WenyXu: Got it

tabokie (approved):
Thanks!

PR #338, opened 2023-09-14
upgrade lz4-sys

SpadeA-Tang:


Cargo.toml
```diff
@@ -43,7 +43,7 @@ if_chain = "1.0"
 lazy_static = "1.3"
 libc = "0.2"
 log = { version = "0.4", features = ["max_level_trace", "release_max_level_debug"] }
-lz4-sys = "1.9"
+lz4-sys = "1.9.4"
```
tabokie, line 46: Refer to semver syntax, the only difference between "1.9" and "1.9.4" is the lower bound, not the upper bound. You should update TiKV's lock file instead.

PR #339, opened 2023-09-18
Normalize import format

Connor1996:
Normalize import format by the rule which is used by TiKV


Connor1996:
PTAL as well @LykxSassinator

LykxSassinator (approved):
LGTM

tabokie (commented):
Need to add format to stable CI and see if it can pass too. This crate is used not limited to nightly compiler.

PR #343, opened 2023-11-03
Return error instead of panicking if rewriting fails

v01dstar:
Ref [#tikv/15755](https://github.com/tikv/tikv/issues/15755)

Return the error instead of panicking if the error won't cause inconsistency while rotating log files.

Ref #131, this PR modifies the decision 3 made in the issue. After this PR,  `create` no longer panics. `truncate` will be retry-able while appending but non-retry-able while closing.

Also updated Cargo.toml to remove the `TODO`.


tabokie (commented):
Most if not all panics in this codebase is necessary. `fsync` error on certain filesystems could cause all buffered write to be lost, and we (currently) have no way to check if those buffered writes belong to current write request or older ones. (https://github.com/tikv/raft-engine/pull/131).

I thought the `is_no_space_err` has already fixed this issue? @LykxSassinator

LykxSassinator:
Yep, the related case can be reviewed in:
https://github.com/tikv/raft-engine/blob/385182bfc29fc2a082ebbe631df1d111f705e782/tests/failpoints/test_engine.rs#L1169

But even though we got the `is_no_space_err`, we still throw the `panic` to users, reminding users to extend the capacity of disk. 
IMO, when error == `is_no_space_err`, it's safe to return the error to users, as the previous work on flush has been done safely.

src/file_pipe_log/pipe.rs
```diff
@@ -321,12 +321,7 @@ impl<F: FileSystem> SinglePipe<F> {
         fail_point!("file_pipe_log::append");
         let mut writable_file = self.writable_file.lock();
         if writable_file.writer.offset() >= self.target_file_size {
-            if let Err(e) = self.rotate_imp(&mut writable_file) {
-                panic!(
-                    "error when rotate [{:?}:{}]: {e}",
-                    self.queue, writable_file.seq,
-                );
-            }
+            self.rotate_imp(&mut writable_file)?;
```
LykxSassinator, line 324:
```suggestion
            if let Err(e) = self.rotate_imp(&mut writable_file) {
                // As the disk is already full and is no valid and extra directory to flush data, it can safely return the `no space err` to users.
                if is_no_space_err(&e) && self.paths.len() == 1 {
                    return Err(e);
                }
                panic!(
                        "error when rotate [{:?}:{}]: {e}",
                        self.queue, writable_file.seq,
                );
            }
```

src/file_pipe_log/pipe.rs
```diff
@@ -372,12 +367,7 @@ impl<F: FileSystem> SinglePipe<F> {
                 // - [3] Both main-dir and spill-dir have several recycled logs.
                 // But as `bytes.len()` is always smaller than `target_file_size` in common
                 // cases, this issue will be ignored temprorarily.
-                if let Err(e) = self.rotate_imp(&mut writable_file) {
-                    panic!(
-                        "error when rotate [{:?}:{}]: {e}",
-                        self.queue, writable_file.seq
-                    );
-                }
+                self.rotate_imp(&mut writable_file)?;
```
LykxSassinator, line 370: ditto.

LykxSassinator, line 370:
```suggestion
                if let Err(e) = self.rotate_imp(&mut writable_file) && is_no_space_err(&e) {
                    return Err(e);
                }
```

src/file_pipe_log/pipe.rs
v01dstar: Why `self.paths.len() == 1`

v01dstar:
> Most if not all panics in this codebase is necessary. `fsync` error on certain filesystems could cause all buffered write to be lost, and we (currently) have no way to check if those buffered writes belong to current write request or older ones. (#131).
> 
The issue seems indicating that `fasync` may return OK even if it actually failed. While, IIUC, your concern was that failed `fsync` may clear the buffer? Which, I don't think it is the case here? Could you please clarify? @tabokie 

> I thought the `is_no_space_err` has already fixed this issue? @LykxSassinator

src/file_pipe_log/pipe.rs
LykxSassinator: > Why `self.paths.len() == 1`

Ignore it, it's redundant as `rotate_imp` will check whether there exists available directory to generate a new file.

LykxSassinator:
FYI, the panic may also happen in the `purge` progress.  U should check the following callings:
- `must_rewrite_append_queue`
- `must_purge_all_stale`
- `rewrite_rewrite_queue`

/cc @v01dstar

tabokie:
The case we are concerned with is after the first fsync fails and clears the buffer, the second fsync returns success, producing the false impression that what hasn't been flushed out in the first fsync is persisted in the second one.

src/file_pipe_log/pipe.rs
```diff
@@ -322,9 +322,13 @@ impl<F: FileSystem> SinglePipe<F> {
         let mut writable_file = self.writable_file.lock();
         if writable_file.writer.offset() >= self.target_file_size {
             if let Err(e) = self.rotate_imp(&mut writable_file) {
+                // As the disk is already full and is no valid and extra directory to flush data, it can safely return the `no space err` to users.
+                if is_no_space_err(&e) && self.paths.len() == 1 {
```
LykxSassinator, line 326:
```suggestion
                if is_no_space_err(&e) {
```

LykxSassinator: As mentioned, no needs to check whether `paths.len() == 1`.

src/file_pipe_log/pipe.rs
```diff
@@ -372,11 +376,8 @@ impl<F: FileSystem> SinglePipe<F> {
                 // - [3] Both main-dir and spill-dir have several recycled logs.
                 // But as `bytes.len()` is always smaller than `target_file_size` in common
                 // cases, this issue will be ignored temprorarily.
-                if let Err(e) = self.rotate_imp(&mut writable_file) {
-                    panic!(
-                        "error when rotate [{:?}:{}]: {e}",
-                        self.queue, writable_file.seq
-                    );
+                if let Err(e) = self.rotate_imp(&mut writable_file) && is_no_space_err(&e) {
+                    return Err(e);
```
LykxSassinator, line 380:
```suggestion
                if let Err(e) = self.rotate_imp(&mut writable_file) {
                    if is_no_space_err(&e) {
                        return Err(e);
                    }
                    panic!(
                        "error when rotate [{:?}:{}]: {e}",
                        self.queue, writable_file.seq
                    );
```

src/file_pipe_log/pipe.rs
```diff
@@ -322,9 +322,13 @@ impl<F: FileSystem> SinglePipe<F> {
...
             if let Err(e) = self.rotate_imp(&mut writable_file) {
+                // As the disk is already full and is no valid and extra directory to flush data, it can safely return the `no space err` to users.
+                if is_no_space_err(&e) && self.paths.len() == 1 {
+                    return Err(e);
+                }
                 panic!(
-                    "error when rotate [{:?}:{}]: {e}",
-                    self.queue, writable_file.seq,
+                        "error when rotate [{:?}:{}]: {e}",
+                        self.queue, writable_file.seq,
```
LykxSassinator, line 331:
```suggestion
                    "error when rotate [{:?}:{}]: {e}",
                    self.queue, writable_file.seq,
```

src/file_pipe_log/pipe.rs
LykxSassinator: And please check the errors in the ut.

src/file_pipe_log/pipe.rs
LykxSassinator, line 326:
```suggestion
                // As the disk is already full and is no valid directory to flush data, it can safely return
                // the `no space err` to users.
                if is_no_space_err(&e) {
```

LykxSassinator: As mentioned, no needs to check whether `paths.len() == 1`.

v01dstar:
> FYI, the panic may also happen in the `purge` progress. U should check the following callings:
> 
>     * `must_rewrite_append_queue`
> 
>     * `must_purge_all_stale`
> 
Are above 2 functions being called in non-test code path? If not, I don't think we need to propagate the error up.
>     * `rewrite_rewrite_queue`
> 
This is already covered, isn't it.
> 
> /cc @v01dstar

tabokie:
> I think it probably makes more sense to push down the panic close to fsync call.

Need to at least try to do this. AFAIK there's no guarantee that pwrite will fail before fsync in the case of disk full. Especially when raft-engine disk is shared with other parties.

LykxSassinator:
> > I think it probably makes more sense to push down the panic close to fsync call.
> 
> Need to at least try to do this. AFAIK there's no guarantee that pwrite will fail before fsync in the case of disk full.

IMO, specify whether the returned error from `fdatasync` is a `nospace` error is a better choice. Users/Callers can decide `panic` directly or bubble the error to upper calls.  And if it was a `nospace` error, `rotate` could return this error.
```Rust
    // src/env/log_fd/unix.rs
    #[inline]
    fn sync(&self) -> IoResult<()> {
        fail_point!("log_fd::sync::err", |_| {
            Err(from_nix_error(nix::Error::EINVAL, "fp"))
        });
        #[cfg(target_os = "linux")]
        {
            nix::unistd::fdatasync(self.0).map_err(|e| match e {
                Errno::ENOSPC => from_nix_error(e, "nospace"),
                _ => from_nix_error(e, "fdatasync"),
            })
        }
        #[cfg(not(target_os = "linux"))]
        {
            nix::unistd::fsync(self.0).map_err(|e| match e {
                Errno::ENOSPC => from_nix_error(e, "nospace"),
                _ => from_nix_error(e, "fsync"),
            })
        }
    }

```
/cc @v01dstar

tabokie:
Why? `fsync` returning NoSpace does not guarantee you anything. The idea is to still always panic when fsync fails, but we can be smarter and not panic when pwrite fails.

LykxSassinator:
> Why? `fsync` returning NoSpace does not guarantee you anything. The idea is to still always panic when fsync fails, but we can be smarter and not panic when pwrite fails.

Returning the `NOSPC` error can make callers know the disk just reach the limit of capacity, rather than just bubbling an error and directly `panic` with this error.

v01dstar:
> #131

[`fsync` or `fdatasync` do not return ENOSPC error](https://pubs.opengroup.org/onlinepubs/009696799/functions/fdatasync.html)

v01dstar:
> Why? `fsync` returning NoSpace does not guarantee you anything. The idea is to still always panic when fsync fails, but we can be smarter and not panic when pwrite fails.

PTAL. I move panic inside `LogFileWriter::sync()` and `LogFileWriter::truncate()`, 2 operations that may cause inconsistency. However, at the same time, I make raft-engine not panic if `create()`, `write_header()` fail during rotating files, which changes the decisions made in #131 a little bit. I think, it is safe to return the error (no matter what the error is) in such cases, because the caller would not have false expectations, and no inconsistency would happen neither, since raft-engine will recycle the broken file (if any) next time rotating the file.

src/file_pipe_log/log_file.rs
```diff
@@ -67,28 +67,26 @@ impl<F: FileSystem> LogFileWriter<F> {
...
-        self.writer.seek(SeekFrom::Start(0))?;
+        self.writer.rewind()?;
         self.written = 0;
         let mut buf = Vec::with_capacity(LogFileFormat::encoded_len(format.version));
         format.encode(&mut buf).unwrap();
         self.write(&buf, 0)
     }
 
-    pub fn close(&mut self) -> IoResult<()> {
+    pub fn close(&mut self) {
```
Connor1996, line 77: why not let them return the error?

v01dstar: Basically, I want the idea of "panic if truncate or sync fail" to be consistent as much as possible cross the whole module. So, it is best to make it as lower level as possible. Then the call sites won't have to debate on the expected behaviors anymore. `truncate` and `sync` do not return errors, so `close` doesn't need to.

tabokie: "truncate" failure is retriable (at least they should). In fact, we rely on it to recover from disk full (after buffered write is full, we still need to remove partial writes from page cache, otherwise in the next recovery we will abandon all writes beyond this point)

tabokie: Truncate is still retriable. For example when closing file before shutdown, its failure is totally okay. The case where we panic when truncate fails is purely for the sake of convenience. We can also retry truncate for a few times. Fsync failure is different, its characteristics is determined by filesystem and we should always panic regardless of the application or engine code context.

v01dstar: So, you are proposing that, in this PR. In addition to the change of behavior of `create`, change `truncate` as well?

tabokie: The general guideline is, for errors that can be retried without causing data inconsistency, bubble them out, otherwise panic. However we can choose which layer of interface to enforce this protocol. Previously we enforce it for `PipeLog` trait. By "push down the panic", we need to move this layer to lower modules such as writers and private functions. We can refer to this property as "fail-safe" (which is a lesser guarantee than panic-safe).

To be more specific:
- Make `Writer::sync` fail-safe and panic inside if fsync fails.
- Make `SinglePipe::rotate_imp` fail-safe and panic if `Writer::close` fails, because continuing this this failure would result in data inconsistency at next restart. (This inconsistency is "pipe"-level, not file-level. So we shouldn't panic at file writer level.)
- Bubble any error at `SinglePipe::append`.

v01dstar: PTAL

src/file_pipe_log/log_file.rs
```diff
@@ -67,24 +67,26 @@ impl<F: FileSystem> LogFileWriter<F> {
...
+        Ok(())
     }
 
-    pub fn truncate(&mut self) -> IoResult<()> {
+    pub fn truncate(&mut self) -> Result<()> {
         if self.written < self.capacity {
             fail_point!("file_pipe_log::log_file_writer::skip_truncate", |_| {
                 Ok(())
             });
+            // Panic if truncate fails, in case of data loss.
```
LykxSassinator, line 89:
```suggestion
```

src/file_pipe_log/log_file.rs
```diff
@@ -67,24 +67,26 @@ impl<F: FileSystem> LogFileWriter<F> {
...
-        self.writer.seek(SeekFrom::Start(0))?;
+        self.writer.rewind()?;
         self.written = 0;
         let mut buf = Vec::with_capacity(LogFileFormat::encoded_len(format.version));
         format.encode(&mut buf).unwrap();
         self.write(&buf, 0)
     }
 
-    pub fn close(&mut self) -> IoResult<()> {
+    pub fn close(&mut self) -> Result<()> {
```
LykxSassinator, line 77:
```suggestion
    pub fn close(&mut self) -> IoResult<()> {
```

src/file_pipe_log/log_file.rs
```diff
@@ -67,24 +67,26 @@ impl<F: FileSystem> LogFileWriter<F> {
...
         // Necessary to truncate extra zeros from fallocate().
         self.truncate()?;
-        self.sync()
+        self.sync();
+        Ok(())
     }
 
-    pub fn truncate(&mut self) -> IoResult<()> {
+    pub fn truncate(&mut self) -> Result<()> {
```
LykxSassinator, line 84:
```suggestion
    pub fn truncate(&mut self) -> IoResult<()> {
```

LykxSassinator (commented):
Rest LGTM

src/file_pipe_log/pipe.rs
```diff
@@ -72,9 +72,7 @@ pub(super) struct SinglePipe<F: FileSystem> {
 impl<F: FileSystem> Drop for SinglePipe<F> {
     fn drop(&mut self) {
         let mut writable_file = self.writable_file.lock();
-        if let Err(e) = writable_file.writer.close() {
-            error!("error while closing the active writer: {e}");
-        }
+        writable_file.writer.close().unwrap();
```
tabokie, line 75: Revert this.

src/file_pipe_log/log_file.rs
```diff
@@ -117,10 +119,10 @@ impl<F: FileSystem> LogFileWriter<F> {
         Ok(())
     }
 
-    pub fn sync(&mut self) -> IoResult<()> {
+    pub fn sync(&mut self) {
```
tabokie, line 122: I think there's no need to change signature even if it never raises error. Especially for trait method `PipeLog::sync`.

src/file_pipe_log/pipe.rs
```diff
@@ -248,7 +246,7 @@ impl<F: FileSystem> SinglePipe<F> {
         let new_seq = writable_file.seq + 1;
         debug_assert!(new_seq > DEFAULT_FIRST_FILE_SEQ);
 
-        writable_file.writer.close()?;
+        writable_file.writer.close().unwrap();
```
tabokie, line 249: Add a comment to explain "Otherwise it will be considered a corruption at next restart".

src/file_pipe_log/pipe.rs
```diff
@@ -272,7 +270,7 @@ impl<F: FileSystem> SinglePipe<F> {
         };
         // File header must be persisted. This way we can recover gracefully if power
         // loss before a new entry is written.
-        new_file.writer.sync()?;
+        new_file.writer.sync();
         self.sync_dir(path_id)?;
```
tabokie, line 274: This error needs to be handled carefully now. (e.g. remove the newly created file and make sure the old writer is okay to write again) Better just unwrap it as well.

tabokie: `build_file_writer` above is the same.

src/file_pipe_log/log_file.rs
v01dstar: If it return a `Result`, wouldn't it confuse people of how should they deal with the error, like whether to yield the error up or just `unwrap()`?

I think leave it this way ease the future maintenance work?

src/file_pipe_log/pipe.rs
v01dstar: Can you explain more about this comment?

I guess "It" means the writeable file that is to be closed? then, who will consider it a corruption? and what does the "next restart" mean here, if it returns error instead of calling `unwrap()`, there won't be "restart"?

src/file_pipe_log/pipe.rs
v01dstar: Made `sync_dir` panic if it fails.

But `build_file_writer` should be fine, right? It is the type of panic this PR trying to avoid (this can be confirmed by `test_no_space_write_error`). If it fails, the new file  won't be used for writing and will be recycled the next time `rotate_impl` is called. So, it already meet your expectation?

src/file_pipe_log/pipe.rs
v01dstar: Done

src/file_pipe_log/log_file.rs
v01dstar: Done

src/file_pipe_log/log_file.rs
v01dstar: Done

src/file_pipe_log/log_file.rs
v01dstar: Done

src/file_pipe_log/pipe.rs
tabokie: I guess it's not needed anymore, after writer::sync can panic. The original case is when user calls engine::write and it fails, then he writes again, the file might be considered closed when physically it's not. Leaving a section of junk data and when user next restart the server, engine recovery will consider the junk data a potential corruption and refuse to continue.

src/file_pipe_log/log_file.rs
tabokie: The characteristics of fsync (fsync syscall, not sync method here) is file-system dependent. It should not affect how user interface is designed. For example, we can do internal buffering inside writer, then the pwrite syscall before fsync syscall will not trigger a panic. As for "how should they deal with the error", you just need to add some comments to explain the guarantee of the function (both the trait method and writer method).

LykxSassinator (approved):

src/file_pipe_log/log_file.rs
```diff
@@ -119,8 +119,9 @@ impl<F: FileSystem> LogFileWriter<F> {
 
     pub fn sync(&mut self) -> IoResult<()> {
         let _t = StopWatch::new(&*LOG_SYNC_DURATION_HISTOGRAM);
-        self.handle.sync()?;
-        Ok(())
+        // Panic if sync fails, in case of data loss.
+        self.handle.sync().unwrap();
+        IoResult::Ok(())
```
Connor1996, line 124:
```suggestion
        Ok(())
```

src/file_pipe_log/pipe.rs
```diff
@@ -403,15 +393,11 @@ impl<F: FileSystem> SinglePipe<F> {
...
         let mut writable_file = self.writable_file.lock();
-        let seq = writable_file.seq;
         let writer = &mut writable_file.writer;
-        {
-            let _t = StopWatch::new(perf_context!(log_sync_duration));
-            if let Err(e) = writer.sync() {
-                panic!("error when sync [{:?}:{seq}]: {e}", self.queue);
-            }
+        let _t = StopWatch::new(perf_context!(log_sync_duration));
+        if let Err(e) = writer.sync() {
```
Connor1996, line 398:
```suggestion
        writer.sync().map_err(|e| Error::Io(e))?;
```

src/file_pipe_log/pipe.rs
v01dstar: Added a few more verifications in test_file_rotate_error test, should be able to address your concern? PTAL

src/file_pipe_log/log_file.rs
v01dstar: Done

Connor1996 (approved):
LGTM

src/file_pipe_log/pipe.rs
```diff
@@ -273,7 +273,9 @@ impl<F: FileSystem> SinglePipe<F> {
         // File header must be persisted. This way we can recover gracefully if power
         // loss before a new entry is written.
         new_file.writer.sync()?;
-        self.sync_dir(path_id)?;
+        // Panic if sync calls fail, keep consistent with the behavior of
+        // `LogFileWriter::sync()`.
+        self.sync_dir(path_id).unwrap();
```
tabokie, line 278: panic inside sync_dir as well.

src/file_pipe_log/pipe.rs
```diff
@@ -248,7 +248,7 @@ impl<F: FileSystem> SinglePipe<F> {
         let new_seq = writable_file.seq + 1;
         debug_assert!(new_seq > DEFAULT_FIRST_FILE_SEQ);
 
-        writable_file.writer.close()?;
+        writable_file.writer.close().unwrap();
```
tabokie, line 251: No need to unwrap now.

src/file_pipe_log/log_file.rs
```diff
@@ -67,7 +67,7 @@ impl<F: FileSystem> LogFileWriter<F> {
     }
```
tabokie, line 67: Add a comment to this struct stating it should be fail-safe, i.e. user can still use the writer without breaking data consistency if any operation has failed.

src/filter.rs
```diff
@@ -333,7 +333,7 @@ impl RhaiFilterMachine {
                     )?;
                     log_batch.drain();
                 }
-                writer.close()?;
+                writer.close().unwrap();
```
tabokie, line 336: ditto

src/purge.rs
```diff
@@ -273,7 +273,7 @@ where
     // Rewrites the entire rewrite queue into new log files.
     fn rewrite_rewrite_queue(&self) -> Result<Vec<u64>> {
         let _t = StopWatch::new(&*ENGINE_REWRITE_REWRITE_DURATION_HISTOGRAM);
-        self.pipe_log.rotate(LogQueue::Rewrite)?;
+        self.pipe_log.rotate(LogQueue::Rewrite).unwrap();
```
tabokie, line 276: why unwrap this?

tests/failpoints/test_io_error.rs
```diff
@@ -165,20 +165,24 @@ fn test_file_rotate_error() {
     {
```
tabokie, line 165: Make two versions of this test: `fn test_file_rotate_error(restart: bool)`

```rust
// case 1
if restart {
  let engine = Engine::open_with_file_system(cfg.clone(), fs.clone()).unwrap();
}
// case 2
// ...
```

tests/failpoints/test_io_error.rs
```diff
@@ -138,88 +137,113 @@ fn test_file_rotate_error() {
     let fs = Arc::new(ObfuscatedFileSystem::default());
     let entry = vec![b'x'; 1024];
 
-    let engine = Engine::open_with_file_system(cfg.clone(), fs.clone()).unwrap();
-    engine
+    let mut engine = Some(Engine::open_with_file_system(cfg.clone(), fs.clone()).unwrap());
+    let mut engine_ref = engine.as_ref().unwrap();
```
tabokie, line 141: No need, you can re-assign a variable after it's moved, e.g. `drop(engine); engine = Engine::new();`

tabokie (approved):
Rest LG

PR #345, opened 2023-12-12
Revert rustc version

v01dstar:
Turns out it is better to still use rustc 1.66.0, otherwise the downstream TiKV project would require a lot of refactoring works if it wants to update its raft-engine version.

I bumped the version in #343 because`serde` requires rustc > 1.67 for stable build. Seems that we get the CI passed, we just need to update the CI's toolchain version, instead of updating the rustc requirement of raft-engine.


tonyxuqqi (approved):

Connor1996 (approved):
LGTM

tabokie:
Just a nitpick, you should use cargo patch to lock the version of toml to get around this issue. Setting a MSRV lower than actually required is not friendly to potential users. And CI and MSRV *must* align, that's the point of explicitly setting one in the first place.

PR #347, opened 2023-12-31
rewrite: optimize the interval of `sync` when rewriting memtables.

LykxSassinator:
In a cloud environment, refraining from unscheduling sync operations when rewriting memtables might result in an accumulation of unsynced bytes in the buffer. This accumulation has the potential to impede the foreground write progress during sync.

This pull request introduces periodic sync operations when the amount of stashed unsynced bytes exceeds a predefined threshold. This optimization aims to address the issue and enhance performance.

PR #349, opened 2024-01-09
[cherry-pick] rewrite: optimize the interval of `sync` when rewriting memtables. (#347)

LykxSassinator:
This is an manual cherry-pick from #347 .

In a cloud environment, refraining from unscheduling sync operations when rewriting memtables might result in an accumulation of unsynced bytes in the buffer. This accumulation has the potential to impede the foreground write progress during sync.

This pull request introduces periodic sync operations when the amount of stashed unsynced bytes exceeds a predefined threshold. This optimization aims to address the issue and enhance performance.


overvenus (approved):

Connor1996 (approved):
LGTM

PR #358, opened 2024-04-19
engine: supply the metric of the compression ratio on write.

LykxSassinator:
## Description
Supply one extra metric to obeserve the compression ratio change of `LogBatch` when writing.

With this pr, the compression ratio can be reviewed as following shows:
image

Meanwhile, the format of dependencies specified in `Cargo.toml` is cleared up.


src/metrics.rs
```diff
@@ -201,6 +201,12 @@ lazy_static! {
         exponential_buckets(256.0, 1.8, 22).unwrap()
     )
     .unwrap();
+    pub static ref ENGINE_WRITE_COMPRESSION_RATIO_HISTOGRAM: Histogram = register_histogram!(
+        "raft_engine_write_compression_ratio",
+        "Bucketed histogram of Raft Engine write compression ratio",
+        exponential_buckets(0.0005, 1.8, 16).unwrap()
```
glorv, line 207: Does compression ratio fit such a wide range? In general, the ratio should be within 1~10, maybe  `exponential_buckets(0.5, 1.2, 16)` is enough? and it can provide higher precision.

glorv (commented):
rest LGTM

src/metrics.rs
LykxSassinator: Here, the compression_ratio = `compressed_bytes_size / uncompressed_bytes_size`, not referring to Rocksdb, where the compression_ratio is calculated with `uncompressed_bytes_size / compressed_bytes_size`.

overvenus (approved):

src/metrics.rs
glorv: Got it.

glorv (approved):

LykxSassinator:
@glorv Can u help me to merge this pr into master branch ? thx.

PR #370, opened 2024-11-04
Fix `panics` on reading entries by stale index.

LykxSassinator:
Close #https://github.com/tikv/tikv/issues/17383


Connor1996 (approved):
LGTM

PR #379, opened 2025-05-14
Unlock the LOCK files on drop

v01dstar:
We need to restart raft engine in a test case. So we need to unlock the LOCK files on drop, so we can open them again in the same process.

Also update rustc version and fix clippy errors, since some dependencies require higher versions of Rust


Cargo.toml
```diff
@@ -3,7 +3,7 @@ name = "raft-engine"
 version = "0.4.2"
 authors = ["The TiKV Project Developers"]
 edition = "2018"
-rust-version = "1.75.0"
+rust-version = "1.85.0"
```
overvenus, line 6: Out of curiosity, does `fs2::FileExt` require the new rust version?

v01dstar: No, it was not because of fs2::FileExt. It was because raft-engine as a lib does not have Cargo.lock, and one of the dependency`half` (required by citron) that requires 1.85+.
This is fine, since TiKV master branch uses 1.85+ nightly

src/file_pipe_log/pipe.rs
```diff
@@ -464,7 +465,17 @@ impl<F: FileSystem> SinglePipe<F> {
...
+}
+
+impl<F: FileSystem> Drop for DualPipes<F> {
+    fn drop(&mut self) {
+        for lock in &self.dir_locks {
+            if let Err(e) = FileExt::unlock(lock) {
+                error!("error while unlocking directory: {e}");
+            }
+        }
+    }
```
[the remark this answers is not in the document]
v01dstar: Done

overvenus (approved):

zhangjinpeng87 (approved):
LGTM

overvenus (approved):

LykxSassinator (approved):

PR #381, opened 2025-06-24
Release dir locks on engine closing

v01dstar:
This is a cherry-pick of #379

We need to restart the Raft Engine within a test case to apply a configuration change. To do this, we must ensure the LOCK files are released upon drop, allowing the engine to be reopened within the same process.

Also update rustc version and fix clippy errors, since some dependencies require higher versions of Rust


LykxSassinator (approved):

Connor1996 (approved):
LGTM

PR #382, opened 2025-07-02
Optimize `fetch_entries_to` function to avoid stucking `write` too long.

LykxSassinator:
## Background
PR [#370](https://github.com/tikv/raft-engine/pull/370) fixed a panic issue caused by concurrent updates to `Memtable` followed by reads on stale indexes. However, we observed that this fix introduced performance regressions under concurrent read/write workloads.

## Issue
When `fetch_entries_to` retrieves a large batch of entries from disk, write operations are blocked until the read completes, delaying `Memtable` updates and degrading throughput.

## Solution
This PR optimizes `fetch_entries_to` to reduce contention and improve performance under mixed workloads.

## Results
| Branch | Status |
| --- | --- |
| [Master](https://github.com/tikv/raft-engine/commit/392f5e66f8286dc1b6d7cf69f2bc20ed72d40123) | image |
| This PR | image |

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -334,14 +334,27 @@ where
         let _t = StopWatch::new(&*ENGINE_READ_ENTRY_DURATION_HISTOGRAM);
         if let Some(memtable) = self.memtables.get(region_id) {
             let mut ents_idx: Vec<EntryIndex> = Vec::with_capacity((end - begin) as usize);
-            // Ensure that the corresponding memtable is locked with a read lock before
-            // completing the fetching of entries from the raft logs. This
-            // prevents the scenario where the index could become stale while
-            // being concurrently updated by the `rewrite` operation.
-            let immutable = memtable.read();
-            immutable.fetch_entries_to(begin, end, max_size, &mut ents_idx)?;
+            memtable.read().fetch_entries_to(begin, end, max_size, &mut ents_idx)?;
             for i in ents_idx.iter() {
-                vec.push(read_entry_from_file::<M, _>(self.pipe_log.as_ref(), i)?);
+                vec.push({
+                    match read_entry_from_file::<M, _>(self.pipe_log.as_ref(), i) {
+                        Ok(entry) => entry,
+                        Err(e) => {
+                            // The index is not found in the file, it means the entry is already
+                            // stale by `compact` or `rewrite`. Retry to read the entry from the memtable.
+                            let immutable = memtable.read();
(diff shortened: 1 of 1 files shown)


src/engine.rs
```diff
@@ -334,14 +334,31 @@ where
...
-            // being concurrently updated by the `rewrite` operation.
-            let immutable = memtable.read();
-            immutable.fetch_entries_to(begin, end, max_size, &mut ents_idx)?;
+            memtable
+                .read()
+                .fetch_entries_to(begin, end, max_size, &mut ents_idx)?;
             for i in ents_idx.iter() {
-                vec.push(read_entry_from_file::<M, _>(self.pipe_log.as_ref(), i)?);
+                vec.push({
+                    match read_entry_from_file::<M, _>(self.pipe_log.as_ref(), i) {
```
glorv, line 342: Can we ensure the safety to read entry without holding the lock? Is it possible that some entries are truncated and target wal files are Gced(or reused) before this read, then the result of this read is undefined behavior

LykxSassinator: If it uses a stale file handle to get the entry, it will get the `Error(OutOfSeq)`, ref:
https://github.com/LykxSassinator/raft-engine/blob/392f5e66f8286dc1b6d7cf69f2bc20ed72d40123/src/file_pipe_log/pipe.rs#L238

So, if the first `fetch` uses a stale index to get the entry, it will return an Error and trigger the second retry, where it will use the latest index to access the entry.

glorv (approved):
LGTM

src/engine.rs
```diff
@@ -2246,6 +2263,14 @@ pub(crate) mod tests {
             old_perf_context.apply_duration,
             new_perf_context.apply_duration
         );
+
+        use rand::{thread_rng, Rng};
+        let huge_data: Vec<u8> = (0..3 * 1024 * 1024).map(|_| thread_rng().gen()).collect();
+        for i in 5..200 {
+            engine.append(rid, i, i + 1, Some(&huge_data));
+            let new_perf_context = get_perf_context();
+            println!("[i: {}] - the details of perf: {:?}", i, new_perf_context);
```
overvenus, line 2272: Forgot to remove `println`?

overvenus (commented):
Rest LGTM

src/engine.rs
LykxSassinator: My mistake, it's used for debugging other performance issues.

overvenus (approved):

PR #397, opened 2026-04-24
Gate lz4-sys behind an optional lz4-compression feature

humancto:
## Summary

Make the `lz4-sys` C FFI crate optional behind a new `lz4-compression` Cargo feature (default-on). Consumers who build with `default-features = false` and without `lz4-compression` get a pure-Rust build graph with `lz4-sys` fully absent; existing consumers on defaults see zero behavior change.

## Motivation

Projects with a strict pure-Rust policy (no C dependencies in the build graph) currently cannot consume `raft-engine` at all because `lz4-sys = "=1.9.5"` is unconditional. Making the C dep opt-in-by-default removes that adoption blocker without affecting anyone who stays on defaults.

## Changes

- `Cargo.toml`: `lz4-sys` marked `optional = true`; new feature `lz4-compression = ["dep:lz4-sys"]`; `default = ["internals", "scripting", "lz4-compression"]` preserves existing BC.
- `src/util.rs`: `pub mod lz4` split into two `#[cfg]`-gated variants with identical public surface (`DEFAULT_LZ4_COMPRESSION_LEVEL`, `append_compress_block`, `decompress_block`). Real implementation behind `#[cfg(feature = "lz4-compression")]`. Stub under `#[cfg(not(...))]` returns `Error::Other` with a message naming the feature flag. Call sites in `src/log_batch.rs` compile unchanged under both arms.
- `src/config.rs`: `Config::sanitize` rejects `batch_compression_threshold > 0` when the feature is off, matching the existing reject patterns for `enable_log_recycle`/`prefill_for_recycle`. Rustdoc on `batch_compression_threshold` spells out the constraint. New unit test covers both reject and accept paths.
- `CHANGELOG.md` + `README.md`: entry under `[Unreleased]` and a line in the "Available Cargo features" list.
- `.github/workflows/rust.yml`: new `cargo build --no-default-features --verbose` step in the stable job to regression-guard the feature gate.

## Operator contract

Callers who disable the feature:

- MUST set `batch-compression-threshold = 0` in `Config` (enforced by `sanitize`).
- CANNOT decode log files written by a build that had LZ4 compression active — those entries surface `Error::Other` on read, naming the feature flag.

## Test plan

- [x] `cargo check --lib` passes (0 errors).
- [x] `cargo check --lib --no-default-features` passes (0 errors).
- [x] `cargo tree --no-default-features --edges no-dev | grep lz4` → empty.
- [x] `cargo tree --edges no-dev | grep lz4` → `lz4-sys v1.9.5` still present (defaults unchanged).
- [x] Build verification covered in CI by the new `--no-default-features` step.
- [x] Adversarial review by an outside Rust reviewer: two rounds (REVISE → APPROVE_WITH_NITS). Both rounds' feedback folded in.

## Pre-existing warning flagged in review

The new CI step will surface a pre-existing `warning: Patch cc v1.0.98 ... was not used in the crate graph.` This is the `[patch.crates-io] cc = { ... tag = "1.0.98" }` pin added for [rust-lang/cc-rs#984](https://github.com/rust-lang/cc-rs/issues/984), which no longer matches the resolved `cc 1.2.61` in the lockfile. The warning is orthogonal to this change and predates it — left as a follow-up since the original #984 workaround should be re-evaluated on its own merits rather than bundled into a feature-gate patch.

## Background

Part of the [humancto/mango](https://github.com/humancto/mango) project's adoption of `raft-engine` as its WAL. Tracking the fork's retirement against this upstream PR's merge.

## Summary by CodeRabbit

* **New Features**
  * Added an optional, default-enabled LZ4 compression feature that toggles native compression; builds without it remain usable but will error if compression is requested or compressed data is encountered.

* **Documentation**
  * Updated README and CHANGELOG with feature behavior, compatibility notes, and required configuration when disabling compression.

* **Chores**
  * CI now runs an additional build-and-test step with default features disabled.

* **Tests**
  * Added tests validating configuration rejection when compression is disabled.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -38,8 +38,10 @@ jobs:
         run: make clippy
         env:
           WITH_STABLE_TOOLCHAIN: 'force'
-      - name: Build (no default features)
-        run: cargo build --no-default-features --verbose
+      - name: Build and test (no default features)
+        run: |
+          cargo check --no-default-features --all-targets --verbose
+          cargo test --no-default-features --verbose
       - name: Run tests
         run: make test
         env:
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -38,6 +38,8 @@ jobs:
         run: make clippy
         env:
           WITH_STABLE_TOOLCHAIN: 'force'
+      - name: Build (no default features)
+        run: cargo build --no-default-features --verbose
       - name: Run tests
         run: make test
         env:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### New Features
 
 * Add a new Prometheus metric `raft_engine_write_compression_ratio` to track compression ratio of write #358
+* Gate the `lz4-sys` C FFI dependency behind a new `lz4-compression` Cargo feature (on by default). Consumers who build with `default-features = false` and without `lz4-compression` get a fully pure-Rust build graph; `Config::sanitize` rejects a non-zero `batch-compression-threshold` in that configuration.
 
 ## [0.4.2] - 2024-04-16
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -46,7 +46,7 @@ log = { version = "0.4", features = [
   "max_level_trace",
   "release_max_level_debug",
 ] }
-lz4-sys = { version = "=1.9.5" }
+lz4-sys = { version = "=1.9.5", optional = true }
 memmap2 = { version = "0.9", optional = true }
 nix = "0.26"
 num-derive = "0.4"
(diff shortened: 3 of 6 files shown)


humancto:
Addressed the CodeRabbit nitpick on the no-default-features CI step in ea4e85c413eba37784193b47e1698096cdacfde6.

The previous `cargo build --no-default-features --verbose` step neither compiled nor ran the new `test_sanitize_rejects_nonzero_threshold_without_lz4_feature` test in `src/config.rs` (gated under `#[cfg(not(feature = "lz4-compression"))]`), so the test silently never executed. Replaced with:

```yaml
- name: Build and test (no default features)
  run: |
    cargo check --no-default-features --all-targets --verbose
    cargo test --no-default-features --verbose
```

`--all-targets` covers tests/benches/examples in the feature-off graph, and the explicit `cargo test` invocation actually exercises both the reject (`threshold > 0` → `Err`) and accept (`threshold == 0` → `Ok`) paths in `Config::sanitize`.

PR #401, opened 2026-04-29
memtable: Shrink sparse raft-engine memtable entry index buffers more proactively after compaction. 

### LINKED ISSUE #19544: raft-engine's memory usage increases slowly overtime which the workload is stable
### What version of TiKV are you using?

v8.5.4

### What did happened?

From a prod cluster, I observe the memory usage of raft-engine increase slowly overtime which the workload and region number has little change.

Image

The instance restartad at 03-17. Since the, the memory trace of raft-engine increased consistently.

Image

Memory profiling show the memory is consumed by raft-engine's memory table.

https://github.com/tikv/raft-engine/blob/f0fcebe922c384fcb673576f1f5b638203550ee7/src/memtable.rs#L622-L626

From raft-engine's memory table implementation, it look like the memory increment is related to the memory buffer shrink policy. It only trigger shrink when the vector's capacity beyond 1024 which is relatively a large threshold.

LykxSassinator:
## Problem

Close https://github.com/tikv/tikv/issues/19544

`raft-engine` memtable memory can grow slowly over time even when workload and region count stay relatively stable.

A notable source is `MemTable::entry_indexes`: after entries are compacted away, the underlying deque may continue to retain a large amount of spare capacity. The previous shrink policy only triggered once capacity grew beyond `1023`, and when triggered it used `shrink_to_fit()`, which was coarse and could still leave room for improvement in long-running stable clusters.

## Solution

This PR makes entry index shrinking more proactive, but keeps it conservative enough to avoid unnecessary capacity churn.

The new policy still only runs after `compact_to()`, and shrinks `entry_indexes` only when all of the following are true:
- the buffer has grown past a minimum capacity threshold;
- live entries occupy at most 50% of the allocated slots;
- shrinking would reclaim a meaningful amount of memory.

Instead of shrinking all the way down to `len`, the new logic keeps a small headroom window (`len + CAPACITY_INIT`) so a freshly compacted memtable does not immediately regrow on the next small append burst.

In short, this change targets long-lived sparse memtables while avoiding a simple “lower the threshold and always shrink harder” strategy.

## Why this should not introduce a meaningful performance regression

The new logic does not run on the append hot path. It is only evaluated after `compact_to()`, and the fast path is just a few integer checks.

Actual reallocations become possible earlier than before, but only for clearly sparse memtables. The added utilization gate, minimum reclaim gate, and retained headroom together are intended to avoid frequent shrink/regrow oscillation.

This means the tradeoff is:
- slightly more compact-side realloc opportunities in sparse cases;
- noticeably better reclamation of long-lived unused entry index capacity.

Given that the logic stays off the append hot path and avoids `shrink_to_fit()` churn, the risk of a meaningful performance regression is low.

## Validation

Added targeted regression tests in `src/memtable.rs`:

1. `test_memtable_compact_shrinks_sparse_entry_indexes`
- Verifies that a sparse memtable actually shrinks after compaction.
- Also verifies the new policy keeps headroom instead of shrinking all the way to `len`.

2. `test_memtable_compact_keeps_busy_entry_indexes_capacity`
- Verifies that a still-busy memtable does not shrink just because compaction happened.

3. `test_memtable_compact_keeps_capacity_when_reclaim_is_too_small`
- Verifies that the deque is not shrunk when the reclaimable space is too small to justify a reallocation.

4. `test_memtable_append_after_compact_shrink_keeps_span_consistent`
- Verifies span, fetch, and live entry accounting remain correct after a shrink followed by new appends.

Build validation:
- `cargo +nightly-2025-04-03 fmt --all --check`
- `cargo +nightly-2025-04-03 check --lib`
- `cargo +nightly-2025-04-03 check --lib --no-default-features --features internals`

Note: full `cargo test` is currently blocked in this environment by the existing Darwin/arm64 `grpcio-sys` toolchain issue rather than by this patch itself.

## Summary by CodeRabbit

* **Refactor**
  * Revised container shrinking to use utilization-aware capacity planning, making memory reclamation more aggressive while preserving a slack window to avoid frequent resizes.

* **Tests**
  * Added tests for sparse-shrink, high-utilization (no shrink), minimal-reclaim (skip shrink), and post-shrink correctness with subsequent appends.

* **Style**
  * Minor diagnostic message formatting tweaks for clearer error output.

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -2616,7 +2616,7 @@ pub(crate) mod tests {
             assert!(data.remove(&rid), "{}", rid);
             assert_eq!(engine.get(rid, &key).unwrap(), value);
         }
-        assert!(data.is_empty(), "data loss {:?}", data);
+        assert!(data.is_empty(), "data loss {data:?}");
     }
 
     #[test]
--- a/src/file_pipe_log/log_file.rs
+++ b/src/file_pipe_log/log_file.rs
@@ -111,7 +111,7 @@ impl<F: FileSystem> LogFileWriter<F> {
             self.writer
                 .seek(SeekFrom::Start(self.written as u64))
                 .unwrap_or_else(|e| {
-                    panic!("failed to reseek after write failure: {}", e);
+                    panic!("failed to reseek after write failure: {e}");
                 });
         })?;
         self.written = new_written;
--- a/src/memtable.rs
+++ b/src/memtable.rs
@@ -1741,7 +1741,7 @@ mod tests {
         let capacity_before = memtable.entry_indexes.capacity();
         assert_eq!(capacity_before, CAPACITY_SHRINK_THRESHOLD);
 
-        assert_eq!(memtable.compact_to(next - 40), (next - 40) as u64);
+        assert_eq!(memtable.compact_to(next - 40), next - 40);
         assert_eq!(memtable.entry_indexes.len(), 40);
         assert_eq!(memtable.entry_indexes.capacity(), capacity_before);
         memtable.consistency_check();
--- a/src/memtable.rs
+++ b/src/memtable.rs
@@ -66,8 +66,12 @@ mod swap_conditional_imports {
 
 use swap_conditional_imports::*;
 
-/// Attempt to shrink entry container if its capacity reaches the threshold.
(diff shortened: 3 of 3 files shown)


glorv (approved):
LGTM

overvenus (approved):

